PR #70, opened 2019-01-12
fix fescar server must start with port

yiailake:


skyesx:
agree with @sharajava ,  see #76

Issue #79, opened 2019-01-14
Expection fired when using server = "NATIVE" on Linux


tuohai666:
application.conf:

transport {
  #tcp udt unix-domain-socket
  type = "TCP"
  #NIO NATIVE
  **server = "NATIVE"**
  #thread factory for netty
  thread-factory {
    boss-thread-prefix = "NettyBoss"
    worker-thread-prefix = "NettyServerNIOWorker"
    server-executor-thread-prefix = "NettyServerBizHandler"
    share-boss-worker = false
    client-selector-thread-prefix = "NettyClientSelector"
    client-selector-thread-size = 1
    client-worker-thread-prefix = "NettyClientWorkerThread"
    # netty boss thread size,will not be used for UDT
    boss-thread-size = 1
    #auto default pin or 8
    worker-thread-size = 8
  }
}
service {
  #vgroup->rgroup
  vgroup_mapping.my_test_tx_group = "localRgroup"
  #only support single node
  localRgroup.grouplist = "127.0.0.1:8091"
  #degrade current not support
  enableDegrade = false
  #disable
  disable = false
}

client {
  async.commit.buffer.limit = 10000
  lock {
    retry.internal = 10
    retry.times = 30
  }
}

Expection fired when start application:

Exception in thread "main" java.lang.IllegalStateException: incompatible event loop type: io.netty.channel.nio.NioEventLoop
        at io.netty.channel.AbstractChannel$AbstractUnsafe.register(AbstractChannel.java:469)
        at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:80)
        at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:74)
        at io.netty.channel.MultithreadEventLoopGroup.register(MultithreadEventLoopGroup.java:86)
        at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:333)
        at io.netty.bootstrap.AbstractBootstrap.doBind(AbstractBootstrap.java:282)
        at io.netty.bootstrap.AbstractBootstrap.bind(AbstractBootstrap.java:278)
        at io.netty.bootstrap.AbstractBootstrap.bind(AbstractBootstrap.java:253)
        at com.alibaba.fescar.core.rpc.netty.AbstractRpcRemotingServer.start(AbstractRpcRemotingServer.java:143)
        at com.alibaba.fescar.core.rpc.netty.RpcServer.init(RpcServer.java:135)
[...]

PR #80, opened 2019-01-14
#79, Expection fired when using server = "NATIVE" on Linux

tuohai666:
#79, Expection fired when using server = "NATIVE" on Linux


sharajava:
The fix seems ok to me. Later will be reviewed by @slievrly

PR #91, opened 2019-01-14
bugfix: when no changes in localBranch will lead to exception in TC

skyesx:


rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/exec/AbstractDMLBaseExecutor.java
```diff
@@ -64,7 +69,12 @@ protected T executeAutoCommitTrue(Object[] args) throws Throwable {
                 }
             }
 
-        } finally {
+        } catch(Exception e) {
```
sharajava: what exception will be caught here?

skyesx: some runtimeException besides LockConflictException, just like timeout in connectionProxy.commit(), other failure in executeAutoCommitFalse(args)。

I add catch statement here because  some exception occur in finally statement connectionProxy.setAutoCommit(true) DID cover up the orignal exception

sharajava:
Let me see if we should ignore a branch when nothing changed. It's not friendly for transaction chain monitoring which may be an important feature in the future.

BTW, would you please change your code format template to follow Alibaba style?

@slievrly will release a contributing guide including code style.

skyesx:
You are right it's not friendly for tracing, and not register to RC will help little about performance because it rarely happen.

i will change the logic, let it register an empty key list but not a null list.

is this the Alibaba code style check plugin? https://github.com/alibaba/p3c

PR #98, opened 2019-01-15
#97, remove the redundant dependencies

tuohai666:
of netty-transport-native-epoll and netty-transport-native-kqueue. (#97 )


slievrly:
@tuohai666 Accept

PR #124, opened 2019-01-16
fix issue #122

### LINKED ISSUE #122: NettyBaseConfig#raiseUnsupportedTransportError() method internal String.format() usage error
`NettyBaseConfig#raiseUnsupportedTransportError` method use placeholder `{}` to format string, but `String.format()` method use like `%s` placeholder.

```java
private static void raiseUnsupportedTransportError() throws RuntimeException {
	String errMsg = String.format("Unsupported provider type :[{}] for transport:[{}].", TRANSPORT_SERVER_TYPE,
		TRANSPORT_PROTOCOL_TYPE);
	...
}
```

zffenger:
fixes #122


slievrly:
@zffenger why closed?

PR #125, opened 2019-01-16
use Integer.parseInt() instead of Integer.valueOf().intValue(), Long.…

lyennon:


sharajava:
Thanks. It's better to fire an issue before submit the PR. 👍

PR #159, opened 2019-01-17
examples demo show sql

systp:
Will debit, order, balance Sql print log in the console

Let developers be more convenient when debugging demos.Did not solve the bug


systp:
抱歉，请关闭请求，我稍后将重新提交

PR #160, opened 2019-01-17
Fix 3 problems of GlobalTransactionScanner class

tswstarplanet:
1.Fix the problem of duplicate assign new instance of field 'interceptor'
2.Remove the unnecessary type parameter
3.Let static field 'PROXYED_SET' just contains the names of object of class that has methods annotated by @GlobalTransactional

### DIFF
--- a/spring/src/main/java/com/alibaba/fescar/spring/annotation/GlobalTransactionScanner.java
+++ b/spring/src/main/java/com/alibaba/fescar/spring/annotation/GlobalTransactionScanner.java
@@ -180,7 +180,6 @@ public class GlobalTransactionScanner extends AbstractAutoProxyCreator implement
                 if (PROXYED_SET.contains(beanName)) {
                     return bean;
                 }
-                PROXYED_SET.add(beanName);
                 Class<?> serviceInterface = findTargetClass(bean);
                 Method[] methods = serviceInterface.getMethods();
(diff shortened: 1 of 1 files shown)


slievrly:
@tswstarplanet you should proposal an PR

slievrly:
@tswstarplanet please review f933140ab87d951df1f8253652cc949452609a2e

tswstarplanet:
@slievrly What do you mean ? Isn’t this a pr?

slievrly:
image

slievrly:
@tswstarplanet

slievrly:
can DingTalk @清铭

tswstarplanet:
Sorry I can’t find you with nickname “清铭” in dingtalk

slievrly:
image

slievrly:
@tswstarplanet

tswstarplanet:
But if we don’t move the line of code, it seems every bean wired by spring will be added to the set. Maybe we can move the line of code after the else clause?

slievrly:
> But if we don’t move the line of code, it seems every bean wired by spring will be added to the set. Maybe we can move the line of code after the else clause?

Just to judge whether it has been proxyed by fescar

tswstarplanet:
I hava pushed a commit again, would you please see if it is right now?

PR #168, opened 2019-01-18
Replace attrs [applicationId & transactionServiceGroup]  in branchSes…

### LINKED ISSUE #143: In the examples, when configuring different databases, an error is reported.
订单、余额、库存 3个表都放同一个数据库中运行成功，但当我分库后，运行出错。求大神告知哪里配置错了还是目前fescar就不支持分库。

ad201717:
…sion with branch`s, not the same with globalSession

When the branchSession is added, its property (applicationId and txServiceGroup) is set the same as globalSession. When TC do the commit job, it will retrive the RM channel with the same applicationId, then all BranchCommitRequest will sent to the same one RM. This may cause problems with commit or rollback when RMs use same ipAddress and different DB

fixes #143 

deploy the examples at one server with different DB, and run test.

### DIFF
--- a/core/src/main/java/com/alibaba/fescar/core/model/ResourceManagerOutbound.java
+++ b/core/src/main/java/com/alibaba/fescar/core/model/ResourceManagerOutbound.java
@@ -17,13 +17,14 @@
 package com.alibaba.fescar.core.model;
 
 import com.alibaba.fescar.core.exception.TransactionException;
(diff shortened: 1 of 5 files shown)


sharajava:
I think interfaces in com/alibaba/fescar/core/model/ package should NOT depend on RpcContext. 

Any way, it's great that you can share your PR. 👍 👍 👍

PR #176, opened 2019-01-18
Update SelectForUpdateExecutor.java

546621120:
if select query return no records, lockkey will by null, DefaultLockManagerImpl#isLockable will raise null point exception.

if select query return no records, no needs to check lock or not, currently DefaultLockManagerImpl#isLockable has no null check logic for lockkey value.

### DIFF
--- a/rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/exec/SelectForUpdateExecutor.java
+++ b/rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/exec/SelectForUpdateExecutor.java
@@ -87,7 +87,9 @@ public class SelectForUpdateExecutor<S extends Statement> extends BaseTransactio
                     }
 
                     TableRecords selectPKRows = TableRecords.buildRecords(getTableMeta(), rsPK);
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/exec/SelectForUpdateExecutor.java
```diff
@@ -87,7 +87,9 @@ public Object doExecute(Object... args) throws Throwable {
                     }
 
                     TableRecords selectPKRows = TableRecords.buildRecords(getTableMeta(), rsPK);
-                    statementProxy.getConnectionProxy().checkLock(selectPKRows);
+                    if(!selectPKRows.getRows().isEmpty()) {
```
sharajava: I would prefer to put such check inside com.alibaba.fescar.rm.datasource.ConnectionProxy#checkLock

BTW, it's better to fire an issue before PR. 👍

546621120: I see, thx!

PR #228, opened 2019-01-23
add toString method

xiajunhust:


slievrly:
@xiajunhust comment should use EN, please add UnitTest

xiajunhust:
> @xiajunhust comment should use EN, please add UnitTest

ok

slievrly:
@xiajunhust why closed？

PR #239, opened 2019-01-23
add unit test for some request class

xiajunhust:


slievrly:
@xiajunhust  PR propose on branch develop

PR #246, opened 2019-01-23
extract jdbc config on demo project

tigerMoon:
extract db config from xml.

run demo class

run demo class


slievrly:
@tigerMoon These three libraries should be logically independent, I think it should be similar to jdbc.order.url, jdbc.acount.url ...

tigerMoon:
> @tigerMoon These three libraries should be logically independent, I think it should be similar to jdbc.order.url, jdbc.acount.url ...

good idea

PR #258, opened 2019-01-23
fix #257 when timeout should break

### LINKED ISSUE #257: when timeout not return

### Ⅰ. Issue Description
when timeout can also go on 

### Ⅱ. Describe what happened

  
image

```
```

 it should throw exception or return.

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

github-ygy:
when timeout it should not go on ; b

fix #257 when timeout should break


github-ygy:
> @github-ygy why not use return but nest 「if...else」

pretty suggestion

slievrly:
@github-ygy Add some UnitTest based current conf？

PR #262, opened 2019-01-24
add unit test for server package LockManager and DefaultCore

xyzhimo:
add unit test for server package LockManager and DefaultCore


slievrly:
@xyzhimo review UnitTest case

PR #315, opened 2019-01-28
fix threadLocal variables leak problems of GlobalTransactionContext

tankilo:
call GlobalTransactionContext#clean in TransactionalTemplate#execute   if the  GlobalTransaction is not an existing transaction

https://github.com/alibaba/fescar/issues/313


tankilo:
image
Firstly, what does the above `self-requested a review` means? 
I thought you will review and edit this pull request yourself. So i did't do anythings.

I noticed that you correct this problem by yourself instead of merging mine.
https://github.com/alibaba/fescar/issues/325

But it seems that you ignore something that i contained in this pull request, your change will cause problems in nested transactions scene. I will comment on https://github.com/alibaba/fescar/issues/313 later.

slievrly:
it has been fixed

sharajava:
I just proposed another fix: https://github.com/alibaba/fescar/pull/354

Would you please join the review? I think GlobalTransaction instance storing in thread local seems not a necessary mechanism.

tankilo:
> I just proposed another fix: #354
> 
> Would you please join the review? I think GlobalTransaction instance storing in thread local seems not a necessary mechanism.

I've read it and didn't see any problem.

PR #320, opened 2019-01-29
change examples pom-packaging to pom and add springboot examples

sherlock-hao:
Try to split examples to examples-dubbo and examples-springboot and add springboot examples.
The examples-springboot requires java8


lovepoem:
close via https://github.com/fescar-group/fescar-samples/pull/2

PR #412, opened 2019-02-15
fixed issue #75,support BETWEEN in update statement.

### LINKED ISSUE #75: com.alibaba.druid.sql.ast.expr.SQLBetweenExpr cannot be cast to com.alibaba.druid.sql.ast.expr.SQLBinaryOpExpr
jdbcTemplate.update("update itest2 set x=x+1 where id between 9 and 11 ");

throw an exception "com.alibaba.druid.sql.ast.expr.SQLBetweenExpr cannot be cast to com.alibaba.druid.sql.ast.expr.SQLBinaryOpExpr"

"id" is primarykey for table of itest2 , it is not support for "between" syntax ? ?

xingfudeshi:
fixed issue #75,support BETWEEN in update statement.

fixes #75


xingfudeshi:
I referred to the MySQL manual about conditional statements，also looked at Druid's source code，i found that this problem only occurs with UPDATE statements.don't know if it is a bug of Druid.but it's the best solution for now.

PR #413, opened 2019-02-15
NettyClientConfig Class member variable named incorrectly

xuxiaofei4519:
Rename the member variable DEFAULT_POOL_FIFO in the NettyClientConfig class, Because the config object is actually assigned to a LIFO rather than a FIFO
Because it already exists


xuxiaofei4519:
Rename the member variable DEFAULT_POOL_FIFO in the NettyClientConfig class,

xuxiaofei4519:
This naming problem makes it easy for users to misconfigure

PR #434, opened 2019-02-20
add rollbackOn in GlobalTransactional

sherlock-hao:
add rollbackOn in GlobalTransactional

If rollbackOn was set and not empty, roll back on specified exception types


tm/src/main/java/com/alibaba/fescar/tm/api/TransactionalTemplate.java
```diff
@@ -75,10 +78,13 @@ public Object execute(TransactionalExecutor business) throws TransactionalExecut
 
         } catch (TransactionException txe) {
             // 4.1 Failed to commit
```
slievrly: when commit exx，biz exx will not be throw

PR #478, opened 2019-02-28
Feature registry redis

klboke:
add apollo configuration center and redis registry support


discovery/src/main/java/com/alibaba/fescar/discovery/registry/RedisRegistryServiceImpl.java
```diff
@@ -0,0 +1,234 @@
...
+    private ExecutorService threadPoolExecutor = new ScheduledThreadPoolExecutor(1,
+            new NamedThreadFactory("RedisRegistryService", 1));
+    private RedisRegistryServiceImpl() {
+        Configuration fescarConfig = ConfigurationFactory.getInstance();
+        this.clusterName = fescarConfig.getConfig(REDIS_FILEKEY_PREFIX + REGISTRY_CLUSTER_KEY, DEFAULT_CLUSTER);
+        String password = fescarConfig.getConfig(getRedisPasswordFileKey());
+        String serverAddr = fescarConfig.getConfig(getRedisAddrFileKey());
+        String[] serverArr = serverAddr.split(":");
+        String host = serverArr[0];
+        int port = Integer.valueOf(serverArr[1]);
```
slievrly: Integer.valueOf->Integer.parseInt

discovery/src/test/java/resources/apollo.properties
```diff
@@ -0,0 +1,5 @@
+#apollo config
```
slievrly: Is it necessary?

klboke: Non-essential, as a configuration reference example in apollo, similar to nacos-config.txt

PR #485, opened 2019-02-28
fixes #459.Avoid the problem of keywords as field names and table names

### LINKED ISSUE #459: rollback sql fix
the samples rollback generate rollback sql:
UPDATE t_asset_assign SET desc = ?, status = ?, update_time = ?, update_user = ? WHERE id = ?;
reason is that the parameter ‘desc ’ is mysql keyword, suggest to add `` char on rollback sql parameter at AbstractUndoExecutor.buildUndoSQL()

xingfudeshi:
1.Avoid the problem of keywords as field names and table names
2.fixes #459.

fixes #459


xingfudeshi:
So we need to check all keywords.

slievrly:
> So we need to check all keywords.



> So we need to check all keywords.

not only field name but table name

xingfudeshi:
Thanks for your review.i will do that.

rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/undo/mysql/MySQLUndoDeleteExecutor.java
```diff
@@ -42,6 +43,7 @@ public MySQLUndoDeleteExecutor(SQLUndoLog sqlUndoLog) {
 
     @Override
     protected String buildUndoSQL() {
+        MySQLKeywordChecker mySQLKeywordChecker=new MySQLKeywordChecker();
```
slievrly: why not KeywordChecker type？

xingfudeshi: Thanks for your carefully review,i will improve this by using a Factoryt ype.

slievrly:
@xingfudeshi please dingtalk ping me.

xingfudeshi:
> @xingfudeshi please dingtalk ping me.

ok

slievrly:
[测试过程.docx](https://github.com/alibaba/fescar/files/2929826/default.docx)

rm-datasource/src/main/java/com/alibaba/fescar/rm/datasource/undo/KeywordCheckerFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * get keyword checker
+     *
+     * @param dbType
+     * @return
+     */
+    public static KeywordChecker getKeywordChecker(String dbType) {
+        if (dbType.equals(JdbcConstants.MYSQL)) {
+            return new MySQLKeywordChecker();
```
slievrly: Singleton mode？

xingfudeshi: i will do it.

PR #551, opened 2019-03-07
Try to fix ConfigType NPE

sofay:
### Ⅰ. Try to fix ConfigType NPE


config/src/main/java/com/alibaba/fescar/config/ConfigurationFactory.java
```diff
@@ -46,14 +46,18 @@
...
-                FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
-                    + ConfigurationKeys.FILE_ROOT_TYPE));
+            configTypeName = FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
+                    + ConfigurationKeys.FILE_ROOT_TYPE);
+            configType = ConfigType.getType(configTypeName);
         } catch (Exception exx) {
             LOGGER.error(exx.getMessage());
         }
         Configuration configuration;
+        if (configType == null) {
```
slievrly: ==null -> ConfigType.Default ?

slievrly:
UT build failed

config/src/main/java/com/alibaba/fescar/config/ConfigType.java
```diff
@@ -45,14 +48,11 @@
...
-        } else if (Apollo.name().equalsIgnoreCase(name)) {
-            return Apollo;
-        }else {
-            throw new NotSupportYetException("unsupport type:" + name);
+        for (ConfigType configType : values()) {
+            if (configType.name().equalsIgnoreCase(name)) {
+                return configType;
+            }
         }
+        return Default;
```
jovany-wang: Let's use `throw Illegal Argument` instead of `return Default` here.

config/src/main/java/com/alibaba/fescar/config/ConfigurationFactory.java
jovany-wang: Also we should process the exception cases here.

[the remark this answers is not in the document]
sofay: if config a not exists` ConfigType` enum will throw an exception, then be catched and logged in `ConfigurationFactory` and go on, if no `if` block will throw `NullPointException` when `switch`

config/src/main/java/com/alibaba/fescar/config/ConfigType.java
```diff
@@ -45,14 +46,11 @@
...
-        } else if (Apollo.name().equalsIgnoreCase(name)) {
-            return Apollo;
-        }else {
-            throw new NotSupportYetException("unsupport type:" + name);
+        for (ConfigType configType : values()) {
+            if (configType.name().equalsIgnoreCase(name)) {
+                return configType;
+            }
         }
+        throw new NotSupportYetException("unsupport type:" + name);
```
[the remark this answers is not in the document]
sofay: just go back to what is was before...
i think both of them are ok

config/src/main/java/com/alibaba/fescar/config/ConfigurationFactory.java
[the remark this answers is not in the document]
jovany-wang: `getType ()` never return `null` right? 
```java
for (ConfigType configType : values()) {
     if (configType.name().equalsIgnoreCase(name)) {
             return configType;
    }
}
throw E;
```

config/src/main/java/com/alibaba/fescar/config/ConfigType.java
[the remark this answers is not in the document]
jovany-wang: The semantics of these exceptions are different a lot. It's better to use Illegal argument exception instead.

config/src/main/java/com/alibaba/fescar/config/ConfigurationFactory.java
```diff
@@ -46,14 +46,18 @@
...
+        String configTypeName = null;
         try {
-            configType = ConfigType.getType(
-                FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
-                    + ConfigurationKeys.FILE_ROOT_TYPE));
+            configTypeName = FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
+                    + ConfigurationKeys.FILE_ROOT_TYPE);
+            configType = ConfigType.getType(configTypeName);
         } catch (Exception exx) {
             LOGGER.error(exx.getMessage());
```
slievrly: move below if(configType == null){throw  xxx}  into above catch block  add throw exx？

sofay: fine, have done

config/src/main/java/com/alibaba/fescar/config/ConfigurationFactory.java
```diff
@@ -46,12 +46,14 @@
...
+            throw new NotSupportYetException("not support register type: " + configTypeName);
```
jovany-wang: If it throws an exception, why not remove the log here?

sofay: i have removed the log

PR #584, opened 2019-03-12
fix #576 when target is jdk-proxy, GlobalTransactional not function

### LINKED ISSUE #576: GlobalTransactionScanner can't find @GlobalTransactional
### Issue Description

If I use `<aop:aspectj-autoproxy />`, then the aspect and @GlobalTransactional point to the same method, which will cause the GlobalTransactionalInterceptor fail to get @GlobalTransactional

### Describe what happened

```java

public interface ITest {

    void execute();
}

public class ITestImpl implements ITest {

    @GlobalTransactional
    @Override
    public void execute() {
    }
}

@Aspect
public class MyAspect {

    @Before("execution(* com.xx.ITestImpl.execute(..))")
    public void before(){
    }
}

```

### Describe what you expected to happen
forcing the use of cglib can solve this problem, but what is the reason for building a new AutoProxyCreator (GlobalTransactionScanner)?Is it better to use spring built-in, e.g. @annotation(com.xxx.GlobalTransactional) directly? Thx

CoderEugene:

Fix #576

fixes #576

test on fescar-samples

1. enable `<aop:aspectj-autoproxy />`

2. create Aspect

   ```java
   @Aspect
   public class MyAspect {
   
       @Before("execution(* com.alibaba.fescar.samples.dubbo.service.impl.BusinessServiceImpl.purchase(..))")
       public void before(){
       }
   }
   ```

GlobalTransactionScanner can correctly detect @GlobalTransactional

### DIFF
--- a/spring/src/main/java/com/alibaba/fescar/spring/annotation/GlobalTransactionalInterceptor.java
+++ b/spring/src/main/java/com/alibaba/fescar/spring/annotation/GlobalTransactionalInterceptor.java
@@ -32,6 +32,9 @@ import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.aop.support.AopUtils;
+import org.springframework.core.BridgeMethodResolver;
+import org.springframework.util.ClassUtils;
 
 /**
  * The type Global transactional interceptor.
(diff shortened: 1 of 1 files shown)


spring/src/main/java/com/alibaba/fescar/spring/annotation/GlobalTransactionalInterceptor.java
```diff
@@ -59,9 +62,12 @@ public GlobalTransactionalInterceptor(FailureHandler failureHandler) {
...
     public Object invoke(final MethodInvocation methodInvocation) throws Throwable {
-        final GlobalTransactional globalTrxAnno = getAnnotation(methodInvocation.getMethod(),
-            GlobalTransactional.class);
-        final GlobalLock globalLockAnno = getAnnotation(methodInvocation.getMethod(), GlobalLock.class);
+        Class<?> targetClass = (methodInvocation.getThis() != null ? AopUtils.getTargetClass(methodInvocation.getThis()) : null);
+        Method specificMethod = ClassUtils.getMostSpecificMethod(methodInvocation.getMethod(), targetClass);
+        final Method method = BridgeMethodResolver.findBridgedMethod(specificMethod);
+
+        final GlobalTransactional globalTrxAnno = getAnnotation(method, GlobalTransactional.class);
+        final GlobalLock globalLockAnno = getAnnotation(method, GlobalLock.class);
```
jovany-wang: globalLockAnnotation

PR #602, opened 2019-03-14
 fix #600 arrange registry.conf 

### LINKED ISSUE #600: arrange registry.conf

### Ⅰ. Issue Description
arrange registry.conf 

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:
arrange registry.conf 

fix #600 arrange registry.conf


config/src/main/resources/registry.conf
```diff
@@ -1,12 +1,17 @@
 registry {
```
jovany-wang: Should we rename this to `default_refistry.conf`?

slievrly: It will be refactored in a future release, and if it is renamed now it will result in a low version incompatibility issue.

PR #654, opened 2019-03-26
add TC server metrics

slievrly:

add TC server metrics


slievrly:
do not format ，In case of impact review

PR #658, opened 2019-03-26
support metrics for tc server

zhengyangyong:

make tc support metrics and add default prometheus publisher

https://github.com/alibaba/fescar/issues/384

https://fescar.io/zh-cn/docs/ops/operation.html


metrics/metrics-api/src/main/java/com/alibaba/fescar/metrics/Constants.java
```diff
@@ -0,0 +1,61 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.alibaba.fescar.metrics;
+
+public class Constants {
```
ujjboy: same class name in `com.alibaba.fescar.common.Constants`, `MetricsConstants` is better in this module.

ujjboy:
@luyiisme  is the owner of sofa-lookout (multi-dimensional metrics refer to the Metrics 2.0 standard ). 

@luyiisme Please help review this pr

luyiisme:
I don't think it is a good idea to create  a metric api module in seata.  Seata is not a  framework or an application container.  I recommend  that Seata provide some extensions to expose these action which can change key statuses.

PR #660, opened 2019-03-26
fix: throw NPE when jdk-proxy has no target

XCXCXCXCX:
In the process of integrating spring-cloud-alibaba-stream-rocketmq with fescar, I throw a null pointer exception in the fescar-spring module's com.alibaba.fescar.spring.annotation.GlobalTransactionScanner. I found that com.alibaba was called. The .fescar.spring.util.SpringProxyUtils#findTargetClass method will recursively call target.getClass() when it gets the target object of the proxy class to generate a null pointer. The jdkProxy object generated by @EnableBinding has no target object, and there may be similar in the future. The scene, such as the proxy object generated by interface + annotation in Feign, also has no target object.
So I am modifying the com.alibaba.fescar.spring.util.SpringProxyUtils#findTargetClass part of the code:

`public static Class<?> findTargetClass(Object proxy) throws Exception {
        if (AopUtils.isAopProxy(proxy)) {
            AdvisedSupport advised = getAdvisedSupport(proxy);
            if(AopUtils.isJdkDynamicProxy(proxy)){
                TargetSource targetSource = advised.getTargetSource();
                return targetSource instanceof EmptyTargetSource ?  getFirstInterfaceByAdvised(advised): targetSource.getTarget().getClass();
            }
            Object target = advised.getTargetSource().getTarget();
            return findTargetClass(target);
        } else {
            return proxy.getClass();
        }
    }
    private static Class<?> getFirstInterfaceByAdvised(AdvisedSupport advised) {
        Class<?>[] interfaces = advised.getProxiedInterfaces();
        if(interfaces.length > 0){
            return interfaces[0];
        }else{
            throw new IllegalStateException("Find the jdk dynamic proxy class that does not implement the interface");
        }
    }`

When it is a jdk object, get its target object, if it can't get it, get the first interface it implements.
Ps. Perhaps you should think of a way to scan it to both (target object and interface)
Then my project is running normally.

Maybe I should mention an issue first.

NPE will be reported when integrating spring-cloud-stream and fescar

### DIFF
--- a/spring/src/main/java/com/alibaba/fescar/spring/util/SpringProxyUtils.java
+++ b/spring/src/main/java/com/alibaba/fescar/spring/util/SpringProxyUtils.java
@@ -16,8 +16,10 @@
 
 package com.alibaba.fescar.spring.util;
 
+import org.springframework.aop.TargetSource;
 import org.springframework.aop.framework.AdvisedSupport;
 import org.springframework.aop.support.AopUtils;
+import org.springframework.aop.target.EmptyTargetSource;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Proxy;
@@ -39,10 +41,23 @@ public class SpringProxyUtils {
     public static Class<?> findTargetClass(Object proxy) throws Exception {
         if (AopUtils.isAopProxy(proxy)) {
             AdvisedSupport advised = getAdvisedSupport(proxy);
+            if(AopUtils.isJdkDynamicProxy(proxy)){
+                TargetSource targetSource = advised.getTargetSource();
+                return targetSource instanceof EmptyTargetSource ?  getFirstInterfaceByAdvised(advised): targetSource.getTarget().getClass();
+            }
             Object target = advised.getTargetSource().getTarget();
             return findTargetClass(target);
         } else {
             return proxy.getClass();
+        }
+    }
+
+    private static Class<?> getFirstInterfaceByAdvised(AdvisedSupport advised) {
+        Class<?>[] interfaces = advised.getProxiedInterfaces();
+        if(interfaces.length > 0){
+            return interfaces[0];
+        }else{
+            throw new IllegalStateException("Find the jdk dynamic proxy class that does not implement the interface");
         }
     }
 


XCXCXCXCX:
fescar bug

slievrly:
@XCXCXCXCX thank you.

slievrly:
@XCXCXCXCX why not use @GlobalTransactional on Controller, Controller call feign service?

XCXCXCXCX:
For the situation you said, I have tested it separately before, and there is no problem when using fescar alone.
I am here to integrate spring-cloud-stream-rocketmq, @EnableBinding will create proxy objects for DemoSink and DemoSource, but fescar's GlobalTransactionScanner will scan all beans and determine if the bean should add fescar's interceptor, but the proxy object here is There is no target, so NPE is generated.

slievrly:
> For the situation you said, I have tested it separately before, and there is no problem when using fescar alone.
> I am here to integrate spring-cloud-stream-rocketmq, @EnableBinding will create proxy objects for DemoSink and DemoSource, but fescar's GlobalTransactionScanner will scan all beans and determine if the bean should add fescar's interceptor, but the proxy object here is There is no target, so NPE is generated.

can you dingTalk ping me?

XCXCXCXCX:
Ok, in fact this exception happens inside the SpringProxyUtils.findTargetClass() method, I modified some of its code to avoid exceptions.

spring/src/main/java/com/alibaba/fescar/spring/util/SpringProxyUtils.java
```diff
@@ -39,13 +41,26 @@
...
             return findTargetClass(target);
         } else {
             return proxy.getClass();
         }
     }
 
+    private static Class<?> getFirstInterfaceByAdvised(AdvisedSupport advised) {
+        Class<?>[] interfaces = advised.getProxiedInterfaces();
+        if(interfaces.length > 0){
+            return interfaces[0];
```
slievrly: if multi interface impl，@GlobalTransactional mark on interface[1]?

slievrly: will refactor in next version

PR #669, opened 2019-03-27
fix: RedisRegistry's problems

CoffeeLatte007:
1. modify the FileConfiguration in the RedisRegistry.Let it use registry.conf
2. fix uese redis registry, if restart or close the redis coudn't unregister the server.But still can't prevent kill -9.


xingfudeshi:
Thanks for the work.please sign the CLA first.so we can accept your PR.

PR #675, opened 2019-03-28
fix applicationIdMap may be null

guixincui:
- bugfix : applicationIdMap may be null


xingfudeshi:
Can you give a short explain how to re-produce this bug?

guixincui:
signed

guixincui:
> Can you give a short explain how to re-produce this bug?
I start then fescar-server  in '/Users/cuigx/tools/fescar', when debug fescar-samples, it generate a file 'root.data',  and mysql undo_log generate new rollback_info.
when fescar-server restart  it throws null pointer exception

xingfudeshi:
> > Can you give a short explain how to re-produce this bug?
> > Startup  fescar-server  in '/Users/cuigx/tools/fescar' and then debug fescar-samples, it generate a file 'root.data',  and mysql undo_log generate new rollback_info.
> > and then when fescar-server restart  it throws null pointer exception

ok,i will check.

PR #704, opened 2019-03-31
add: recycle buffer

CoffeeLatte007:

In the code, there are many bytebuffers created but not reused. To reduce the number of YGC, ByteBuffer is reused.

I first read netty's recycle and implemented it. But this is a big change and will make the byte[] that was previously returned by SessionStorable into one of our custom objects, so I dropped it.

Then I looked at other store's system, such as RocketMq, which use ThreadLocal for caching,.it is simple and has very little momentum.

I ran the write part of WriteStoreTest, the VM parameter was -Xmx10m, the YGC was 3,400 times before the modification, and the YGC was 900 times after the modification.

I also changed the HeapByteBuffer used by Filechannel to DirectHeapByteBuffer.

run write part of WriteStoreTest, VM parameter -Xmx10m
wathching jstat -gc pid 1000

### DIFF
--- a/server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
+++ b/server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
@@ -33,6 +33,8 @@ import com.alibaba.fescar.server.store.SessionStorable;
  */
 public class BranchSession implements Lockable, Comparable<BranchSession>, SessionStorable {
 
+    private static ThreadLocal<ByteBuffer> byteBufferThreadLocal = ThreadLocal.withInitial(() -> ByteBuffer.allocate(2048));
+
     private long transactionId;
 
     private long branchId;
@@ -268,7 +270,10 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
 
     @Override
     public byte[] encode() {
-        ByteBuffer byteBuffer = ByteBuffer.allocate(2048);
+        ByteBuffer byteBuffer = byteBufferThreadLocal.get();
+        //recycle
+        byteBuffer.clear();
+
         byteBuffer.putLong(transactionId);
         byteBuffer.putLong(branchId);
         if (null != resourceId) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
```diff
@@ -33,6 +33,8 @@
  */
 public class BranchSession implements Lockable, Comparable<BranchSession>, SessionStorable {
 
+    private static ThreadLocal<ByteBuffer> byteBufferThreadLocal = ThreadLocal.withInitial(() -> ByteBuffer.allocate(2048));
```
github-ygy: better to give the num（2048） staic value with a name

server/src/main/java/com/alibaba/fescar/server/session/GlobalSession.java
```diff
@@ -32,6 +32,8 @@
  */
 public class GlobalSession implements SessionLifecycle, SessionStorable {
 
+    private static ThreadLocal<ByteBuffer> byteBufferThreadLocal = ThreadLocal.withInitial(() -> ByteBuffer.allocate(512));
```
github-ygy: +1

server/src/main/java/com/alibaba/fescar/server/store/FileTransactionStoreManager.java
```diff
@@ -262,6 +262,8 @@ private void closeFile(RandomAccessFile raf) {
      */
     class WriteDataFileRunnable implements Runnable {
 
+        private ThreadLocal<ByteBuffer> writeBufferLocal = ThreadLocal.withInitial(() -> ByteBuffer.allocateDirect(4096));
```
github-ygy: +2

server/src/main/java/com/alibaba/fescar/server/store/FileTransactionStoreManager.java
```diff
@@ -299,8 +303,10 @@ public void run() {
 
         private boolean writeDataFile(byte[] bs) {
             int retry = 0;
-            byte[] byWrite = new byte[bs.length + 4];
-            ByteBuffer byteBuffer = ByteBuffer.wrap(byWrite);
+            ByteBuffer byteBuffer = wireteBuffer;
```
slievrly: if size > 4096?

PR #705, opened 2019-04-01
Supports service discovery with Consul

xingfudeshi:
Supports service discovery with Consul.

### DIFF
--- a/server/src/main/resources/registry.conf
+++ b/server/src/main/resources/registry.conf
@@ -1,5 +1,5 @@
 registry {
-  # file 、nacos 、eureka、redis、zk
+  # file 、nacos 、eureka、redis、zk、consul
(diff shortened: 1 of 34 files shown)


server/src/main/resources/registry.conf
```diff
@@ -1,6 +1,6 @@
 registry {
-  # file 、nacos 、eureka、redis、zk
-  type = "file"
+  # file 、nacos 、eureka、redis、zk、consul
+  type = "consul"
```
slievrly: default type= file

xingfudeshi: i will change it back.it was changed for testing.

xingfudeshi:
I will submit a new PR after I have sorted out the code.

PR #713, opened 2019-04-02
ADD: session size configuration

CoffeeLatte007:
1. Enable users to configure gloab session, branch session and write buffer size

2. For branch session, if it is too large, it will first try to compress lockkey, and if it is larger than the maximum, an exception will be thrown.

3. I want to do the compression of RPC, but the RPC protocol is not very friendly. If we want to do the body compression of RPC, we need to adjust the protocol design

### DIFF
--- a/server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
+++ b/server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
@@ -16,17 +16,24 @@
 
 package com.alibaba.fescar.server.session;
 
+import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.alibaba.fescar.common.util.CompressUtil;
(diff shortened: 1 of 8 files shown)


server/src/main/java/com/alibaba/fescar/server/store/StoreConfig.java
```diff
@@ -0,0 +1,41 @@
+/**
```
xingfudeshi: copyright

server/src/main/java/com/alibaba/fescar/server/session/BranchSession.java
```diff
@@ -35,9 +42,12 @@
  */
 public class BranchSession implements Lockable, Comparable<BranchSession>, SessionStorable {
 
-    public static final int DEFAULT_BRANCH_SESSION_BUFFER_SIZE = 2048;
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileTransactionStoreManager.class);
```
xingfudeshi: FileTransactionStoreManager.class?

xingfudeshi:
i left some comments

CoffeeLatte007:
> i left some comments

done

PR #744, opened 2019-04-08
FIX:fix ZK Register bugs

CoffeeLatte007:
this can replace the shutdownHook in the RedisRegistry.Another benefit is that the shutdown code is unified

1.fix ZK USES persistent nodes and does not support disconnected reconnections
2.A Chinese document has been modified


discovery/src/main/java/com/alibaba/fescar/discovery/registry/ZookeeperRegisterServiceImpl.java
```diff
@@ -81,7 +88,21 @@ public void register(InetSocketAddress address) throws Exception {
...
-        getClientInstance().createPersistent(path, true);
+        doRegister(path);
+    }
+
+    private boolean doRegister(String path) {
+        if (checkExists(path)) {
+            return false;
+        }
+        getClientInstance().createEphemeral(path, true);
+        REGISTERED_PATH_MAP.putIfAbsent(path, null);
```
slievrly: ConcurrentHashMap null value？

Issue #751, opened 2019-04-09
[Discussion]Add an eventbus mechanism for the status change notification of transactions


zhengyangyong:
## Why you need it?
As https://github.com/seata/seata/issues/384#issuecomment-480713071 mentioned,I think we can add an eventbus mechanism in order to make seata better scalability, then we can implement various metrics or other features based on it.

在https://github.com/seata/seata/issues/384#issuecomment-480713071中已经提到，我想我们可以增加一个事件机制使得seata具有更好的可扩展性，然后在之上实现metrics和其它新特性。

## How it could be?
The fast and simple way is use [EventBus](https://github.com/google/guava/tree/master/guava/src/com/google/common/eventbus) in [Google Guava ](https://github.com/google/guava) lib, after we can make a higher performance implementation by ourself.

最简单快速的方法是使用[谷歌 Guava](https://github.com/google/guava)库中的[EventBus](https://github.com/google/guava/tree/master/guava/src/com/google/common/eventbus)，未来我们也可以自己实现更简单更高效的EventBus机制。

This mechanism will add into all seata components include TC,TM and RM, it will raise event when the status of transaction changed like started, committed, failed or aborted, not only status is inclued in this events, also transaction name, start time, end time(for compute duration).

这个机制将会在所有的seata组件包括TC,TM和RM中添加，当事务的状态发生变化的时候例如开始、已提交、失败或终止就会触发，事件中不仅包含状态信息，还包含事务名，开始时间和结束时间（用于计算时延）。

## Other related information
https://github.com/seata/seata/issues/384

PR #768, opened 2019-04-11
#751 add event bus mechanism and apply it in tc

zhengyangyong:
Signed-off-by: zhengyangyong <yangyong.zheng@qq.com>

https://github.com/seata/seata/issues/751


server/src/test/java/com/alibaba/fescar/server/coordinator/TransactionEventListener.java
```diff
@@ -0,0 +1,63 @@
...
+  public AtomicInteger getCommitEventCount() {
+    return commitEventCount;
+  }
+
+  public AtomicInteger getRollbackEventCount() {
+    return rollbackEventCount;
+  }
+
+  public TransactionEventListener() {
+    this.startEventCount = new AtomicInteger(0);
```
CoffeeLatte007: magic value

core/src/main/java/com/alibaba/fescar/core/event/GuavaEventBus.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class GuavaEventBus implements EventBus {
+  private final com.google.common.eventbus.EventBus eventBus;
+
+  public GuavaEventBus(String identifier) {
+    this.eventBus = new com.google.common.eventbus.EventBus(identifier);
+  }
+
+  @Override
+  public void register(Object object) {
```
CoffeeLatte007: Why only the test code registers Listener

zhengyangyong: will add listener such as metrics later,not this pr

server/src/test/java/com/alibaba/fescar/server/coordinator/TransactionEventListener.java
zhengyangyong: I think may ok in test code ?

zhengyangyong:
> Is th pull request for monitoring ？There are two more points to consider:
> 1.server shutdown or close ,the data will be lost.Whether the data should be persisted and split by day?
> 2.Whether retrycommit, retryRollback and timeout need to monitored

1. because it is a event bus, I think persistence is not necessary
2. yes it is a good suggestion I will try add them

server/src/test/java/com/alibaba/fescar/server/coordinator/TransactionEventListener.java
CoffeeLatte007: Ok

core/src/main/java/com/alibaba/fescar/core/event/GuavaEventBus.java
CoffeeLatte007: ok

zhengyangyong:
1. Move raise start,commit and rollback event code from DefaultCoordinator to DefaultCore in order to ensure retrycommit and retryRollback monitored
2. Add timeout event in DefaultCoordinator

core/src/main/java/io/seata/core/event/GuavaEventBus.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class GuavaEventBus implements EventBus {
+  private final com.google.common.eventbus.EventBus eventBus;
+
+  public GuavaEventBus(String identifier) {
+    this.eventBus = new com.google.common.eventbus.EventBus(identifier);
+  }
+
+  @Override
+  public void register(Object object) {
```
leizhiyuan: maybe  Event should be a custom interface， not object， we can limit the type of event。

zhengyangyong: Done

leizhiyuan:
LGTM

xingfudeshi:
@zhengyangyong there are errors need to be solved.

server/src/main/java/io/seata/server/event/GlobalTransactionEvent.java
```diff
@@ -0,0 +1,91 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package io.seata.server.event;
+
+import io.seata.core.event.Event;
+import io.seata.core.model.GlobalStatus;
+
+public class GlobalTransactionEvent implements Event {
```
slievrly: Is it more appropriate to move to the core module?

server/src/test/resources/file.conf
```diff
@@ -0,0 +1,5 @@
+#reduce delay for test
```
slievrly: Is it necessary?

server/src/test/resources/registry.conf
```diff
@@ -0,0 +1,7 @@
+config {
```
slievrly: Is it necessary?
It seems that it is not consistent with the configuration file of the main project.

server/src/main/java/io/seata/server/event/GlobalTransactionEvent.java
zhengyangyong: yes,you are right, Done

server/src/test/resources/file.conf
zhengyangyong: yes ,this file is for test in order to reduce asyn-committing-retry-delay (default is 30 second) and timeout-retry-delay (default is 30 second) , then we can fast check event with `TimeoutRollbacking` status will come from core

server/src/test/resources/registry.conf
zhengyangyong: same as above

zhengyangyong:
All new file had add author info

PR #772, opened 2019-04-11
bugfix:fix FileConfiguration config listener logic 

### LINKED ISSUE #771: FileConfiguration couldn't notify the listeners

### Ⅱ. Describe what happened
```
public void run() {
            while (true) {
                try {
                    Map<String, List<ConfigChangeListener>> configListenerMap;
                    if (null != dataId && null != listener) {
                        configListenerMap = new ConcurrentHashMap<>(8);
                        configListenerMap.put(dataId, new ArrayList<>());
                        configListenerMap.get(dataId).add(listener);
                    } else {
                        configListenerMap = configListenersMap;
                    }
                    for (Map.Entry<String, List<ConfigChangeListener>> entry : configListenerMap.entrySet()) {
                        String configId = entry.getKey();
                        String currentConfig = getConfig(configId);
                        if (null != currentConfig && currentConfig.equals(listenedConfigMap.get(configId))) {
                            listenedConfigMap.put(configId, currentConfig);
                            notifyAllListener(configId, configListenerMap.get(configId));

                        } else if (null == currentConfig && null != listenedConfigMap.get(configId)) {
                            notifyAllListener(configId, configListenerMap.get(configId));
                        }
                    }
                    Thread.sleep(LISTENER_CONFIG_INTERNAL);
                } catch (Exception exx) {
                    LOGGER.error(exx.getMessage());
                }

            }
```
if the config with some configid is not null,  but the listenedConfigMap.get(configId) is null, the code listenedConfigMap.put(configId, currentConfig) will never run, so the method notfyAllListener will never be called

as the config value changed, the listener who watch these should always be notify
### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

liujianhuiouc:
when listeners change, add or remove the config correspond to the config id, as the config change, it mean current config different from the config in listenconfigmap, the listener match that config should be called

### DIFF
--- a/config/src/main/java/com/alibaba/fescar/config/FileConfiguration.java
+++ b/config/src/main/java/com/alibaba/fescar/config/FileConfiguration.java
@@ -31,6 +31,7 @@ import com.alibaba.fescar.config.ConfigFuture.ConfigOperation;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+import org.apache.commons.lang3.ObjectUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -123,6 +124,7 @@ public class FileConfiguration extends AbstractConfiguration<ConfigChangeListene
     public void addConfigListener(String dataId, ConfigChangeListener listener) {
         configListenersMap.putIfAbsent(dataId, new ArrayList<ConfigChangeListener>());
         configListenersMap.get(dataId).add(listener);
+        listenedConfigMap.putIfAbsent(dataId, getConfig(dataId));
         if (null != listener.getExecutor()) {
             ConfigChangeRunnable configChangeTask = new ConfigChangeRunnable(dataId, listener);
             listener.getExecutor().submit(configChangeTask);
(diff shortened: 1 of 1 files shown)


config/src/main/java/com/alibaba/fescar/config/FileConfiguration.java
```diff
@@ -123,6 +124,7 @@ public boolean removeConfig(String dataId, long timeoutMills) {
     public void addConfigListener(String dataId, ConfigChangeListener listener) {
         configListenersMap.putIfAbsent(dataId, new ArrayList<ConfigChangeListener>());
         configListenersMap.get(dataId).add(listener);
+        listenedConfigMap.putIfAbsent(dataId, getConfig(dataId));
```
CoffeeLatte007: maybe put?

CoffeeLatte007: if have many listeners and  this value has been updated. the data is old value

liujianhuiouc: if exists listener, the value in listenedconfigmap will be update in the ConfigChangeRunnable,

CoffeeLatte007: > if exists listener, the value in listenedconfigmap will be update in the ConfigChangeRunnable,

yes u are right

liujianhuiouc:
> Please modify the title.it will be a commit log.please refer others PR.

done

PR #779, opened 2019-04-12
release 0.4.2

slievrly:


lovepoem:
1. why so many commites ? 
image

2.  You can use  `reversion` to avoid changing version numbers for each subproject

PR #786, opened 2019-04-13
fix #785: GlobalSession thread Safety

### LINKED ISSUE #785: GlobalSession has thread safety issues

### Ⅰ. Issue Description
1. In the GlobalSession.add method, we use ArrayList which is not a thread safety collection.
GlobalSession.add is called by multiple threads，other method such as remove is called by single threads.
2. there's an extreme case,when excute add BranchSession,If GlobalSession already rollbck. The BranchSession will never excute rollback or commit
### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

CoffeeLatte007:
fix https://github.com/seata/seata/issues/785.

In the GlobalSession.add method, we use ArrayList which is not a thread safety collection.
GlobalSession.add is called by multiple threads，other method such as remove is called by single threads.

use spin lock, just lock status judgment and status change


xingfudeshi:
Please modify the title.it will be a commit log.please refer others PR.

server/src/main/java/com/alibaba/fescar/server/session/GlobalSession.java
```diff
@@ -467,4 +476,72 @@ public void decode(byte[] a) {
...
+     */
+    private class GlobalSessionSpinLock{
+        /**
+         * true: Can lock, false : in lock.
+         */
+        private AtomicBoolean globalSessionSpinLock = new AtomicBoolean(true);
+
+        private static final int PARK_TIMES_BASE = 10;
+
+        public void lock() {
```
XCXCXCXCX: Why not check the lock timeout?

CoffeeLatte007: I don't think it's necessary，Since the optimistic lock is used, the locking time must be very short

server/src/main/java/io/seata/server/coordinator/DefaultCore.java
```diff
@@ -135,19 +131,24 @@ public GlobalStatus commit(String xid) throws TransactionException {
...
-            if (globalSession.canBeCommittedAsync()) {
-                asyncCommit(globalSession);
-                return GlobalStatus.Committed;
-            } else {
-                doGlobalCommit(globalSession, false);
+        // just lock changeStatus
+        boolean shouldCommit = globalSession.lockAndExcute(() -> {
+            GlobalStatus status = globalSession.getStatus();
+            globalSession.closeAndClean(); // Highlight: Firstly, close the session, then no more branch can be registered.
+            if (status == GlobalStatus.Begin) {
```
github-ygy: remove GlobalStatus status = globalSession.getStatus(); use globalSession.getStatus() to equal

PR #798, opened 2019-04-16
 optimize the timeout settings of rpc, timeout exception occurs when high cocurrent

xiangyue520:
在进行dubbo调用时,经常会出现服务端调用超时回滚的现象

### II. Describe how to verify it
在server中的file.conf文件中,将rpc.request.timeout的值进行重新设置,再次进行并发调用即可,如果客户端需要进行一样的设置,一样需要在samples中的file.conf进行设置

ps:如果不能进行合并,请改进一下配置的超时时间,方便调用

### DIFF
--- a/core/src/main/java/com/alibaba/fescar/core/rpc/netty/NettyBaseConfig.java
+++ b/core/src/main/java/com/alibaba/fescar/core/rpc/netty/NettyBaseConfig.java
@@ -16,6 +16,10 @@
 
 package com.alibaba.fescar.core.rpc.netty;
 
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


xiangyue520:
> Service call timeout should be set in the rpc framework？
yeah,i think should be like this,but in seata not,it just set in com.alibaba.fescar.core.rpc.netty.NettyClientConfig.java and etc,use dubbo framework,just because my computer only can run this way.

zhangthen:
Change the title to English plz .

l81893521:
I m so sorry about that we do not want to open the configuration of the RPC_REQUEST_TIMEOUT for the user according to our discussion. Hope for you next contribution.

PR #824, opened 2019-04-18
add time limit when transaction retry on the server

XCXCXCXCX:
add time limit when transaction retry on the server 
Provide configuration items in file.conf :
  max.commit.retry.timeout=
  max.rollback.retry.timeout=
default permanent.

The retry timeout period is determined by the user. If it is not configured or configured as a negative number, it will run permanently; if configured to 0, it will be removed from the retry list on the first detection.


common/src/main/java/io/seata/common/util/DurationUtil.java
```diff
@@ -0,0 +1,73 @@
...
+            return null;
+        }
+        try {
+            return Long.parseLong(str);
+        } catch (NumberFormatException e) {
+            throw new UnsupportedOperationException("\"" + str + "\" can't parse to Duration", e);
+        }
+    }
+
+    public static void main(String[] args) {
```
leizhiyuan: remove the main method？

XCXCXCXCX: OK

zhangthen:
Plz add new config items to the nacos-config.txt too .

XCXCXCXCX:
> Plz add new config items to the nacos-config.txt too .

OK.

zhangthen:
@XCXCXCXCX 

It shoud be retry all the time until the rollback get success, this is the guarantee of final-consistency(BASE).
Stopping retries after a certain number of times may result in transactions not being rollbacked, which is contrary to the final consistency(BASE).

So is it necessary for this PR ?  
If the rollback is not necessary to be executed, then why do you need distributed transaction of Seata?

slievrly:
@XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.

zhangthen:
> @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.

If retry stop, how to make  the transaction to be final consistency ?

slievrly:
> > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> 
> If retry stop, how to make the transaction to be final consistency ?

If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.

zhangthen:
> > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > 
> > 
> > If retry stop, how to make the transaction to be final consistency ?
> 
> If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.

If retry stop, how to make the transaction to be final consistency ?

I think it shoud allways retry, even if database down or application fails , unless  there is a manual intervention to make the transaction to be final consistency， not  stop retry automatically, must stop by manual-intervention.

In addition , without artificial judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.

slievrly:
> > > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > > 
> > > 
> > > If retry stop, how to make the transaction to be final consistency ?
> > 
> > 
> > If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.
> 
> If retry stop, how to make the transaction to be final consistency ?
> 
> I think it shoud allways retry, even if database down or application fails , unless there is a manual intervention to make the transaction to be final consistency， not stop retry automatically, must stop by manual-intervention.
> 
> In addition , without manual judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.

This value defaults to permanent retry. When the user determines that this cannot be recovered because of a physical failure, the user can stop the retry by configuring this value.

CoffeeLatte007:
> > > > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > > > 
> > > > 
> > > > If retry stop, how to make the transaction to be final consistency ?
> > > 
> > > 
> > > If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.
> > 
> > 
> > If retry stop, how to make the transaction to be final consistency ?
> > I think it shoud allways retry, even if database down or application fails , unless there is a manual intervention to make the transaction to be final consistency， not stop retry automatically, must stop by manual-intervention.
> > In addition , without manual judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.
> 
> This value defaults to permanent retry. When the user determines that this cannot be recovered because of a physical failure, the user can stop the retry by configuring this value.

I have a suggest, the retry provides a configuration，it is up to the user to decide whether to retry permanently. 
If not permanent try again, then we can modify the retry mechanism, don't use Java ScheduledThreadPoolExecutor, we can use the Time Wheel, using an Exponential Backoff way retry

CoffeeLatte007:
> > > > > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > > > > 
> > > > > 
> > > > > If retry stop, how to make the transaction to be final consistency ?
> > > > 
> > > > 
> > > > If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.
> > > 
> > > 
> > > If retry stop, how to make the transaction to be final consistency ?
> > > I think it shoud allways retry, even if database down or application fails , unless there is a manual intervention to make the transaction to be final consistency， not stop retry automatically, must stop by manual-intervention.
> > > In addition , without manual judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.
> > 
> > 
> > This value defaults to permanent retry. When the user determines that this cannot be recovered because of a physical failure, the user can stop the retry by configuring this value.
> 
> I have a suggest, the retry provides a configuration，it is up to the user to decide whether to retry permanently.
> If not permanent try again, then we can modify the retry mechanism, don't use Java ScheduledThreadPoolExecutor, we can use the Time Wheel, using an Exponential Backoff way retry

zhangthen:
> > > > > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > > > > 
> > > > > 
> > > > > If retry stop, how to make the transaction to be final consistency ?
> > > > 
> > > > 
> > > > If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.
> > > 
> > > 
> > > If retry stop, how to make the transaction to be final consistency ?
> > > I think it shoud allways retry, even if database down or application fails , unless there is a manual intervention to make the transaction to be final consistency， not stop retry automatically, must stop by manual-intervention.
> > > In addition , without manual judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.
> > 
> > 
> > This value defaults to permanent retry. When the user determines that this cannot be recovered because of a physical failure, the user can stop the retry by configuring this value.
> 
> I have a suggest, the retry provides a configuration，it is up to the user to decide whether to retry permanently.
> If not permanent try again, then we can modify the retry mechanism, don't use Java ScheduledThreadPoolExecutor, we can use the Time Wheel, using an Exponential Backoff way retry

I think this implementation is too simple, will affect the rollback of all transactions；
The retry can not stop (unless stopped manually), which is the basic guarantee for the final consistency.
If too many failures, the retry frequency can be reduced appropriately.

CoffeeLatte007:
> > > > > > > @XCXCXCXCX @zhangthen This requirement is what I proposed. The default should be a permanent retry. When a database downtime or a single-point application fails to start, you need to stop retrying.
> > > > > > 
> > > > > > 
> > > > > > If retry stop, how to make the transaction to be final consistency ?
> > > > > 
> > > > > 
> > > > > If the physical failure of the database hangs, it is known that it cannot be started, and how to ensure consistency.
> > > > 
> > > > 
> > > > If retry stop, how to make the transaction to be final consistency ?
> > > > I think it shoud allways retry, even if database down or application fails , unless there is a manual intervention to make the transaction to be final consistency， not stop retry automatically, must stop by manual-intervention.
> > > > In addition , without manual judgment, how to know the cause of failure, it is too hasty to stop retrying automatically after several failures.
> > > 
> > > 
> > > This value defaults to permanent retry. When the user determines that this cannot be recovered because of a physical failure, the user can stop the retry by configuring this value.
> > 
> > 
> > I have a suggest, the retry provides a configuration，it is up to the user to decide whether to retry permanently.
> > If not permanent try again, then we can modify the retry mechanism, don't use Java ScheduledThreadPoolExecutor, we can use the Time Wheel, using an Exponential Backoff way retry
> 
> I think this implementation is too simple, will affect the rollback of all transactions；
> The retry can not stop (unless stopped manually), which is the basic guarantee for the final consistency.
> If too many failures, the retry frequency can be reduced appropriately.

agree

XCXCXCXCX:
@zhangthen @slievrly 
So I think we can stop its retry through a management entry. When the user determines that a transaction cannot be recovered due to a physical failure, it can be marked with a new failure status (eg. CANCEL) and removed from the retrying list. 
In addition, it is better to use Time Wheel for situations where there may be a lot of retry tasks.
@CoffeeLatte007

zhangthen:
> @zhangthen @slievrly
> So I think we can stop its retry through a management entry. When the user determines that a transaction cannot be recovered due to a physical failure, it can be marked with a new failure status (eg. CANCEL) and removed from the retrying list.
> In addition, it is better to use Time Wheel for situations where there may be a lot of retry tasks.
> @CoffeeLatte007

I agree this solution.

CoffeeLatte007:
> @zhangthen @slievrly
> So I think we can stop its retry through a management entry. When the user determines that a transaction cannot be recovered due to a physical failure, it can be marked with a new failure status (eg. CANCEL) and removed from the retrying list.
> In addition, it is better to use Time Wheel for situations where there may be a lot of retry tasks.
> @CoffeeLatte007

yes, i think use time wheel and  Exponential Backoff  is better for retry

server/src/main/resources/file.conf
```diff
@@ -29,6 +29,8 @@ service {
   enableDegrade = false
   #disable
   disable = false
+  max.commit.retry.timeout = "1d"
```
slievrly: how to config permanent？

config/seata-config-core/src/main/java/io/seata/config/AbstractConfiguration.java
```diff
@@ -61,6 +66,22 @@ public long getLong(String dataId) {
...
+        return getDuration(dataId, Duration.ZERO);
+    }
+
+    @Override
+    public Duration getDuration(String dataId, Duration defaultValue) {
+        return getDuration(dataId, defaultValue, DEFAULT_CONFIG_TIMEOUT);
+    }
+
+    @Override
+    public Duration getDuration(String dataId, Duration defaultValue, long timeoutMills) {
```
slievrly: if value=-1 parse error？

slievrly:
@zhangthen Currently configured for permanent retry, this modification entry is reserved for higher-order users. After the webconsole is perfected, the console UI needs to join the manual stop retry button.

server/src/main/resources/file.conf
```diff
@@ -29,6 +29,8 @@ service {
   enableDegrade = false
   #disable
   disable = false
+  max.commit.retry.timeout = "-1"
```
CoffeeLatte007: give example . because you use many unit such as m,s. Maybe they'll use min.

server/src/main/resources/file.conf
XCXCXCXCX: negative value or missing configuration

config/seata-config-core/src/main/java/io/seata/config/AbstractConfiguration.java
XCXCXCXCX: I have added no unit processing, now it works fine.

server/src/main/resources/file.conf
XCXCXCXCX: Yes, you are right.I added comments here to help users understand how to configure.

CoffeeLatte007: what is -1

PR #828, opened 2019-04-18
add RPCMessage protocol codec comment

leizhiyuan:

only add an rpc protocol format graph for users

no

no change code

please review it


leizhiyuan:
diff is not beautiful，please see

https://github.com/seata/seata/blob/653bb93560712d996bfacc283d7a45a2d223d441/core/src/main/java/io/seata/core/rpc/netty/MessageCodecHandler.java

leizhiyuan:
> is typecode occupy 2 bytes？

4 bytes, short type . I will remove '0 for hb ' comment

Issue #818, opened 2019-04-17
DataSourceProxy should be registered when it is created, not when it is getting a connection.


philoli11:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
DataSourceProxy should be registered when it is created, not when it is getting a connection.

### Ⅱ. Describe what happened
If there is no request coming in after the service is restarted, the branch rollback of the TC will always be retried.

PR #830, opened 2019-04-18
bugfix:RM delete delay registration

philoli11:
Delete the delayed registration of RM, and resolve the bug that the rollback retry fails after the service restarts.

#818


philoli11:
> check TableMetaTest

done

PR #831, opened 2019-04-18
An optimization about MessageFuture

XCXCXCXCX:
In the MessageFuture class, it is currently implemented using CountDownLatch.
This synchronization method is relatively heavy, we should replace it with CompletableFuture to achieve.
1. Will not change the MessageFuture API
2. Consider that the server will process a large number of Messages in the future, and the Get mode of the CompletableFuture will perform better when testing a large number of messages.


core/src/main/java/io/seata/core/protocol/MessageFuture.java
```diff
@@ -26,13 +29,12 @@
  * @author jimin.jm @alibaba-inc.com
  * @date 2018 /10/9
  */
-public class MessageFuture {
+public class MessageFuture{
```
CoffeeLatte007: { white space

github-ygy: add a space right the class

core/src/main/java/io/seata/core/protocol/MessageFuture.java
```diff
@@ -54,18 +56,22 @@ public boolean isTimeout() {
...
     public Object get(long timeout, TimeUnit unit) throws TimeoutException,
         InterruptedException {
-        boolean success = latch.await(timeout, unit);
-        if (!success) {
+        Object result = null;
+        try {
+            result = origin.get(timeout, unit);
+        } catch (ExecutionException e) {
+            throw new ShouldNeverHappenException("Should not get results in a multi-threaded environment", e);
+        } catch (TimeoutException e){
```
CoffeeLatte007: { white space

core/src/main/java/io/seata/core/protocol/MessageFuture.java
XCXCXCXCX: OK

core/src/main/java/io/seata/core/protocol/MessageFuture.java
XCXCXCXCX: OK

core/src/main/java/io/seata/core/protocol/MessageFuture.java
XCXCXCXCX: OK

core/src/main/java/io/seata/core/protocol/MessageFuture.java
```diff
@@ -74,8 +80,7 @@ public Object get(long timeout, TimeUnit unit) throws TimeoutException,
      * @param obj the obj
      */
     public void setResultMessage(Object obj) {
-        this.resultMessage = (obj == null ? NULL : obj);
-        latch.countDown();
+        origin.complete(obj == null ? NULL : obj);
```
github-ygy: how about remove the Null ？

github-ygy:
@CoffeeLatte007 do u  have any other questions ？

PR #834, opened 2019-04-19
fix bug for non SQLException not thrown

leizhiyuan:

Exception which not SQLException， should also be thrown

fixes https://github.com/seata/seata/issues/816

no need

cr


rm-datasource/src/main/java/io/seata/rm/datasource/exec/ExecuteTemplate.java
```diff
@@ -103,10 +103,10 @@
 
         } catch (Throwable ex) {
             if (ex instanceof SQLException) {
-                throw (SQLException)ex;
+                throw (SQLException) ex;
```
github-ygy: how about transfer the ex (non SQLException) to sqlException in the if{} , then remove the else and throw the ex directly ?

leizhiyuan: ok，done

PR #860, opened 2019-04-19
use state manager for globalsession status

leizhiyuan:

use state manager for global session status，operation drive status changing，not changing directly.
 

when you invoke changeStatus, you will be afraid of what you do, because maybe you will change the status to an incorrect status.

so we need to keep the correct ensure, according to the domain model, not the logic.

fixes https://github.com/seata/seata/issues/796

after your reviews, I will add some test case.

please review it, the state  transform was adjective, we can not know which status transform is a error, 

if ok,  I will add more test case, 

they are  very complex now, please review the `io.seata.server.session.GlobalSession#registerAcceptStatusGraph` to ensure the original logic .


server/src/main/java/io/seata/server/session/GlobalSession.java
```diff
@@ -58,6 +62,165 @@
 
     private ArrayList<BranchSession> branchSessions = new ArrayList<>();
 
+    protected StateGraph stateGraph = new StateGraph();
```
github-ygy: how ablout init it in static ？because it seems to apply to all GlobalSession,and  the stageGraph rule can't change dynamically.

leizhiyuan: ok，thanks for your advice。done

CoffeeLatte007:
What is the purpose of using a state manager?
just for check ?

leizhiyuan:
> What is the purpose of using a state manager?
> just for check ?

limit the state transform, in case of the error status change in the future.

and make it clear how seata changes the transaction status.

slievrly:
Why not use Squirrel if you are using a state machine?

leizhiyuan:
use squirrel to rewrite the machine status，if it is ok to us，we will use the operation to drive the status change ，not change target status  directly

@slievrly

a364176773:
Resolve code conflicts

funky-eyes:

PR #861, opened 2019-04-21
Success state after retrying is synchronized back to the client, and print log.

XCXCXCXCX:
At present, the client can only perceive the first commit or rollback failure. In fact, the transaction will be retried on the server at this time, so the client needs to be aware of the final state of the retry transaction.
At first I thought I needed to design the server push mode, and found that need to change a lot of code.
Eventually I used the client polling mode and added my logic to DefaultFailureHandlerImpl.


tm/src/test/java/io/seata/tm/api/DefaultFailureHandlerImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+                }
+
+                @Override
+                public String getXid() {
+                    return null;
+                }
+            }, null);
+        }
+        Thread.sleep(11000);
+        System.out.println("main end");
```
slievrly: remove system.out

XCXCXCXCX: OK

tm/src/test/java/io/seata/tm/api/DefaultFailureHandlerImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+                    return result;
+                }
+
+                @Override
+                public String getXid() {
+                    return null;
+                }
+            }, null);
+        }
+        Thread.sleep(11000);
```
xingfudeshi: TimeUnit sleep

zhangthen:
The retry number is unlimited ？ It's better to add a limit for retry number.

XCXCXCXCX:
> The retry number is unlimited ？ It's better to add a limit for retry number.

Good advice, I will modify it later.

tm/src/test/java/io/seata/tm/api/DefaultFailureHandlerImplTest.java
XCXCXCXCX: OK

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
```diff
@@ -16,18 +16,40 @@
...
+     */
+    private static final int RETRY_MAX_TIMES = 6 * 60;
+
+    private static final long SCHEDULE_INTERVAL_SECONDS = 10;
+
+    private static final long TICK_DURATION = 1;
+
+    private static final int TICKS_PER_WHEEL = 8;
+
+    private HashedWheelTimer timer = new HashedWheelTimer(
```
CoffeeLatte007: we need custom threadFactory

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
```diff
@@ -36,10 +58,54 @@ public void onBeginFailure(GlobalTransaction tx, Throwable cause) {
...
+
+        @Override
+        public void run(Timeout timeout) throws Exception {
+            if(!isStopped){
+                if(++count > RETRY_MAX_TIMES){
+                    LOGGER.error("transaction[" + tx.getXid() + "] retry fetch status times exceed the limit [" + RETRY_MAX_TIMES + " times]");
+                    return;
+                }
+                isStopped = shouldStop(tx, required);
+                timer.newTimeout(this, SCHEDULE_INTERVAL_SECONDS, TimeUnit.SECONDS);
```
CoffeeLatte007: can we use exponential backoff 10s,20s,40s.... ns

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
```diff
@@ -36,10 +58,54 @@ public void onBeginFailure(GlobalTransaction tx, Throwable cause) {
...
+            }
+        }
+    }
+
+    private boolean shouldStop(final GlobalTransaction tx, GlobalStatus required){
+        try {
+            GlobalStatus status = tx.getStatus();
+            LOGGER.warn("transaction[" + tx.getXid() + "] current status is [" + status + "]");
```
CoffeeLatte007: info

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
```diff
@@ -36,10 +58,54 @@ public void onBeginFailure(GlobalTransaction tx, Throwable cause) {
...
+        }
+
+        @Override
+        public void run(Timeout timeout) throws Exception {
+            if(!isStopped){
+                if(++count > RETRY_MAX_TIMES){
+                    LOGGER.error("transaction[" + tx.getXid() + "] retry fetch status times exceed the limit [" + RETRY_MAX_TIMES + " times]");
+                    return;
+                }
+                isStopped = shouldStop(tx, required);
```
CoffeeLatte007: the globalTransaction maybe should be update from server?

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
XCXCXCXCX: I think it is not necessary in such a scenario.

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
XCXCXCXCX: I will modify it later.

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
XCXCXCXCX: I don't understand what you mean...

tm/src/main/java/io/seata/tm/api/DefaultFailureHandlerImpl.java
XCXCXCXCX: yes, we should.

PR #869, opened 2019-04-23
Supports service discovery with Etcd

### LINKED ISSUE #812: Supports service discovery with Etcd
i will do this.

xingfudeshi:
Supports service discovery with Etcd

fixes #812 

i have done integration test.


discovery/seata-discovery-etcd3/src/main/java/io/seata/discovery/registery/etcd/EtcdRegistryServiceImpl.java
```diff
@@ -0,0 +1,403 @@
...
+        if (null != subscribeSet) {
+            Set<Watch.Listener> newSubscribeSet = new HashSet<>();
+            for (Watch.Listener eventListener : subscribeSet) {
+                if (!eventListener.equals(listener)) {
+                    newSubscribeSet.add(eventListener);
+                }
+            }
+            listenerMap.put(cluster, newSubscribeSet);
+        }
+        watcherMap.remove(cluster).stop();
```
slievrly: can not stop EtcdLifeKeeper#process？

xingfudeshi: > can not stop EtcdLifeKeeper#process？

EtcdLifeKeeper is used to refresh the lease which binding to registry key.like the name,it's for keeping the instance in healthy.I think when calling unsubscribe method, this means that the client no longer needs to get any updates from the registry, but the Seata server should still be running, so EtcdLifeKeeper should not be stopped.

discovery/seata-discovery-etcd3/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>27.0.1-jre</version>
```
github-ygy: how about init the vertion in parent pom ？

xingfudeshi: ok.it's done.

PR #877, opened 2019-04-23
bugfix:use queue instead of list to avoid concurrent modify exception

### LINKED ISSUE #803: ConcurrentModificationException in AsyncWork

### Ⅰ. Issue Description
ConcurrentModificationException will be throw in doBranchCommits, at the same time , some thread invoke branchCommit to add Phase2Context into ASYNC_COMMIT_BUFFER 

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

liujianhuiouc:
fix  #803


liujianhuiouc:
@zhangthen please check😊

PR #879, opened 2019-04-23
fix #878 close the statement

### LINKED ISSUE #878: statement should close

### Ⅰ. Issue Description
statement should close

### Ⅱ. Describe what happened

```
```
statement should close，when we use DruidPooledPreparedStatement we can use cache

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

github-ygy:
when over，close the statement，if we use DruidPooledPreparedStatement，so wo can use cache

fix #878


rm-datasource/src/main/java/io/seata/rm/datasource/undo/UndoLogManager.java
```diff
@@ -239,24 +238,34 @@ public static void undo(DataSourceProxy dataSourceProxy, String xid, long branch
...
+            deletePST = conn.prepareStatement(batchDeleteSql);
+            int paramsIndex = 1;
+            for (Long branchId : branchIds) {
+                deletePST.setLong(paramsIndex++,branchId);
+            }
+            for (String xid: xids){
+                deletePST.setString(paramsIndex++, xid);
+            }
+            int deleteRows = deletePST.executeUpdate();
+            if (LOGGER.isInfoEnabled()) {
```
leizhiyuan: maybe this log level should be debug?

github-ygy: thanks，done

PR #899, opened 2019-04-25
updateExecutor afterImage this select  for update  do without

xiaomin0322:

updateExecutor afterImage   select  sql .    for update  do without


zhangthen:
Replace of file,  which causes the modification of code to be invisible.

PR #903, opened 2019-04-25
UpdateExecutort afterImage  his select  for update  do without

xiaomin0322:
UpdateExecutort afterImage  his select  for update  do without


zhangthen:
Commit account (zhangzengmin) is not same with xiaomin0322

PR #904, opened 2019-04-25
bugfix:remove UpdateExecutort#afterImage for update

xiaomin0322:
UpdateExecutort afterImage his select for update do without


rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
```diff
@@ -140,8 +140,9 @@ protected TableRecords afterImage(TableRecords beforeImage) throws SQLException
             }
         }
         List<Field> pkRows = beforeImage.pkRows();
+       //TODO this select  for update  do without
```
slievrly: remove comment

PR #907, opened 2019-04-25
bugfix: when run server Module ServerTest failed, it occur "hostname can't be null" errors. see issue #743

### LINKED ISSUE #743: The server Module ServerTest failed, it occur "hostname can't be null" errors
- [x] I have searched the [issues](https://github.com/alibaba/fescar/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
I downloads the newest source code, get the dependencies, open the server module run the ServerTest.java , it has errors. After traceing the errors, I found the code execute on this line 
" RegistryFactory.getInstance().register(new InetSocketAddress(XID.getIpAddress(), XID.getPort()));" in AbstractRpcRemotingServer.java file, the error just happend.  It seem that XID have no values, so the XID.getIpAddress()'s call return null, then **CHECKHOST** method catch that and throw the exception.  SO I want to know  is there something wrong or I have some mistakes.
### Ⅱ. Describe what happened

```
```
2019-04-08 14:53:21.436 INFO [main]c.a.fescar.core.rpc.netty.AbstractRpcRemotingServer.start:153 -Server started ... 
Exception in thread "main" java.lang.RuntimeException: java.lang.IllegalArgumentException: hostname can't be null
	at com.alibaba.fescar.core.rpc.netty.AbstractRpcRemotingServer.start(AbstractRpcRemotingServer.java:158)
	at com.alibaba.fescar.core.rpc.netty.RpcServer.init(RpcServer.java:131)
	at ServerTest.main(ServerTest.java:46)
Caused by: java.lang.IllegalArgumentException: hostname can't be null
	at java.net.InetSocketAddress.checkHost(InetSocketAddress.java:149)
	at java.net.InetSocketAddress.<init>(InetSocketAddress.java:216)
	at com.alibaba.fescar.core.rpc.netty.AbstractRpcRemotingServer.start(AbstractRpcRemotingServer.java:154)
	... 2 more
2019-04-08 14:53:21.452 INFO [ShutdownHook]com.alibaba.fescar.core.rpc.netty.RpcServer.destroy:177 -destroyed rpcServer

1. just downloads the newest source code 
2. get the dependencies ready
3. open the server module, run the ServerTest.java in test directory , then you will see the exception

### Ⅵ. Environment:

- JDK version :jdk1.8.0_172
- OS : windows 10
- Others:

codewaltz1994:
FixBug #743 
When run server Module ServerTest failed, it occur "hostname can't be null" errors. see issue seata#743. According to logics in server.class just add XID set* ops into ServerTest class, then bug is free.

fixes #743 

Because this bug occur in a test class

just run ServerTest before and after this bugfix.


server/src/test/java/ServerTest.java
```diff
@@ -42,8 +43,9 @@ public static void main(String[] args) {
         RpcServer rpcServer = new RpcServer(workingThreads);
         rpcServer.setHandler(new DefaultCoordinator(rpcServer));
```
leizhiyuan: please change this class to a test case, not main method.

codewaltz1994: Hi, according your advice, I have done with it.  but in server module , the original authors's tests used the Main method, you can view the original code for this , so I want to know that we need to be consistent with him, or should just fix the bug. THX.

xingfudeshi: > please change this class to a test case, not main method.

for this situation,i thinks it's not necessary to use @Test.otherwise,the server will never be closed.

leizhiyuan: you can use @ after to shutdown the test server

xingfudeshi: after the server started,the main thread will be block, the @after will not get a chance to execute.

codewaltz1994: @xingfudeshi Then should  I change the @Test method back to Main method which the original author does, and we just focus on the bugfix?

xingfudeshi: yes.

codewaltz1994: @xingfudeshi It had already done.

codewaltz1994: @leizhiyuan Hi, Add @Test and @After method seem not suitable for this case, No as good as we focus on bugfix. So I change back into Main mothod as the original author does.

leizhiyuan:
LGTM

server/src/test/java/ServerTest.java
codewaltz1994: @leizhiyuan Hi, since we have common in changes after we discussed before, please approve these changes, not request changes. after this, changes can be merge into develop branch, I think. Thanks all

codewaltz1994:
Thanks all of you. It's my pleasure.

Issue #920, opened 2019-04-26
If the select for update query result set is empty, an exception occurs when getting the global lock


philoli11:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
If the select for update query result set is empty, an exception occurs when getting the global lock.

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
org.springframework.jdbc.UncategorizedSQLException: StatementCallback; uncategorized SQLException for SQL [select * from account_tbl where id='101' for update ]; SQL state [null]; error code [0]; io.seata.core.exception.TransactionException: Response[RuntimeException[null]]; nested exception is java.sql.SQLException: io.seata.core.exception.TransactionException: Response[RuntimeException[null]]
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:89)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1444)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:388)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:418)
	at com.alibaba.fescar.workshop.service.impl.OrderServiceImpl.create(OrderServiceImpl.java:93)
	at com.alibaba.dubbo.common.bytecode.Wrapper2.invokeMethod(Wrapper2.java)
	at com.alibaba.dubbo.rpc.proxy.javassist.JavassistProxyFactory$1.doInvoke(JavassistProxyFactory.java:47)
	at com.alibaba.dubbo.rpc.proxy.AbstractProxyInvoker.invoke(AbstractProxyInvoker.java:76)
	at com.alibaba.dubbo.config.invoker.DelegateProviderMetaDataInvoker.invoke(DelegateProviderMetaDataInvoker.java:52)
	at com.alibaba.dubbo.rpc.protocol.InvokerWrapper.invoke(InvokerWrapper.java:56)
[...]

PR #921, opened 2019-04-26
bugfix:select for update exception

philoli11:
fix select for update exception

#920


core/src/main/java/io/seata/core/protocol/MergeResultMessage.java
```diff
@@ -116,7 +116,11 @@ public void decode(ByteBuffer byteBuffer) {
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder("MergeResultMessage ");
-        for (AbstractMessage msg : msgs) { sb.append(msg.toString()).append("\n"); }
+        if (msgs != null) {
```
CoffeeLatte007: if(msg == null){
return sb.toString
}
for (AbstractMessage msg : msgs) {
                sb.append(msg.toString()).append("\n");
            }
       
        return sb.toString();

leizhiyuan: which sence? the msg is null ,and we need to return MergeResultMessage to client? maybe there are some bugs ?

rm-datasource/src/main/java/io/seata/rm/datasource/exec/SelectForUpdateExecutor.java
```diff
@@ -106,6 +106,9 @@ public T doExecute(Object... args) throws Throwable {
 
                     TableRecords selectPKRows = TableRecords.buildRecords(getTableMeta(), rsPK);
                     String lockKeys = buildLockKey(selectPKRows);
+                    if (lockKeys == null) {
```
CoffeeLatte007: StringUtils

philoli11: done

core/src/main/java/io/seata/core/protocol/MergeResultMessage.java
philoli11: I simulated the transfer sence, NPE will appear when testing with multiple threads.

PR #930, opened 2019-04-27
avoid some possible risk

KAMIJYOUDOUMA:
1.prevent users from changing constant value 
2.add some NPE check

### DIFF
--- a/common/src/main/java/io/seata/common/exception/FrameworkErrorCode.java
+++ b/common/src/main/java/io/seata/common/exception/FrameworkErrorCode.java
@@ -94,7 +94,8 @@ public enum FrameworkErrorCode {
     /**
      * 未定义错误
      */
-    UnknownAppError("10000", "unknown error", "内部错误"),;
+    UnknownAppError("10000", "unknown error", "内部错误"),
+    ;
 
     /**
      * The Err code.
--- a/common/src/main/java/io/seata/common/exception/FrameworkErrorCode.java
+++ b/common/src/main/java/io/seata/common/exception/FrameworkErrorCode.java
@@ -94,26 +94,37 @@ public enum FrameworkErrorCode {
     /**
      * 未定义错误
      */
-    UnknownAppError("10000", "unknown error", "内部错误"),
-    ;
+    UnknownAppError("10000", "unknown error", "内部错误"),;
 
     /**
      * The Err code.
      */
-    public String errCode;
+    private String errCode;
     /**
      * The Err message.
      */
-    public String errMessage;
+    private String errMessage;
     /**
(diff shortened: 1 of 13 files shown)


common/src/main/java/io/seata/common/exception/FrameworkErrorCode.java
```diff
@@ -94,28 +94,39 @@
     /**
      * 未定义错误
      */
-    UnknownAppError("10000", "unknown error", "内部错误"),
-    ;
+    UnknownAppError("10000", "unknown error", "内部错误"),;
```
CoffeeLatte007: ,
;

core/src/main/java/io/seata/core/rpc/netty/RmMessageListener.java
```diff
@@ -94,7 +94,10 @@ private void handleBranchCommit(long msgId, String serverAddress,
             resultMessage = (BranchCommitResponse)handler.onRequest(branchCommitRequest, null);
             sender.sendResponse(msgId, serverAddress, resultMessage);
         } catch (Exception e) {
-            LOGGER.error(FrameworkErrorCode.NetOnMessage.errCode, e.getMessage(), e);
+            LOGGER.error(FrameworkErrorCode.NetOnMessage.getErrCode(), e.getMessage(), e);
+            if(resultMessage == null){
```
CoffeeLatte007: maybe resultMessage already is not null

tcc/src/main/java/io/seata/rm/tcc/api/BusinessActionContext.java
```diff
@@ -68,7 +68,7 @@ public Object getActionContext(String key) {
      * @return the branch id
      */
     public long getBranchId() {
-        return branchId!=null?Long.valueOf(branchId):-1;
+        return branchId!=null?Long.parseLong(branchId):-1;
```
CoffeeLatte007: code format . branchId != null ? Long.parseLong(branchId) : -1;

core/src/main/java/io/seata/core/rpc/netty/RmMessageListener.java
KAMIJYOUDOUMA: > maybe resultMessage already is not null

`handler.onRequest` may throw a `RuntimeException`,then `resultMessage` will be null.

CoffeeLatte007: > > maybe resultMessage already is not null
> 
> `handler.onRequest` may throw a `RuntimeException`,then `resultMessage` will be null.

yes

Issue #936, opened 2019-04-29
TransactionManager should have a service load mechanism, just like ResourceManager


sharajava:
## Why you need it?
 Is your feature request related to a problem? Please describe in details

## How it could be?
A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

TransactionManager should have a service load mechanism, just like ResourceManager. 

Then, it'll be nice for other TM provider.

PR #938, opened 2019-04-29
issue #936 add service load mechanism for TM

sharajava:

issue #936 add service load mechanism for TM


tm/src/main/java/io/seata/tm/TransactionManagerHolder.java
```diff
@@ -0,0 +1,74 @@
...
+    private static class SingletonHolder {
+
+        private static TransactionManager INSTANCE = null;
+
+        static {
+            try {
+                INSTANCE = EnhancedServiceLoader.load(TransactionManager.class);
+                LOGGER.info("TransactionManager Singleton " + INSTANCE);
+            } catch (Throwable anyEx) {
+                LOGGER.error("Failed to load TransactionManager Singleton! ", anyEx);
```
slievrly: should throw exx?

CoffeeLatte007:
Who needs custom TransactionManager？Can you describe the usage scenarios？

sharajava:
> Who needs custom TransactionManager？Can you describe the usage scenarios？

TM and RM are core interfaces in Seata, I think it's better to take them forward to transaction standard, not just Java SDK. Implementations of TM and RM should be the endpoint for service providers of Seata standard. I hope this change will be the first step of making Seata as a standard.

Native Seata

Seata Service Provider

tm/src/main/java/io/seata/tm/TransactionManagerHolder.java
sharajava: Not sure, I was just considering that exception thrown from static block will cause some error of class initialization. 

Reference: https://stackoverflow.com/questions/2070293/why-doesnt-java-allow-to-throw-a-checked-exception-from-static-initialization-b

tm/src/main/java/io/seata/tm/TransactionManagerHolder.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    /**
+     * Get transaction manager.
+     *
+     * @return the transaction manager
+     */
+    public static TransactionManager get() {
+        return SingletonHolder.INSTANCE;
```
wgs13579: is there need NULL check ?

sharajava: Yes, I will add some check here.

tm/src/main/java/io/seata/tm/TransactionManagerHolder.java
CoffeeLatte007: maybe catch check exception EnhancedServiceNotFoundException. other excepiton should throw?

PR #939, opened 2019-04-29
fix doc

IanCao:

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Please follow the [template](https://github.com/seata/seata/blob/develop/.github
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


IanCao:
> please related your github ID

I cant get your point -_-

slievrly:
image
@IanCao

xingfudeshi:
git config --global user.name "username"
git config --global user.email "email"
use the above commands to config your username and email，then these two informations will be attached with every commit of you,so we can valid the source of commit.remember that the username and email should be exactly same as what they are in your github account.

IanCao:
> git config --global user.name "username"
> git config --global user.email "email"
> use the above commands to config your username and email，then these two informations will be attached with every commit of you,so we can valid the source of commit.remember that the username and email should be exactly same as what they are in your github account.

ok,i know.
i'll changed caoyixiong to IanCao and open a new pr.

PR #982, opened 2019-05-06
Fix conflict

leizhiyuan:


leizhiyuan:
https://github.com/seata/seata/pull/910

PR #986, opened 2019-05-07
Suggestions for tuning the tm-api TransactionalTemplate and GlobalLockTemplate template classes.

lynk-coder:
The TransactionalTemplate and GlobalLockTemplate template classes have been modified to make all methods "pubulic static", making it easier to directly call methods in the "tm-api" template during project development.


leizhiyuan:
maybe.. 

I do not think it is a good way to use so many static and public methods

tm/src/main/java/io/seata/tm/api/TransactionalTemplate.java
```diff
@@ -97,39 +97,39 @@ private void completeTransactionAfterThrowing(TransactionInfo txInfo, GlobalTran
         }
     }
 
-    private void commitTransaction(GlobalTransaction tx) throws TransactionalExecutor.ExecutionException {
+    public static void commitTransaction(GlobalTransaction tx) throws TransactionalExecutor.ExecutionException {
```
slievrly: why public ？not use template#execute?

lynk-coder: Because I found that many projects use block method to execute SQL, it is unnecessary to copy the methods in the void commitTransaction again, instead of directly calling the methods of TransactionalTemplate.

lynk-coder: Sorry, maybe I misread this code.

funky-eyes:

PR #1047, opened 2019-05-15
bugfix:initialized id,may cause the transaction to fail

### LINKED ISSUE #1046: When the maximum id of recovery is greater than the currently initialized id, it may cause the transaction to fail.

### Ⅰ. Issue Description

When the maximum id of recovery is greater than the currently initialized id, it may cause the transaction to fail.
1.  currently tid or bid is  123456 and transaction not completed.
2.  server down recover from file，UUIDGenerator init id as 100000，retrying transaction（123456）rollbacking but not successful.
3.  accept client transaction request, UUIDGenerator incrementAndGet, when the new bid is 123456, the transaction will fail because the undolog has a primary key conflict

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:

initialized id，may cause the transaction to fail

fix #1046

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -346,6 +346,11 @@
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.alipay.sofa</groupId>
+            <artifactId>sofa-rpc-all</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -111,7 +111,6 @@
                 <groupId>ch.qos.logback</groupId>
                 <artifactId>logback-classic</artifactId>
                 <version>${logback-classic.version}</version>
-                <scope>provided</scope>
             </dependency>
             <dependency>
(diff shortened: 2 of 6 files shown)


server/src/main/java/io/seata/server/session/FileBasedSessionManager.java
```diff
@@ -194,9 +197,30 @@ private void restore(List<TransactionWriteStore> stores, Map<Long, BranchSession
...
+        if (sessionStorable instanceof GlobalSession) {
+            currentId = ((GlobalSession)sessionStorable).getTransactionId();
+        } else if (sessionStorable instanceof BranchSession) {
+            currentId = ((BranchSession)sessionStorable).getBranchId();
+        }
+
+        return maxRecoverId > currentId ? maxRecoverId : currentId;
+    }
+
+    private void setMaxId(long maxRecoverId) {
```
CoffeeLatte007: getMaxId() contains ` maxRecoverId > currentId ? maxRecoverId : currentId` and setMaxId is single thread
maybe setMaxId should direct  set not use while or if

slievrly: fixed

PR #1048, opened 2019-05-15
RegisterTMRequest unit test class

### LINKED ISSUE #189: Add unit test to fescar-core module
Add unit test to fescar-core module

Danaykap:
This pr enhances the test coverage. The class added tests RegisterTMRequest class

fixes #189


core/src/test/java/io/seata/core/protocol/RegisterTMRequestTest.java
```diff
@@ -0,0 +1,49 @@
+package io.seata.core.protocol;
+
+import org.junit.jupiter.api.*;
```
xingfudeshi: please import explicit class type.

core/src/test/java/io/seata/core/protocol/RegisterTMRequestTest.java
```diff
@@ -0,0 +1,49 @@
+package io.seata.core.protocol;
```
xingfudeshi: please add copyright header.

core/src/test/java/io/seata/core/protocol/RegisterTMRequestTest.java
```diff
@@ -0,0 +1,49 @@
...
+	/** Constructor with arguments **/
+	@BeforeAll
+	public static void setupWithValues() {
+		registerTMRequest = new RegisterTMRequest(APP_ID, TSG, ED);
+	}
+
+	/** Test get type code **/
+	@Test
+	public void testGetTypeCode() {
+		Assertions.assertEquals(101, registerTMRequest.getTypeCode());
```
xingfudeshi: please define a constant for the value "101"

kaitithoma:
> Can you add test cases for other methods of this class? Such as encode and decode methods

We could try it! Should we create a new pr for this purpose?

xingfudeshi:
> > Can you add test cases for other methods of this class? Such as encode and decode methods
> 
> We could try it! Should we create a new pr for this purpose?

it's better to add the other tests of this class in the same PR.

kaitithoma:
> > > Can you add test cases for other methods of this class? Such as encode and decode methods
> > 
> > 
> > We could try it! Should we create a new pr for this purpose?
> 
> it's better to add the other tests of this class in the same PR.

encode and decode methods are in AbstractIdentifyRequest class but we intended to test RegisterTMRequest which extends it. Aren' t they supposed to be different classes?

xingfudeshi:
Each test class should correspond to a concrete class. RegisterTMRequest, although it inherits from AbstractIdentifyRequest, overrides the decode and encode methods, so the unit test case here is only for RegisterTMRequest.

kaitithoma:
We added test methods for decoding. Sorry if there are unnecessary code lines and mistakes but we aren' t very familiar to ByteBuf variables. Thank you!

xingfudeshi:
No worries.we can help you improve the code.

Issue #1049, opened 2019-05-15
Optimize the business logic of the undo method


ujjboy:
## How it could be?

When running `AbstractUndoExecutor#executeOn`, There are several optimization logic codes that can be added:

- [x] compare beforeImage and afterImage, If they are equal, then you don't need to execute undo
- [x] execute `dataValidation()`(PS: the method is empty now.), compare current data in the database with destination image, if they are not equal, it means have dirty data, will throw an exception
- [x] Compare current data in the database and the original image, if they are equal, also no need to execute undo
- [ ] Add a switch for this feature.

leizhiyuan:
has this been finished?

ujjboy:
No, last one pr is adding a switch for this feature.

PR #1051, opened 2019-05-15
Break undo if data snapshot is not changed.

ujjboy:

- Add DataCompareUtils
- Add some test case.
- Do not execute `undo` if the before data snapshot is equivalent to the after data snapshot.

Part of #1049

### DIFF
--- a/common/src/main/java/io/seata/common/util/StringUtils.java
+++ b/common/src/main/java/io/seata/common/util/StringUtils.java
@@ -32,6 +32,7 @@ import javax.sql.rowset.serial.SerialBlob;
  * The type String utils.
  *
  * @author jimin.jm @alibaba-inc.com
+ * @author Geng Zhang
  */
 public class StringUtils {
 
@@ -45,7 +46,7 @@ public class StringUtils {
      * @param str the str
      * @return the boolean
      */
-    public static final boolean isNullOrEmpty(String str) {
+    public static boolean isNullOrEmpty(String str) {
         return (str == null) || (str.isEmpty());
     }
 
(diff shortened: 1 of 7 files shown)


common/src/main/java/io/seata/common/util/StringUtils.java
```diff
@@ -196,4 +197,32 @@ public static String toString(Object obj){
...
     }
+
+    /**
+     * Equals boolean.
+     *
+     * @param a the a
+     * @param b the b
+     * @return boolean
+     */
+    public static boolean equals(String a, String b) {
```
xingfudeshi: What if the string is a blank character?should it be considered?

xingfudeshi: like a="    "

ujjboy: Do you mean  " "  and "" ?   I think there are not equals, maybe add new method `equalsIngoreBlank`

xingfudeshi: yes

xingfudeshi: > Do you mean " " and "" ? I think there are not equals, maybe add new method `equalsIngoreBlank`

can you add it in this PR,please?

ujjboy: I will add it when we need it.

xingfudeshi: > I will add it when we need it.

sure

PR #1060, opened 2019-05-17
Do data validation when undo.

ujjboy:

- Add logic:
  - compare current data in the database with destination image, if they are not equal, it means have dirty data,
  - then compare current data in the database and the original image, if they are equal, also no need to execute undo
  - else  will throw an exception
- Add some test case

Part of #1049


ujjboy:
I will add more test case which uses h2 database later.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/AbstractUndoExecutor.java
```diff
@@ -145,9 +164,115 @@ protected void undoPrepare(PreparedStatement undoPST, ArrayList<Field> undoValue
      * Data validation.
      *
      * @param conn the conn
-     * @throws SQLException the sql exception
+     * @return return true if data validation is ok and need continue
+     * @throws SQLException the sql exception such as has dirty data
      */
-    protected void dataValidation(Connection conn) throws SQLException {
+    protected boolean dataValidation(Connection conn) throws SQLException {
+
```
zhangthen: It ‘s better to add a switcher to close  the dirty data check manually, the switcher is open default.

rm-datasource/src/test/java/io/seata/rm/datasource/undo/UndoExecutorTest.java
```diff
@@ -287,9 +294,12 @@ public void testDelete() {
...
-
+        MockConnection connection = new MockConnection();
+        AbstractUndoExecutor spy = Mockito.spy(executor);
         try {
-            executor.executeOn(new MockConnection());
+            // skip data validation
+            Mockito.doReturn(true).when(spy).dataValidation(connection);
+            spy.executeOn(connection);
         } catch (SQLException e) {
             e.printStackTrace();
```
leizhiyuan: don't use e.printTrace()

rm-datasource/src/test/java/io/seata/rm/datasource/undo/UndoExecutorTest.java
```diff
@@ -217,9 +221,12 @@ public void testInsert() {
...
```
leizhiyuan: don't use e.printTrace()

rm-datasource/src/main/java/io/seata/rm/datasource/undo/AbstractUndoExecutor.java
ujjboy: OK, I will add it in the next standalone PR.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/AbstractUndoExecutor.java
```diff
@@ -72,11 +91,11 @@ public SQLUndoLog getSqlUndoLog() {
      */
     public void executeOn(Connection conn) throws SQLException {
 
-        // no need undo if the before data snapshot is equivalent to the after data snapshot.
-        if (DataCompareUtils.isRecordsEquals(sqlUndoLog.getBeforeImage(), sqlUndoLog.getAfterImage())) {
+        // when the data validation is not ok 
+        if (!dataValidation(conn)) {
             return;
```
leizhiyuan: should we throw some exception explicitly for the code change later？

ujjboy: Sorry, I don’t get your point.

ujjboy: I add some log and change the method name for more explicitly.

leizhiyuan: as communicated offline

ujjboy: @leizhiyuan   done.

PR #1075, opened 2019-05-20
Multiple environmental isolation

XCXCXCXCX:
Switch environments by specifying env environment variables, let seata use different environment configurations.

example:
past
+file.conf
+registry.conf

now
+file-dev.conf
+registry-dev.conf
+file.conf
+registry.conf

Use -Denv=dev to recognize the suffix, and choose registry-dev.conf configuration.
when -Denv=default or null, use registry.conf.
when -Denv=unknown, use NULL configuration, at this point some objects will report an error due to missing configuration.


xingfudeshi:
Good!

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -30,11 +30,14 @@
...
 public final class ConfigurationFactory {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigurationFactory.class);
-    private static final String REGISTRY_CONF = "registry.conf";
+    private static final String REGISTRY_CONF_PREFIX = "registry";
+    private static final String REGISTRY_CONF_SUFFIX = ".conf";
     /**
      * The constant FILE_INSTANCE.
      */
-    public static final Configuration FILE_INSTANCE = new FileConfiguration(REGISTRY_CONF);
+    private static final String ENV_VALUE = System.getProperty("env");
```
leizhiyuan: maybe you can provide a util class to set or resolve the env value
default it can fetch from system properties
when users use spring ，maybe they want to config from the application.properties ，so they can invoke your util method  instead of  use `System.setProperty` or pass by `-D`

xingfudeshi:
@leizhiyuan To read configs from application,it's better to create a module separately to do that.like seata-spring-boot-starter.

leizhiyuan:
> @leizhiyuan To read configs from application,it's better to create a module separately to do that.like seata-spring-boot-starter.

Yes， What I mean is  if we write like this ，we need to invoke  `System.setxxx` in  seata-spring-boot-starter, It is arbitrary..

xingfudeshi:
@leizhiyuan in spring-boot starter ,we  can provide another way to read configurations.
use the benefit of profile to do the configuration isolation,This means that the client can discard the file.conf file entirely.

xingfudeshi:
@leizhiyuan We provide full default values ，user can override them in the application files.

leizhiyuan:
only available in server，ok

xingfudeshi:
This kind of configuration isolation is acceptable on the server side, but may not be so good on the client side, such as spring-boot.

xingfudeshi:
I think an environment variable can be added. The name is SEATA_CONFIG_ENV. Before loading the configuration file, both the server and the client should read the environment variable of the current System (using system.getenv).If it's not blank then load the corresponding configuration files, such as registry-dev.conf and file-dev.conf, where dev is the value read from the environment variable.

XCXCXCXCX:
> I think an environment variable can be added. The name is SEATA_CONFIG_ENV. Before loading the configuration file, both the server and the client should read the environment variable of the current System (using system.getenv).If it's not blank then load the corresponding configuration files, such as registry-dev.conf and file-dev.conf, where dev is the value read from the environment variable.

Good idea.

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -28,10 +29,26 @@
...
+    private static final String ENV_SYSTEM_KEY = "SEATA_CONFIG_ENV";
+    private static final String ENV_PROPERTY_KEY = "env";
     /**
      * The constant FILE_INSTANCE.
      */
-    public static final Configuration FILE_INSTANCE = new FileConfiguration(REGISTRY_CONF);
+    private static String ENV_VALUE;
+    static {
+        Map<String, String> map = System.getenv();
+        String env = map.get(ENV_SYSTEM_KEY);
```
xingfudeshi: You can use System.getEnv(ENV_SYSTEM_KEY);

XCXCXCXCX: OK

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -28,10 +28,25 @@
...
+        String env = System.getenv(ENV_SYSTEM_KEY);
+        if(env != null && System.getProperty(ENV_PROPERTY_KEY) == null){
+            //Help users get
+            System.setProperty(ENV_PROPERTY_KEY, env);
+        }
+        ENV_VALUE = System.getProperty(ENV_PROPERTY_KEY);
+    }
+
+    private static final Configuration DEFAULT_FILE_INSTANCE = new FileConfiguration(REGISTRY_CONF_PREFIX + REGISTRY_CONF_SUFFIX);
+    public static final Configuration CURRENT_FILE_INSTANCE = (ENV_VALUE == null || "default".equals(ENV_VALUE)) ? DEFAULT_FILE_INSTANCE : new FileConfiguration(REGISTRY_CONF_PREFIX + "-" + ENV_VALUE + REGISTRY_CONF_SUFFIX);
```
xingfudeshi: it's better to define a constant for "default".

a364176773:
由于前期Seata社区治理规范问题部分代码作者未签署CLA，可能引发社区知识产权风险问题。请所有在Seata社区贡献过代码（包含：主项目、官网、samples和多语言项目等）的 contributor 帮忙在这个链接登录github账号签署相应的开发者CLA，https://cla-assistant.io/seata/seata  。2023.1.31 未签署CLA的代码将会被rewrite，拜托大家帮忙签一下。

PR #1083, opened 2019-05-21
Add switch of undo data validation.

### LINKED ISSUE #1049: Optimize the business logic of the undo method

ujjboy:

- Add switch of undo data validation.

Last PR of #1049 

Fix #1049

### DIFF
--- a/config/seata-config-core/src/main/resources/file.conf
+++ b/config/seata-config-core/src/main/resources/file.conf
@@ -70,3 +70,7 @@ client {
   }
   report.retry.count = 5
 }
+
+transaction {
+  undo.data.validation = true
+}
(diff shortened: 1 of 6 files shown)


ujjboy:
@slievrly OK,  I already added.

PR #1094, opened 2019-05-23
enable to set config file base dir by jvm param or System.setProperty

wuzunqian:
enable to set config file base dir by System.setProperty

no

simple change

1.start app with jvm param: -Dseata.config.basedir=/xxx
2.put config file (like registry.conf) in dir /xxx
3.verify if config file effect


config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
```diff
@@ -77,17 +79,23 @@ public FileConfiguration() {
      * @param name the name
      */
     public FileConfiguration(String name) {
+        String baseDir = System.getProperty(CONFIG_BASEDIR);
         if (null == name) {
             CONFIG = ConfigFactory.load();
-        } else {
+        } else if(baseDir == null || baseDir.length() == 0){
```
xingfudeshi: StringUtils.isNullOrEmpty

xingfudeshi:
Thanks for your contribution!I have a doubt,are these changes used for configuration isolation in different environments?

wuzunqian:
> Thanks for your contribution!I have a doubt,are these changes used for configuration isolation in different environments?

no,i need to load config file from path not only in jar/war, isolation of different enviroments can not satisfy me.

xingfudeshi:
> > Thanks for your contribution!I have a doubt,are these changes used for configuration isolation in different environments?
> 
> no,i need to load config file from path not only in jar/war, isolation of different enviroments can not satisfy me.

Could you please add some test cases to verify this function?

a364176773:
resolve code conflicts

PR #1095, opened 2019-05-23
Undo pr 1033 and adjust the maven plugin of seata-all.xml

zhangthen:
Change points:

- Undo the pr #1033 

Set the createDependencyReducedPom to true.

The code of seata-* has all be packed into the seata-all.jar, and should not be inclueded in the seata-all.pom.

- Modify the all.xml

Change  the  configs of maven-javadoc-plugin in seata-all.xml

### DIFF
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -423,7 +423,7 @@
                             <createSourcesJar>true</createSourcesJar>
                             <promoteTransitiveDependencies>false</promoteTransitiveDependencies>
                             <keepDependenciesWithProvidedScope>false</keepDependenciesWithProvidedScope>
-                            <createDependencyReducedPom>false</createDependencyReducedPom>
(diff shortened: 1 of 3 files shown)


all/pom.xml
```diff
@@ -428,7 +428,7 @@
                             <createSourcesJar>true</createSourcesJar>
                             <promoteTransitiveDependencies>false</promoteTransitiveDependencies>
                             <keepDependenciesWithProvidedScope>false</keepDependenciesWithProvidedScope>
-                            <createDependencyReducedPom>false</createDependencyReducedPom>
+                            <createDependencyReducedPom>true</createDependencyReducedPom>
```
lovepoem: The switch is open ,  then can create createDependencyReducedPom .xml files，so please add this file in .gitignore

Issue #1116, opened 2019-05-28
mysql field type is datetime afterRecords not equals currentRecords


jsbxyyx:
![image](https://user-images.githubusercontent.com/26616196/58444903-cf6bbf80-812c-11e9-960b-11a2f4470af1.png)

![image](https://user-images.githubusercontent.com/26616196/58444909-d5fa3700-812c-11e9-8263-d79d99c722d3.png)

suggestion: 
You can convert the field value to the field type, then compare

- JDK version : 1.8
- OS : window 7
- Others: mysql5.7

leizhiyuan:
Fix in https://github.com/seata/seata/pull/1090

PR #1117, opened 2019-05-28
fix field type is datetime equals fail

jsbxyyx:
fix: #1116


slievrly:
@jsbxyyx modify your pr title，will be as commit log.

ujjboy:
I think this is not the problem, the reason for #1116 is the `json` serialization loses millisecond precision.

jsbxyyx:
> I think this is not a real solution, the reason for #1116 is the `json` serialization loses millisecond precision.

I also think this is not a good solution. the problem solution requires a serialization method that preserves the type.

zhangthen:
The actual reason is that the serialization protocol(fastjson) used by undo log is not a good one.

PR #1120, opened 2019-05-28
bugfix : use xid wrong when do branch commit and rollback

### LINKED ISSUE #1118: Branch transcation cannot be committed when there are multiple transaction coordinator instance

### Ⅰ. Issue Description

When deploying multiple seata server instance, branch transaction may not be committed after starting committing global transaction.

### Ⅱ. Describe what happened

  A global transaction's XID is generated by method `XID.generateXID()`, which composes `ipAddress`, `port` and `uuid`.So XID is machine-related.
```
public class XID {

    private static int port;

    private static String ipAddress;

    /**
     * Generate xid string.
     *
     * @param tranId the tran id
     * @return the string
     */
    public static String generateXID(long tranId) {
        return ipAddress + ":" + port + ":" + tranId;
    }

    //...
}
```

  Global transation is committd asynchronously, so this phase may be processed on another seata server instance.When another seata server try to commit global transaction, in `DefaultCore.doGlobalCommit()`:
```
 BranchStatus branchStatus = resourceManagerInbound.branchCommit(
					branchSession.getBranchType(),
                    XID.generateXID(branchSession.getTransactionId()), 
					branchSession.getBranchId(),
                    branchSession.getResourceId(), 
					branchSession.getApplicationData());
```
XID is used to look for global transaction from the repository, while this XID is generated by this seata server's address(whick is different from it's orginal one).When setting db table as repository, None record can be found and branch committing is a default success, in `DefaultCoordinator.branchCommit()`:
```
GlobalSession globalSession = SessionHolder.findGlobalSession(xid);
if(globalSession == null){
	return BranchStatus.PhaseTwo_Committed;
}
```
After all branch transation is processed as default success,  global transaction is ended which seata server will delete it from db repository, in `DefaultCore.doGlobalCommit()` :
```
SessionHelper.endCommitted(globalSession);
```

As a result, so some RM cannot receive branchTransactionCommitRequest.

### Ⅵ. Environment:

Seata Version: 0.6.0

zhangthen:
fix #1118 

Use the xid wrong when do branch commit and rollback.

### DIFF
--- a/server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
+++ b/server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
@@ -298,7 +298,7 @@ public class DefaultCoordinator extends AbstractTCInboundHandler
                      * Prevent thread safety issues
                      */
                     SessionHolder.getRetryCommittingSessionManager().removeGlobalSession(rollbackingSession);
-                    LOGGER.error("GlobalSession rollback retry timeout [{}]", rollbackingSession.getTransactionId());
+                    LOGGER.error("GlobalSession rollback retry timeout [{}]", rollbackingSession.getXid());
                     continue;
                 }
                 rollbackingSession.addSessionLifecycleListener(SessionHolder.getRootSessionManager());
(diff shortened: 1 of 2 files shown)


zhangthen:
> Standalone is normal, this problem occurs with stateless clusters？XID.generateXID dependent on localIp

It occur when transaction start on one server, and commit on another server.

Issue #1122, opened 2019-05-28
Add a field for saving the context of undo log. 


ujjboy:
## Why you need it?

For supporting multiple serializations of `UndoLogParser`, we need to add a `rollback_ctx` field for saving the context of the undo log content.(such as serializer).

## How it could be?

Add `rollback_ctx` field to table `undo_log`, but may be incompatible with the old version.

PR #1123, opened 2019-05-28
Add a rollback_ctx field for saving the context of undo log. 

ujjboy:
Add a `rollback_ctx` field for saving the context of undo log. such as serializer/compress.

> **WARN：Incompatible with the old version** 

- Add `rollback_ctx` field to table `undo_log`.
- Save context when encode undo log.
- Parse context and get serializer before decode undo log.
- Add `getName` and `getDefaultContent()` of UndoLogParser.

#1122


zhangthen:
I think it's necessary to consider scalability.

If the default protostaff is found to be defective after 1.0, it will be a troublesome to replace the serialization protocol. The compatibility of existing data should be considered in upgrading.

ujjboy:
@slievrly  Do you mean all the undo log must use one serialization type?  

I thought about it and I think if we save context to each line can handle the following scenes:
-  Our user changes their serialization type from `json` to `protostuff`, We need to be compatible with historical data
- When we add the compress feature, and we can add key/value into this context, need to be compatible with historical data, too.

zhangthen:
> Is this serialization type better placed in a global file configuration? Instead of putting it on each line of undolog data records.

slievrly:
@ujjboy There is indeed a scene of switching serialization that you said, but the switch requires the user to restart the service, so whether you can ask the user to ensure that the undolog has no data, it seems a waste of each field. In fact, I don't think it is necessary to make a switch configuration because the undolog mechanism should be transparent to the user.

ujjboy:
@slievrly 

> so whether you can ask the user to ensure that the undo log has no data

But how to make sure they have no undo log data?  stop all the system? 

>  it seems a waste of each field

Yes, maybe tens of bytes, but will discharge by better serialization.

> In fact, I don't think it is necessary to make a switch configuration because the undolog mechanism should be transparent to the user.

Users can choose a serialization by considering the performance and resource usage, instead of specifying one by the framework. We need to give the user more choices, also user no need to know the detail of undo log mechanism.

leizhiyuan:
I think it is ok, because of the undo log will be deleted later, so store more data is acceptable.

common/src/main/java/io/seata/common/util/CollectionUtils.java
```diff
@@ -107,4 +109,59 @@ public static boolean isSizeEquals(Collection col0, Collection col1) {
             }
         }
     }
+
+    private static final String KV_SPLIT = "=";
+    
+    private static final String VALUE_SPLIT = "&";
```
slievrly: PAIR_SPLIT ?

slievrly:
> @slievrly
> 
> > so whether you can ask the user to ensure that the undo log has no data
> 
> But how to make sure they have no undo log data? stop all the system?
> 
> > it seems a waste of each field
> 
> Yes, maybe tens of bytes, but will discharge by better serialization.
> 
> > In fact, I don't think it is necessary to make a switch configuration because the undolog mechanism should be transparent to the user.
> 
> Users can choose a serialization by considering the performance and resource usage, instead of specifying one by the framework. We need to give the user more choices, also user no need to know the detail of undo log mechanism.

undolog context is necessary. Currently only useful for the rollback_info field, so whether to put it directly into the rollback_info field, do a general parsing first, parse out the context, and then serialize the rollback_info according to the context.

common/src/main/java/io/seata/common/util/CollectionUtils.java
ujjboy: OK, I will change the code.

Issue #1124, opened 2019-05-28
Add more serializers implements of UndoLogParser.


ujjboy:
## Why you need it?

JSON has good readability, good compatibility over languages, but poor performance and using more capacity.
We can add more serializers implements of UndoLogParser, provide more options for users to choose

 such as `protostuff`, `kyro`, `hessian`...

PR #1125, opened 2019-05-28
Add protostuff as serializer of UndoLogParser.

ujjboy:

- Add ProtostuffUndoLogParser
- Add test case

Part of #1124 

Fastjson run 100k times:
```
20:47:12.971 [main] INFO io.seata.rm.datasource.undo.BaseUndoLogParserTest - elapsed time 2212 ms.
20:47:12.991 [main] INFO io.seata.rm.datasource.undo.BaseUndoLogParserTest - data size:1387
```

protostuff run 100k times
```
20:48:46.432 [main] INFO io.seata.rm.datasource.undo.BaseUndoLogParserTest - elapsed time 951 ms.
20:48:46.447 [main] INFO io.seata.rm.datasource.undo.BaseUndoLogParserTest - data size:395

```


leizhiyuan:
please add some toString method to BranchUndoLog，SQLUndoLog ，so we can print some log for debugging

ujjboy:
@leizhiyuan  Do you mean to add into this PR? or another PR?

leizhiyuan:
> @leizhiyuan Do you mean to add into this PR? or another PR?

ok，maybe we can do it in another pr

rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/ProtostuffUndoLogParser.java
```diff
@@ -0,0 +1,209 @@
...
+            return ProtostuffIOUtil.toByteArray(branchUndoLog, schema, buffer);
+        } finally {
+            buffer.clear();
+        }
+    }
+
+    @Override
+    public BranchUndoLog decode(byte[] bytes) {
+        if (bytes.length == 0) {
+            return new BranchUndoLog();
```
leizhiyuan: return null？

ujjboy: 1. `FastjsonUndoLogParser` default content is `{}`, equivalent to `return new BranchUndoLog()`
2. If return null, we need to add more NPE check in `UndoLogManager`, such as :

```java
BranchUndoLog branchUndoLog = UndoLogParserFactory.getInstance().decode(rollbackInfo);
for (SQLUndoLog sqlUndoLog : branchUndoLog.getSqlUndoLogs()) {  <--- here

```

bom/pom.xml
```diff
@@ -141,6 +142,16 @@
                 <artifactId>sofa-rpc-all</artifactId>
                 <version>${sofa.rpc.version}</version>
             </dependency>
+            <dependency>
```
slievrly: just dependencyManagement in bom.pom and not contains in all.pom, must be dependency separately when the user uses it.

ujjboy: The default serialization is `fastjson` now, so I think if the users want to switch to `protostuff`, they need to add additional dependencies by themselves. But if the default is `protostuff`,  we can add dependencies of `protostuff` by default.    

About this,  @slievrly  @leizhiyuan  Do you have any good opinions?

leizhiyuan: I think we can add by default.

slievrly: I think the solution is the same as the existing multiple registration centers and configuration centers.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/ProtostuffUndoLogParser.java
```diff
@@ -0,0 +1,209 @@
...
+    private final static DefaultIdStrategy ID_STRATEGY = ((DefaultIdStrategy) RuntimeEnv.ID_STRATEGY);
+
+    static {
+        ID_STRATEGY.registerDelegate(new DateDelegate());
+        ID_STRATEGY.registerDelegate(new TimestampDelegate());
+        ID_STRATEGY.registerDelegate(new SqlDateDelegate());
+        ID_STRATEGY.registerDelegate(new TimeDelegate());
+    }
+
+    private Schema<BranchUndoLog> schema = RuntimeSchema.getSchema(BranchUndoLog.class);
```
slievrly: schema cache(map or static) or singleton spi instance.

rm-datasource/src/test/java/io/seata/rm/datasource/undo/parser/ProtostuffUndoLogParserTest.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package io.seata.rm.datasource.undo.parser;
+
+import io.seata.rm.datasource.undo.BaseUndoLogParserTest;
+import io.seata.rm.datasource.undo.UndoLogParser;
+
+/**
+ * @author Geng Zhang
+ */
+class ProtostuffUndoLogParserTest extends BaseUndoLogParserTest {
```
slievrly: where used？add some case in UndoLogParserProviderTest.java?

ujjboy: OK, there are two test method: `testEncodeAndDecode()` and `testPerformance()` in `BaseUndoLogParserTest`, and I will add test case of load in `UndoLogParserProviderTest`.

PR #1145, opened 2019-05-31
Change LockMode from MEMORY to DB when the StoreMode is DB

zhangthen:

If the StoreMode is  'DB' and the LockMode is 'MEMORY', the locks in 'MEMORY' cann't be released normally  after the transaction commit or rollback.

Change LockMode from MEMORY to DB automatically when the StoreMode is DB .


server/src/main/java/io/seata/server/lock/LockerFactory.java
```diff
@@ -72,7 +76,15 @@ public static synchronized final LockManager getLockManager() {
      * @return the lock manager
      */
     public static synchronized final Locker get(BranchSession branchSession) {
+        String storeMode = CONFIG.getConfig(ConfigurationKeys.STORE_MODE);
         String lockMode = CONFIG.getConfig(ConfigurationKeys.LOCK_MODE);
+        //storeMode == db and lockMode == memeory is not allowed
```
leizhiyuan: can we remove the settings of lock mode, only one setting: "storeMode"?

Issue #976, opened 2019-05-06
buffer size is hard-code as 1K, which cannot support bulk update cases


bbcpan:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
buffer size is hard-code as 1K, which cannot support bulk update cases

### Ⅱ. Describe what happened
when use update statements for bulk records in on local transaction 

  If there is an exception, please attach the exception trace:
![image](https://user-images.githubusercontent.com/22337047/57206421-2849ab80-6ff8-11e9-921f-f75348a8bc33.png)

Just paste your stack trace here!
```
![image](https://user-images.githubusercontent.com/22337047/57206428-313a7d00-6ff8-11e9-80f3-dafeba691035.png)



### Ⅲ. Describe what you expected to happen


### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅴ. Anything else we need to know?


### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

xingfudeshi:
Could you.submit a PR to improve this?thanks.

PR #1148, opened 2019-06-02
bugfix:the buffer may cause overflows when sql statement is long

kezhenxu94:
Due to the `java.nio.ByteBuffer` can not grow dynamically, its size is hardcoded now, this patch replaces `java.nio.ByteBuffer` with `io.netty.buffer.ByteBuf`, which can grow on demand.

fixes issue #976 

There is already tests upon this method.

CI passes

null

### DIFF
--- a/core/src/main/java/io/seata/core/protocol/MergedWarpMessage.java
+++ b/core/src/main/java/io/seata/core/protocol/MergedWarpMessage.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
  */
 public class MergedWarpMessage extends AbstractMessage implements Serializable, MergeMessage {
     private static final long serialVersionUID = -5758802337446717090L;
+
     /**
      * The Msgs.
      */
@@ -51,7 +52,7 @@ public class MergedWarpMessage extends AbstractMessage implements Serializable, 
 
     @Override
     public byte[] encode() {
-        final ByteBuf buffer = Unpooled.buffer();
+        final ByteBuf buffer = Unpooled.buffer(1024);
         buffer.writeInt(0); // write placeholder for content length
 
         buffer.writeShort((short) msgs.size());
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/seata/core/protocol/MergedWarpMessage.java
```diff
@@ -49,21 +51,21 @@ public short getTypeCode() {
...
     public byte[] encode() {
-        int bufferSize = msgs.size() * 1024;
-        ByteBuffer byteBuffer = ByteBuffer.allocate(bufferSize);
-        byteBuffer.putShort((short)msgs.size());
-        for (AbstractMessage msg : msgs) {
-            //msg.setChannelHandlerContext(ctx);
-            byte[] data = msg.encode();
-            byteBuffer.putShort(msg.getTypeCode());
-            byteBuffer.put(data);
+        final ByteBuf buffer = Unpooled.buffer();
```
leizhiyuan: why do you use unpooled? maybe pooled is safer?

kezhenxu94: make sense, but it's what Netty recommends to create a `ByteBuf`:

https://github.com/netty/netty/blob/128403b492d8490d9fc770b7a2dad669433b2790/buffer/src/main/java/io/netty/buffer/ByteBuf.java#L37-L41

and there is no something like `Pooled`, or should I call `new PooledByteBufAllocator(false).buffer()` directly? which is not recommended.

What do you think? @leizhiyuan

slievrly: default buffer size=256，can we set a larger size and reduce the number of resize?

kezhenxu94: @slievrly maybe set to what we hardcoded before? (say 1024)

Issue #1139, opened 2019-05-31
DataCompareUtils#isFieldEquals equals type


jsbxyyx:
the isFieldEquals method only compare object (the object must override equals and hashcode)

can compare type:
String, BigDecimal, boolean, byte, short, int, long, float, double, Date, Time, Timestamp, URL

for other type can close validation

JDBC Type | Java Type
-- | --
CHAR | String
VARCHAR | String
LONGVARCHAR | String
NUMERIC | java.math.BigDecimal
DECIMAL | java.math.BigDecimal
BIT | boolean
BOOLEAN | boolean
TINYINT | byte
SMALLINT | short
INTEGER | int
BIGINT | long
REAL | float
FLOAT | double
DOUBLE | double
BINARY | byte[]
VARBINARY | byte[]
LONGVARBINARY | byte[]
DATE | java.sql.Date
TIME | java.sql.Time
TIMESTAMP | java.sql.Timestamp
CLOB | Clob
BLOB | Blob
ARRAY | Array
DISTINCT | mapping of underlying type
STRUCT | Struct
REF | Ref
DATALINK | java.net.URL
JAVA_OBJECT | underlying Java class

xingfudeshi:
Could you submit a PR to improve this?thanks.

jsbxyyx:
other ways to achieve

dk-lockdown:
> other ways to achieve

how this problem solved?i reviewed the code,find just BigDecimal processing has been done,orther types such as BigInteger are still causing bugs

jsbxyyx:
@dk-lockdown you see lastest version

PR #1149, opened 2019-06-03
fix DataCompareUtils#isFieldEquals equals type

jsbxyyx:
fix: #1139


rm-datasource/src/main/java/io/seata/rm/datasource/DataCompareUtils.java
```diff
@@ -65,6 +65,9 @@ public static boolean isFieldEquals(Field f0, Field f1) {
                             if (f1Type == Types.TIMESTAMP && f1.getValue().getClass().equals(String.class)) {
                                 f1.setValue(Timestamp.valueOf(f1.getValue().toString()));
                             }
+                            if (!checkType(f0.getValue())) {
+                                return true;
```
xingfudeshi: return false?

jsbxyyx: ignore other type compare

xingfudeshi: > @xingfudeshi ignore other type compare, because other type not override equals and hashcode, when comparing f0.value and f1.value, the actual comparison is the object address. the compare result certain false.

If true is returned here, does f0. value equals to f1. value?

xingfudeshi: if f0.value is not any instance of there types,then f0.value equals f1.value?

xingfudeshi: what i mean is if the other type dose not override equals and hashcode method,so we don't need to compare them,right?

jsbxyyx: you are right, I understand your meanning, I will update it later.

rm-datasource/src/test/java/io/seata/rm/datasource/undo/BaseH2Test.java
```diff
@@ -15,23 +15,14 @@
...
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.mockito.Mockito;
 
-import java.sql.Connection;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.sql.Types;
+import java.sql.*;
```
xingfudeshi: Please import explicit class type.do not use .*

rm-datasource/src/test/java/io/seata/rm/datasource/undo/BaseH2Test.java
```diff
@@ -15,23 +15,14 @@
  */
 package io.seata.rm.datasource.undo;
 
-import io.seata.rm.datasource.sql.struct.ColumnMeta;
-import io.seata.rm.datasource.sql.struct.Field;
-import io.seata.rm.datasource.sql.struct.KeyType;
-import io.seata.rm.datasource.sql.struct.Row;
-import io.seata.rm.datasource.sql.struct.TableMeta;
-import io.seata.rm.datasource.sql.struct.TableRecords;
+import io.seata.rm.datasource.sql.struct.*;
```
xingfudeshi: Please import explicit class type.do not use .*

jsbxyyx: ok

rm-datasource/src/test/java/io/seata/rm/datasource/undo/BaseH2Test.java
jsbxyyx: ok

jsbxyyx:
other ways to achieve

PR #1169, opened 2019-06-04
bugfix: use Set to avoid duplicate listeners. fixes #1126

### LINKED ISSUE #1126: There are too many identical SessionLifecycleListener`s in GlobalSession
在GlobalSession中的lifecycleListeners字段中有很多相同实例的监听器，导致每次更新或删除时都会操作很多遍，考虑是否可以用set代替list

kezhenxu94:
This patch replaces `ArrayList` with `HaseSet` to avoid duplicated listeners as described in #1126 , and also fix some typos

fixes #1126 

Already have

CI passes

null

### DIFF
--- a/server/src/main/java/io/seata/server/session/GlobalSession.java
+++ b/server/src/main/java/io/seata/server/session/GlobalSession.java
@@ -18,6 +18,8 @@ package io.seata.server.session;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -68,7 +70,7 @@ public class GlobalSession implements SessionLifecycle, SessionStorable {
 
     private boolean active = true;
 
-    private ArrayList<BranchSession> branchSessions = new ArrayList<>();
+    private final ArrayList<BranchSession> branchSessions = new ArrayList<>();
 
     private GlobalSessionLock globalSessionLock = new GlobalSessionLock();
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/seata/server/session/GlobalSession.java
```diff
@@ -249,8 +251,7 @@ public BranchSession getBranch(long branchId) {
      * @return the sorted branches
      */
     public ArrayList<BranchSession> getSortedBranches() {
-        ArrayList<BranchSession> sorted = new ArrayList();
-        sorted.addAll(branchSessions);
+        ArrayList<BranchSession> sorted = new ArrayList<>(branchSessions);
```
leizhiyuan: please sort the HashSet

> It makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time.

kezhenxu94: @leizhiyuan sorry, but I just replace the `ArrayList` of `lifecycleListeners`, and the `branchSessions` is still `ArrayList`:

image

I just add a `final` modifier to prevent it being modified in the future because it's used as a lock in `synchronized` block:

image

kezhenxu94: And I'm sure that `new ArrayList<>(branchSessions);` will return a new list that keep the original order

slievrly:
> See the comment

Is the hashset's order of addition and traversal order consistent? There is currently no order requirement in the business.

leizhiyuan:
> > See the comment
> 
> Is the hashset's order of addition and traversal order consistent? There is currently no order requirement in the business.

yes，what I mean is the order of hashset, but the method name is 'getSortedBranches' maybe I think that sort it well..

slievrly:
> > > See the comment
> > 
> > 
> > Is the hashset's order of addition and traversal order consistent? There is currently no order requirement in the business.
> 
> yes，what I mean is the order of hashset, but the method name is 'getSortedBranches' maybe I think that sort it well..

yeah, `getSortedBranches` need sort, but it's a ArrayList used Array copy method.

Issue #1192, opened 2019-06-12
Get the LockMode by StoreMode


zhangthen:
### Ⅰ. Issue Description

Lockmode and storemode are related, we will delete the lockmode, which would be judged by the storemode.

PR #1193, opened 2019-06-12
Get lockmode by the storemode

zhangthen:
#1192


server/src/main/resources/file.conf
```diff
@@ -82,11 +82,7 @@ store {
 }
 lock {
     ## the data row lock store mode: local_db、memory or db
-    mode = "memory"
-
-    memory{
-         ## store lock in memory of server
-    }
+    mode = ""
```
leizhiyuan: should we remove lock.mode from conf

zhangthen: > should we remove lock.mode from conf

I'll use “lock.mode”  in some other cases.

server/src/main/resources/file.conf
```diff
@@ -82,11 +82,7 @@ store {
 }
 lock {
```
slievrly: I think  should delete the lock node, which is strongly associated with the store. As for the table name, we should use our default table name just like `undolog`, and do not need to allow user settings.

zhangthen: > I think should delete the lock node, which is strongly associated with the store. As for the table name, we should use our default table name just like `undolog`, and do not need to allow user settings.

I will store the row locks in the user's database, beside the undo_log table, and need a lock node to configure this switcher.

PR #1214, opened 2019-06-17
add config `shutdown.wait` and update version to 0.7.0-SNAPSHOT

### LINKED ISSUE #1212: when destroy server, lack config `transport.shutdown.wait`

### Ⅰ. Issue Description

when destroy server, lack config `transport.shutdown.wait`

### Ⅱ. Describe what happened

  
image

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:

add config `shutdown.wait` and update version to 0.7.1-SNAPSHOT

fix #1212

### DIFF
--- a/core/src/main/java/io/seata/core/protocol/Version.java
+++ b/core/src/main/java/io/seata/core/protocol/Version.java
@@ -31,7 +31,7 @@ public class Version {
     /**
      * The constant CURRENT.
      */
-    public static final String CURRENT = "0.6.1";
+    public static final String CURRENT = "0.7.0";
 
     /**
      * The constant VERSION_MAP.
--- a/all/pom.xml
+++ b/all/pom.xml
@@ -21,7 +21,7 @@
 
     <groupId>io.seata</groupId>
     <artifactId>seata-all</artifactId>
-    <version>0.6.1</version>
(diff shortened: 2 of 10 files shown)


xingfudeshi:
In the tiltle,the version may be 0.7.0-SNAPSHOT?

core/src/main/java/io/seata/core/protocol/Version.java
```diff
@@ -31,7 +31,7 @@
     /**
      * The constant CURRENT.
      */
-    public static final String CURRENT = "0.6.1";
+    public static final String CURRENT = "0.7.0";
```
lovepoem: How about use the version configured in pom.xml  dynamicly?

lovepoem: or the version in meta-info in the jar ?

slievrly: But when we run with IDE, there is no meta-info here.

lovepoem: > But when we run with IDE, there is no meta-info here.

I think it is a question , we can retain this issue temporarily and merge this pr

Issue #1176, opened 2019-06-06
Field case problem in SQL statement


yangom:
数据库表不是自增主键，自己写的SQL,insert 语句里主键名称用了大写，在使用SEATA代理数据源的时候，他从数据库表取出的元信息里主键是小写，然后SEATA就判定这是个自增主键表，就会提示数据库需要自增字段

xingfudeshi:
Could you provide more details?like exception traces or some code to reproduce this problem?Thanks.

kezhenxu94:
@yangom I don't think this is a bug if you're using Windows (I'm not sure), in Mac OS (and possibly Linux), the column name is case insensitive, I've tested like this and it works well:

- based on the [seata-samples](http://github.com/seata/seata-samples), make the `storage_tbl` not to auto increment:

```sql
alter table storage_tbl modify column id int(11) not null;
```

- modify `io.seata.samples.dubbo.starter.DubboStorageServiceStarter` and start it:

```java
public class DubboStorageServiceStarter {
    /**
     * 1. Storage service is ready . A seller add 100 storage to a sku: C00321
     *
     * @param args the input arguments
     */
    public static void main(String[] args) {
        ClassPathXmlApplicationContext storageContext = new ClassPathXmlApplicationContext(
            new String[]{"spring/dubbo-storage-service.xml"});
        storageContext.getBean("service");
        JdbcTemplate storageJdbcTemplate = (JdbcTemplate) storageContext.getBean("jdbcTemplate");
        storageJdbcTemplate.update("delete from storage_tbl where commodity_code = 'C00321'");
        storageJdbcTemplate.update("insert into storage_tbl(ID, commODity_code, count) values (200, 'C00321', 100)");
        new ApplicationKeeper(storageContext).keep();
    }
}
```

pay special attention to line `storageJdbcTemplate.update("insert into storage_tbl(ID, commODity_code, count) values (200, 'C00321', 100)");`, it use `ID` instead of `id`, and the sql executes normally.

@yangom if you're not sure whether it's a bug of seata, try to replace the datasource with the original (that doesn't proxied by Seata), and see whether the problem exists or not.

cc @xingfudeshi

xingfudeshi:
> @yangom I don't think this is a bug if you're using Windows (I'm not sure), in Mac OS (and possibly Linux), the column name is case insensitive, I've tested like this and it works well:
> 
[...]

PR #1218, opened 2019-06-18
bugfix:containsPK ignoreCase

### LINKED ISSUE #1217: Rm cannot find primary key when primary key is Uppercase in sql ,using MySQL.

### Ⅰ. Issue Description
when I write a Insert SQL which the primary key is uppercase, RM cannot generate afterimge.
I find InsertExecutor.containsPK() judge if pk exists by comparing MySQL table metadata with the result of SQL recognizer. Which compare the lowercase PK and uppercase Pk.
When this table dosenot have auto-increase PK, RM throw exception.
### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others: Seata 0.6.1

slievrly:

bugfix:containsPK ignoreCase

fix #1217 #1176

### DIFF
--- a/common/src/main/java/io/seata/common/util/CollectionUtils.java
+++ b/common/src/main/java/io/seata/common/util/CollectionUtils.java
@@ -177,7 +177,11 @@ public class CollectionUtils {
         if (null == sourceList || sourceList.size() == 0) { return sourceList; }
         List<String> destList = new ArrayList<>(sourceList.size());
         for (String element : sourceList) {
-            destList.add(element.toUpperCase());
+            if (null != element) {
+                destList.add(element.toUpperCase());
+            } else {
+                destList.add(element);
+            }
         }
         return destList;
     }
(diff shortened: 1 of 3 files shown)


VainEffort:
line 179.
Element needs to check null.

common/src/main/java/io/seata/common/util/CollectionUtils.java
```diff
@@ -164,4 +166,19 @@ public static String encodeMap(Map<String, String> map) {
...
+     * To upper list list.
+     *
+     * @param sourceList the source list
+     * @return the list
+     */
+    public static List<String> toUpperList(List<String> sourceList) {
+        if (null == sourceList || sourceList.size() == 0) { return sourceList; }
+        List<String> destList = new ArrayList<>(sourceList.size());
+        for (String element : sourceList) {
+            destList.add(element.toUpperCase());
```
VainEffort: Element needs to check null.

slievrly:
> line 179.
> Element needs to check null.

great found.i will fix it.

common/src/main/java/io/seata/common/util/CollectionUtils.java
slievrly: fixed

PR #1220, opened 2019-06-19
Fix Rm cannot find primary key when primary key is Uppercase in sql ,using MySQL.

### LINKED ISSUE #1217: Rm cannot find primary key when primary key is Uppercase in sql ,using MySQL.

xionganyin:
Rm cannot find primary key when primary key is Uppercase in sql ,using MySQL.

fixes #1217


slievrly:
@xionganyin associate with your github email. `git config user.email 'xxxx'`

PR #1222, opened 2019-06-19
fix bug of refresh cluster

leizhiyuan:

fixes https://github.com/seata/seata/issues/1160


discovery/seata-discovery-eureka/src/main/java/io/seata/discovery/registry/eureka/EurekaRegistryServiceImpl.java
```diff
@@ -159,18 +160,17 @@ public void close() throws Exception {
...
         }
         for (Application app : list) {
             Set<InetSocketAddress> addressSet = new HashSet<>();
             List<InstanceInfo> instances = app.getInstances();
-            if (instances == null || instances.isEmpty()) {
-                break;
-            }
-            for (InstanceInfo instance : instances) {
-                addressSet.add(new InetSocketAddress(instance.getIPAddr(), instance.getPort()));
+            if (CollectionUtils.isNotEmpty(instances)) {
```
slievrly: I think if it is not the application we are concerned about, we should ignore it, so we add an application judgment in front of it, this parameter is passed in through the file.

PR #1270, opened 2019-07-05
bugfix: print error exception

xionganyin:
fix catch exception with no output or throw


core/src/main/java/io/seata/core/exception/AbstractExceptionHandler.java
```diff
@@ -113,8 +117,10 @@ public void exceptionHandleTemplate(Callback callback, AbstractTransactionReques
             callback.execute(request, response);
             callback.onSuccess(request, response);
         } catch (TransactionException tex) {
+            LOGGER.error("执行Rpc异常", tex);
```
slievrly: print request detail

PR #1278, opened 2019-07-08
bugfix: pass txId into TCC interceptor

zhangthen:
### 1、Problem

The transaction context was cleared in PR #1195 , which cause the txId be null when regist branch record.

### 2. Solution

Save the txId when clear transaction context, and pass it into TCC interceptor.


ujjboy:
LGTM

PR #1286, opened 2019-07-11
bugfix: add some configuration and exclude log dependency

slievrly:

bugfix: add some configuration and exclude log dependency


server/src/main/resources/nacos-config.txt
```diff
@@ -43,4 +43,10 @@ recovery.rollbacking-retry-delay=30
 recovery.timeout-retry-delay=30
 transaction.undo.data.validation=true
 transaction.undo.log.serialization=jackson
```
xingfudeshi: "jackson"

slievrly: This is a script that initializes nacos and cannot be quoted. All configuration values are strings. In the configuration center is displayed jackson instead of "jackson".

Issue #1320, opened 2019-07-17
MySql8.0 version service connection is abnormal


jiangbin31:
现象：
在使用mysql8.0版本数据库情况，数据库连接不上。
原因：
新版mysql8.0的jar包 和 驱动不对。
新版jar
    <dependency>
            <groupId>mysql</groupId>
            <artifactId>mysql-connector-java</artifactId>
            <version>8.0.12</version>
        </dependency>

驱动driverClassName：com.mysql.cj.jdbc.Driver

建议：
驱动可以在file.conf 进行配置

jiangbin31:
配置
![image](https://user-images.githubusercontent.com/22318753/61352021-e0b88900-a89e-11e9-84f6-4e0b1566226f.png)

![image](https://user-images.githubusercontent.com/22318753/61352037-e9a95a80-a89e-11e9-970a-2f82306938d7.png)

![image](https://user-images.githubusercontent.com/22318753/61352066-fded5780-a89e-11e9-886a-0df3253ee268.png)


数据库连接异常

vikenlove:
@jiangbin31  please tell me  which is version of mysql ?

vikenlove:
@jiangbin31  i was commit a pr  about config to driverclassname

jiangbin31:
mysql  version : 8.0
The new mysql jar package maven address is  【
    <dependency>
            <groupId>mysql</groupId>
            <artifactId>mysql-connector-java</artifactId>
            <version>8.0.12</version>
        </dependency>
】

Driverclassname  is  【com.mysql.cj.jdbc.Driver】。

slievrly:
@vikenlove thank you.

ChaunceyBee:
step 1.  Change the MySQL JAR file in the lib path to 8.0+
step 2.  Change the 

> > Hi @jiangbin31, we detect non-English characters in the issue. This comment is an auto translation from @fescar-robot to help other users to understand this issue.
> > _**We encourage you to describe your issue in English which is more friendly to other users.**_
> 
> # MySql8.0 version service connection is abnormal
> phenomenon:
> In the case of using the mysql8.0 version of the database, the database connection is not available.
> the reason:
> The new version of mysql8.0 jar package and driver is wrong.
> New jar
> 
> mysql
> mysql-connector-java
> 8.0.12
> Driver driverClassName: com.mysql.cj.jdbc.Driver
> 
> Suggest:
> The driver can be configured in file.conf

[...]

PR #1323, opened 2019-07-17
add undo_log.sql   and add database driver class config

vikenlove:
add a sql file when the first time  to use seata . it is init to sql database  and add database driver class name  of config

### DIFF
--- a/core/src/main/java/io/seata/core/store/db/AbstractDataSourceGenerator.java
+++ b/core/src/main/java/io/seata/core/store/db/AbstractDataSourceGenerator.java
@@ -43,7 +43,17 @@ public abstract class AbstractDataSourceGenerator implements DataSourceGenerator
     protected DBType getDBType() {
         return DBType.valueof(CONFIG.getConfig(ConfigurationKeys.STORE_DB_TYPE));
     }
-
+    /**
+     * get db driver class name
+     * @return the db driver class name
+     */
+    protected String getDriverClassName() {
+		String driverClassName = CONFIG.getConfig(ConfigurationKeys.STORE_DB_DRIVER_CLASS_NAME);
+		if (StringUtils.isBlank(driverClassName)) {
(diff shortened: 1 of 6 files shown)


slievrly:
fix #1320

slievrly:
In fact, the server side does not need the undolog table.

core/src/main/java/io/seata/core/store/db/AbstractDataSourceGenerator.java
```diff
@@ -43,7 +43,17 @@
...
     }
-
+    /**
+     * get db driver class name
+     * @return the db driver class name
+     */
+    protected String getDriverClassName() {
+		String driverClassName = CONFIG.getConfig(ConfigurationKeys.STORE_DB_DRIVER_CLASS_NAME);
+		if (StringUtils.isBlank(driverClassName)) {
+	            throw new StoreException("the {store.db.driver-class-name} can't empty.");
```
xingfudeshi: can't be empty

xingfudeshi:
> In fact, the server side does not need the undolog table.

Yes.But it is a good idea to provide undo_log scripts on the server side. Because it will always be the same as the server version.

core/src/main/java/io/seata/core/store/db/AbstractDataSourceGenerator.java
vikenlove: the other exception is "can't empty." so i copy that :D

vikenlove: 我等会改下

xingfudeshi: Then,please correct all of it.

vikenlove: the exception message was fixed "can't be empty" and commited it

Issue #1318, opened 2019-07-17
RM client may cause memory leak.


xingfudeshi:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
In the logs of my application,i saw the following logs.it may cause memory leak.

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
2019-07-17 11:59:03.808 ERROR 19459 --- [lettuce-epollEventLoop-9-3] io.netty.util.ResourceLeakDetector       : LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:349)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:187)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:178)
	io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:139)
	io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator$MaxMessageHandle.allocate(DefaultMaxMessagesRecvByteBufAllocator.java:114)
	io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:147)
	io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:682)
	io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:617)
	io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:534)
	io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:496)
	io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:906)
	io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	java.lang.Thread.run(Thread.java:748)
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

[...]

PR #1333, opened 2019-07-22
bugfix: netty memory leak

### LINKED ISSUE #1304: Netty memory leak

### Ⅰ. Issue Description

image

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

cmonkey:
fix #1304 #1318

### DIFF
--- a/core/src/main/java/io/seata/core/rpc/netty/v1/ProtocolV1Decoder.java
+++ b/core/src/main/java/io/seata/core/rpc/netty/v1/ProtocolV1Decoder.java
@@ -23,6 +23,8 @@ import io.seata.core.codec.CodecFactory;
 import io.seata.core.protocol.HeartbeatMessage;
 import io.seata.core.protocol.ProtocolConstants;
 import io.seata.core.protocol.RpcMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/seata/core/rpc/netty/v1/ProtocolV1Decoder.java
```diff
@@ -74,8 +74,14 @@ public ProtocolV1Decoder(int maxFrameLength) {
     protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
         Object decoded = super.decode(ctx, in);
         if (decoded instanceof ByteBuf) {
-            ByteBuf frame = (ByteBuf) decoded;
-            return decodeFrame(frame);
+            try {
+                ByteBuf frame = (ByteBuf) decoded;
+                return decodeFrame(frame);
+            }catch(Exception e){
+                throw e;
```
lovepoem: please print some log

core/src/main/java/io/seata/core/rpc/netty/v1/ProtocolV1Decoder.java
```diff
@@ -74,8 +74,14 @@ public ProtocolV1Decoder(int maxFrameLength) {
...
+            }finally {
+                in.release();
```
ujjboy: Not release `in` but `frame`

slievrly: agree，`in` can be released in `ByteToMessageDecoder.channelRead`, but `frame` allocate in LengthFieldBasedFrameDecoder.

cmonkey: change frame.release and add error logger

PR #1337, opened 2019-07-23
batch insert support for non-self-inc primary keys

zjinlei:
feature: batch insert support for non-self-incrementing primary keys

use test case BatchInsertExecutorTest.java


rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
```diff
@@ -82,30 +83,50 @@ protected TableRecords afterImage(TableRecords beforeImage) throws SQLException
...
+                }
+            }
+            for (int i = 0; i < cycleNums; i++) {
+                pkIndexs.add(insertColumnsSize * i + firstPkIndex);
+            }
+            if (pkIndexs.size() == 1) {
+                //adapter test case
+                pkValues = preparedStatementProxy.getParamsByIndex(pkIndexs.get(0));
+            } else {
+                pkValues = pkIndexs.stream().map(pkIndex -> paramters[pkIndex].get(0)).collect(Collectors.toList());
```
slievrly: get(0)?

slievrly: I understand that this should be `firstPkIndex`

xingfudeshi:
Unit Tests are passed.

PR #1338, opened 2019-07-23
bugfix: Lock is not acquired when multiple branches have cross locks#1247

### LINKED ISSUE #1247: RM failed to acquire lock

### Ⅰ. Issue Description

RM fails to acquire global lock for it's operated record.

### Ⅱ. Describe what happened

There are two RM in a global transaction. The first RM inserts a record, the second RM updated this record and another reocord **in a local transaction**.
When I pause global transaction after phase1 completed and before start phase2 where global lock isnot released. I find there is only one global lock record in TC's lock_table for the inserted record.

I check the code for acquiring Lock in DB mode, class `LockStoreDataBaseDAO.class`, method `public boolean acquireLock(List<LockDO> lockDOs)`, line 113 to line 135:

```
            //query
            String checkLockSQL = LockStoreSqls.getCheckLockableSql(lockTable, sb.toString(), dbType);
            ps = conn.prepareStatement(checkLockSQL);
            for (int i = 0; i < lockDOs.size(); i++) {
                ps.setString(i + 1, lockDOs.get(i).getRowKey());
            }
            rs = ps.executeQuery();
            while (rs.next()) {
                if (StringUtils.equals(rs.getString("xid"), lockDOs.get(0).getXid())) {
                    isReLock = true;
                } else {
                    canLock &= false;
                }
            }

            if (!canLock) {
                conn.rollback();
                return false;
            }

            if (isReLock) {
                conn.rollback();
                return true;
            }

```
There is a logic problem when RM try to acquire lock including the ones are acquired and the other are not acquired at the same time. TC find there exist relock so it jump out without acquring the other lock.

### III. Environment:

- JDK version : jdk1.8
- OS : linux
- Others: Seata 0.6.1

zjinlei:
bugfix #1247 

fixes #1247

### DIFF
--- a/core/src/main/java/io/seata/core/store/db/LockStoreDataBaseDAO.java
+++ b/core/src/main/java/io/seata/core/store/db/LockStoreDataBaseDAO.java
@@ -18,6 +18,7 @@ package io.seata.core.store.db;
 import io.seata.common.exception.StoreException;
 import io.seata.common.executor.Initialize;
 import io.seata.common.loader.LoadLevel;
+import io.seata.common.util.CollectionUtils;
 import io.seata.common.util.StringUtils;
 import io.seata.config.Configuration;
 import io.seata.config.ConfigurationFactory;
@@ -31,7 +32,10 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * The type Data base lock store.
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/seata/core/store/db/LockStoreDataBaseDAO.java
```diff
@@ -114,25 +120,29 @@ public boolean acquireLock(List<LockDO> lockDOs) {
             }
             rs = ps.executeQuery();
             while (rs.next()) {
-                if (StringUtils.equals(rs.getString("xid"), lockDOs.get(0).getXid())) {
-                    isReLock = true;
-                } else {
+                if (!StringUtils.equals(rs.getString("xid"), lockDOs.get(0).getXid())) {
                     canLock &= false;
+                    break;
```
slievrly: why break?
db: lock1:xid1   lock2:xid2  lock3:xid1
request: lock1:xid1  lock2:xid1  lock3:xid1

lock3 can add in dbExistedRowKeys?

zjinlei: db: lock2:xid2
request:  lock2:xid1
These two global transactions are different. You cannot get the lock this time.
add break is a quick failure

slievrly: agree

PR #1352, opened 2019-07-26
Update UpdateExecutor.java

Robin135:

When generating before/after images, the UpdateExecutor only records the changed columns, while the AbstractUndoExecutor compares all columns before handling  a rollback, which may cause SQLException("Has dirty records when undo.")  In fact, comparing all columns is necessary. 

I found this problem, but have not created an issue yet.

Just update  some of the columns , then execute a  rollback.

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
@@ -21,7 +21,6 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.StringJoiner;
 
 import io.seata.rm.datasource.ParametersHolder;
 import io.seata.rm.datasource.StatementProxy;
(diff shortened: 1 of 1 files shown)


slievrly:
@Robin135 I recommend doing a switch that only verifies the modified field or all the fields, and the two sides are consistent.

zjinlei:
not a bug,only check changed columns.

beston123:
> ### Ⅰ. Describe what this PR did
> When generating before/after images, the UpdateExecutor only records the changed columns, while the AbstractUndoExecutor compares all columns before handling a rollback, which may cause SQLException("Has dirty records when undo.") In fact, comparing all columns is necessary.
> 
> ### Ⅱ. Does this pull request fix one issue?
> I found this problem, but have not created an issue yet.
> 
> ### Ⅲ. Why don't you add test cases (unit test/integration test)?
> ### Ⅳ. Describe how to verify it
> Just update some of the columns , then execute a rollback.
> 
> ### Ⅴ. Special notes for reviews



> the AbstractUndoExecutor compares all columns

It is not necessary to fix , because the AbstractUndoExecutor only compares the changed columns. See the method 'DataCompareUtils.compareRows', the comparison are based on the columns of beforeImage/afterImage.
```
        for (String fieldName : oldRow.keySet()) {
                Field oldField = oldRow.get(fieldName);
                Field newField = newRow.get(fieldName);
                if (newField == null) {
                    return false;
                }
                if (!isFieldEquals(oldField, newField)) {
                    return false;
                }
            }
```

zjinlei:
sorry,I closed it by mistake.

PR #1355, opened 2019-07-27
test: Add test case for rpc protocol.

ujjboy:
- Add test case of new rpc protocol. 
- Change field type of RpcMessage#headMap to `HashMap`, because of `HashMap` can save the `null` value and RpcMessage only use in one single thread.

### DIFF
--- a/test/src/test/java/io/seata/core/rpc/netty/v1/ProtocolV1CodecTest.java
+++ b/test/src/test/java/io/seata/core/rpc/netty/v1/ProtocolV1CodecTest.java
@@ -1,0 +1,105 @@
+/*
+ *  Copyright 1999-2019 Seata.io Group.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


test/src/test/java/io/seata/core/rpc/netty/v1/ProtocolV1CodecTest.java
```diff
@@ -0,0 +1,105 @@
...
+                        }
+                    }
+                });
+            }
+
+            cnt.await(10, TimeUnit.SECONDS);
+            LOGGER.info("success {}/{}", success.get(), runTimes);
+            Assertions.assertEquals(success.get(), runTimes);
+        } catch (InterruptedException e) {
+            LOGGER.error("", e);
```
xingfudeshi: Please set a Log tag.

test/src/test/java/io/seata/core/rpc/netty/v1/ProtocolV1CodecTest.java
```diff
@@ -0,0 +1,105 @@
...
+                service1.execute(() -> {
+                    while (tag.getAndIncrement() < runTimes) {
+                        try {
+                            Future future = client.sendRpc(head, body);
+                            RpcMessage resp = (RpcMessage) future.get(10, TimeUnit.SECONDS);
+                            if (resp != null) {
+                                success.incrementAndGet();
+                            }
+                        } catch (Exception e) {
+                            LOGGER.error("", e);
```
xingfudeshi: Please set a Log tag.

PR #1356, opened 2019-07-28
optimize sql join

andyqian:

optimize sql join


rm-datasource/src/main/java/io/seata/rm/datasource/exec/DeleteExecutor.java
```diff
@@ -118,6 +93,27 @@ protected TableRecords beforeImage() throws SQLException {
         return beforeImage;
     }
 
+    private String beforeImageSQL(SQLDeleteRecognizer visitor,TableMeta tableMeta,ArrayList<Object> paramAppender){
```
ujjboy: Hi, could you format your code?

andyqian: of couse, it has been completed

rm-datasource/src/main/java/io/seata/rm/datasource/exec/SelectForUpdateExecutor.java
```diff
@@ -106,10 +89,7 @@ public T doExecute(Object... args) throws Throwable {
                                 pstPK.setObject(i + 1, paramAppender.get(i));
                             }
                         } else {
-                            for (int i = 1; i < paramAppenders.size(); i++) {
```
slievrly: why delete？

PR #1358, opened 2019-07-29
test: clear store db

cmonkey:


xingfudeshi:
Is it only used for Testing case?

cmonkey:
> Is it only used for Testing case?

Yes, delete the temporary directory generated during the test

rm-datasource/src/test/java/io/seata/rm/datasource/undo/BaseH2Test.java
```diff
@@ -22,11 +22,16 @@
...
+import org.h2.store.fs.FileUtils;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.mockito.Mockito;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
```
ujjboy: remove unused imports.

PR #1359, opened 2019-07-29
when globalTransactionScanner was initialized before dataSourceProxy，There's a problem that rm will send invalid data to tc though rm will send again。this problem will cause transaction rollback fail。

zfh1038030621:
when globalTransactionScanner was initialized before dataSourceProxy，There's a problem that rm will send invalid data to tc though rm will send again。this problem will cause transaction rollback fail。
The scene of globalTransactionScanner was initialized before dataSourceProxy：
1，debug 
image

2，my demo
github: https://github.com/zfh1038030621/dubbo-demo.git
service-consumer send data to tc
image

temporary solutions of my demo is to control initialization sequenc
image

Ⅱ. Does this pull request fix one issue?
Ⅲ. Why don't you add test cases (unit test/integration test)?
Ⅳ. Describe how to verify it
donot send invalid data to tc
Ⅴ. Special notes for reviews


zfh1038030621:
> Here resourceId never could be null

this picture tell us resourceId may be null，thought second request will send.
image

slievrly:
> > Here resourceId never could be null
> 
> this picture tell us resourceId may be null，thought second request will send.
> image

You can print the log if resourceId==null in this method. In this method, the resourceId cannot be empty. This log is reported in other places.

PR #902, opened 2019-04-25
Added oracle database support

15168326318:
重新提一次pr，和https://github.com/seata/seata/pull/649 的一样。

PR #1376, opened 2019-08-01
Modify the method to get the primary key, and then fix getting the oracle driver. On some versions Jackson reported an error during serialization

majiajue:

fixes #902


all/pom.xml
```diff
@@ -610,12 +631,13 @@
             </build>
             <distributionManagement>
                 <snapshotRepository>
-                    <id>oss_seata</id>
-                    <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+                    <id>nexus-snapshots</id>
```
slievrly: please check your pr files.

majiajue: Ok

PR #1394, opened 2019-08-02
bugfix: EnhancedServiceLoader may throw NoClassDefFoundError terminate class load

zhuyoufeng:
When class loader loads the extensions from META-INF.services, System may throw NoClassDefFoundError and the EnhancedServiceLoader will stop to load the other classes.


common/src/main/java/io/seata/common/loader/EnhancedServiceLoader.java
```diff
@@ -38,6 +24,15 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.net.URL;
+import java.util.*;
```
slievrly: forbid to import *

slievrly:
why not fail fast？

zhuyoufeng:
> why not fail fast

For example, a system use consul as registry server and doesn't dependent on other providers such as nacos , seata will load the the configuration provider from META-INF/services/io.seata.config.ConfigurationProvider line by line, and throw noclassdeffounderror then ternimate the extension loading because it cannot find some class such as nacos, but actually system doesn't need these classes.

image

common/src/main/java/io/seata/common/loader/EnhancedServiceLoader.java
zhuyoufeng: updated, thanks

slievrly:
> > why not fail fast
> 
> For example, a system use consul as registry server and doesn't dependent on other providers such as nacos , seata will load the the configuration provider from META-INF/services/io.seata.config.ConfigurationProvider line by line, and throw noclassdeffounderror then ternimate the extension loading because it cannot find some class such as nacos, but actually system doesn't need these classes.
> 
> image

EnhancedServiceLoader `class not found` means just like `io.seata.discovery.registry.nacos.NacosRegistryProvider` not found, but nacos-client not found. In which case, `io.seata.discovery.registry.nacos.NacosRegistryProvider` will not exist?

zhuyoufeng:
> > > why not fail fast
> > 
> > 
> > For example, a system use consul as registry server and doesn't dependent on other providers such as nacos , seata will load the the configuration provider from META-INF/services/io.seata.config.ConfigurationProvider line by line, and throw noclassdeffounderror then ternimate the extension loading because it cannot find some class such as nacos, but actually system doesn't need these classes.
> > image
> 
> EnhancedServiceLoader `class not found` means just like `io.seata.discovery.registry.nacos.NacosRegistryProvider` not found, but nacos-client not found. In which case, `io.seata.discovery.registry.nacos.NacosRegistryProvider` will not exist?


not NacosRegistryProvider, this class is OK.

NacosConfigurationProvider, should be this, because this class depend on the NacosException, as the attached image, or no need to modify the EnhancedClassLoader, just modify the class NacosConfigurationProvider will be fine.

image

image

slievrly:
@zhuyoufeng Thank you very much, I will reproduce this question.

slievrly:
@zhuyoufeng The basic problem I think is that we should not have specific third-party dependencies in the `*Provider` class. Just like throwing Exception here

zhuyoufeng:
@slievrly My pleasure ^_^, i closed this PR

PR #1396, opened 2019-08-03
bugfix:fix ClassNotFound problem for Nacos config Implementation.

xingfudeshi:
fix ClassNotFound problem for Nacos config Implementation.

#1394 
Thanks to @zhuyoufeng

### DIFF
--- a/config/seata-config-nacos/src/main/java/io/seata/config/nacos/NacosConfiguration.java
+++ b/config/seata-config-nacos/src/main/java/io/seata/config/nacos/NacosConfiguration.java
@@ -70,7 +70,7 @@ public class NacosConfiguration extends AbstractConfiguration<Listener> {
             try {
                 configService = NacosFactory.createConfigService(getConfigProperties());
             } catch (NacosException e) {
-                throw new RuntimeException(e.getMessage());
+                throw new RuntimeException(e);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


config/seata-config-nacos/src/main/java/io/seata/config/nacos/NacosConfiguration.java
```diff
@@ -47,13 +47,31 @@
...
+     * Instantiates a new Nacos configuration.
      */
-    public NacosConfiguration() throws NacosException {
+    public NacosConfiguration() {
         if (null == configService) {
-            configService = NacosFactory.createConfigService(getConfigProperties());
+            try {
+                configService = NacosFactory.createConfigService(getConfigProperties());
+            } catch (NacosException e) {
+                throw new RuntimeException(e.getMessage());
```
slievrly: throw e;

xingfudeshi: throw e won't work here.because NacosException is not a subclass of RuntimeException.

slievrly: throw new RuntimeException(e);

xingfudeshi: > throw new RuntimeException(e);

done

PR #1403, opened 2019-08-05
 Support auto proxy of DataSource.

xingfudeshi:
1.When detect the DataSource bean,will auto proxy it.Integration of spring will be simplified as much as possible.The only thing you need to do is define a bean of GlobalTransactionScanner.
2.Add a switch for spring.datasource.autoproxy.
3.Unify file.conf.

### DIFF
--- a/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
+++ b/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
@@ -15,16 +15,13 @@
  */
 package io.seata.spring.annotation;
 
-import java.lang.reflect.Method;
-import java.util.HashSet;
-import java.util.Set;
-
 import io.seata.common.util.StringUtils;
 import io.seata.config.ConfigurationFactory;
 import io.seata.core.rpc.netty.RmRpcClient;
 import io.seata.core.rpc.netty.ShutdownHook;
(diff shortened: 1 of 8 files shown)


xingfudeshi:
> Is it possible to add a switch?

Yes.I will do a config item to make this configurable.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -297,4 +303,20 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
         this.setBeanFactory(applicationContext);
     }
 
+    @Override
+    public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+        if (bean instanceof DataSource && ConfigurationFactory.getInstance().getBoolean(DATASOURCE_AUTOPROXY, false)) {
+            Map beansOfTypeMaps = applicationContext.getBeansOfType(DataSourceProxy.class);
+            if (beansOfTypeMaps.size() < 1 && applicationContext instanceof ConfigurableApplicationContext) {
```
slievrly: just support one DataSourceProxy？

xingfudeshi: > just support one DataSourceProxy？

For now,yes,but i will re-design it to support multi-datasource.

Issue #1400, opened 2019-08-05
branchRegister error，Data too long for column


dupengcheng:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

missing exception detail , no log ,no throw,

异常在这里被吞掉了，很不利于在线排错

![图片](https://user-images.githubusercontent.com/11847818/62437697-e22af200-b776-11e9-9e5a-fb165705aa90.png)

### Ⅱ. Describe what happened
"Data too long for column  " ，It's a simple question，but without any exception info online

![图片](https://user-images.githubusercontent.com/11847818/62438186-dccea700-b778-11e9-8372-f19d6d9a88a9.png)

好费劲的在测试环境开启remote  debug 才找到原因，希望能加一行日志

### Ⅲ. Describe what you expected to happen

add error log 

### Ⅵ. Environment:

- JDK version : 1.8
- OS :  linux  
- Others:  base image openjdk:8

lightClouds917:
What size of the lockkeys ?

dupengcheng:
varchar(500)

---
my team extend executeBatch  method，we set lock_key  longtext ，but to this issue just add log is fine.

这个问题，我认为加一行日志就好，方便问题定位，lock_key默认值一般就够了。因为我们扩展实现了executeBatch ，所以我们设置这个key为 longtext ，但我不建议其他人这么设置，数据库大字段操作不易优化。

dupengcheng:
fine
@lightClouds917
optimize:add error log for addBranchSession for seata#1400

PR #1404, opened 2019-08-05
Optimize:optimize BranchStatus annotation

lightClouds917:
Ⅰ. Describe what this PR did
Optimize BranchStatus.java annotation,the annotation is very messy.

### DIFF
--- a/server/src/main/java/io/seata/server/coordinator/DefaultCore.java
+++ b/server/src/main/java/io/seata/server/coordinator/DefaultCore.java
@@ -82,6 +82,7 @@ public class DefaultCore implements Core {
                 globalSession.addBranch(branchSession);
             } catch (RuntimeException ex) {
                 branchSession.unlock();
+                LOGGER.error("Failed to add branchSession to globalSession : " + ex);
                 throw new TransactionException(FailedToAddBranch);
             }
             return branchSession.getBranchId();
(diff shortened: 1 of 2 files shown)


lightClouds917:
the second commit id to add log for  issue #1400

server/src/main/java/io/seata/server/coordinator/DefaultCore.java
```diff
@@ -82,6 +82,7 @@ public Long branchRegister(BranchType branchType, String resourceId, String clie
                 globalSession.addBranch(branchSession);
             } catch (RuntimeException ex) {
                 branchSession.unlock();
+                LOGGER.error("Failed to add branchSession to globalSession : " + ex);
```
slievrly: +ex ？

lightClouds917: Just add error log,  easy to find problems,The real cause is still being investigated

xingfudeshi: Please use LOGGER.error("Failed to add branchSession to globalSession:{}",ex.getMessage(),ex);

lightClouds917: Ok,Thank for your patience,I have modified it,but,I read a lot of error logs in seata,The style is not same.for example:
LOGGER.error("transaction[" + tx.getXid() + "] retry fetch status times exceed the limit [" + RETRY_MAX_TIMES + " times]");
LOGGER.error("fetch GlobalTransaction status error", e);
LOGGER.error("error occurred while getting value{}", e.getMessage());
LOGGER.error("not support metrics exporter type: " + exporterTypeName, exx);
......
they all used at catch{LOGGER.erro},unify the writing style will be better.

xingfudeshi: Yes.The problem does exist,will optmize in the future.

PR #1419, opened 2019-08-07
bugfix: if the connection is autocommit=false will cause fail to delete

### LINKED ISSUE #1420: if the connection is autocommit=false will cause fail to delete

### Ⅰ. Issue Description

if druid set  `defaultAutoCommit=false` will cause fail to delete expired undo_log
image

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:

bugfix: if the connection is autocommit=false will cause fail to delete

fix #1420

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
+++ b/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
@@ -15,6 +15,11 @@
  */
 package io.seata.rm;
 
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.Calendar;
+import java.util.Date;
+
 import io.seata.core.model.BranchType;
 import io.seata.core.model.ResourceManager;
 import io.seata.core.protocol.transaction.UndoLogDeleteRequest;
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -54,10 +54,21 @@ public void handle(UndoLogDeleteRequest request) {
...
             int deleteRows;
             do {
-                deleteRows = UndoLogManager.deleteUndoLogByLogCreated(logCreatedSave, dataSourceProxy.getDbType(), LIMIT_ROWS, conn);
+                deleteRows = UndoLogManager.deleteUndoLogByLogCreated(logCreatedSave, dataSourceProxy.getDbType(),
+                    LIMIT_ROWS, conn);
             } while (deleteRows == LIMIT_ROWS);
-        }  catch (Exception e) {
-            LOGGER.warn("Failed to get connection for delete undolog ", e);
+            if (!conn.getAutoCommit()) {
+                conn.commit();
```
github-ygy: how about commit in while ？

slievrly: @github-ygy fixed.

PR #1442, opened 2019-08-11
modify the resourceGroupId of TCCResource

tswstarplanet:
modify the group resource id of TCCResource from 'DEFAULT' to 'TCC', this can avoid the conflict between resource ids of TCCResource and DataSourceProxy


slievrly:
@tswstarplanet Can you describe the problem that occurred?

tswstarplanet:
> @tswstarplanet Can you describe the problem that occurred?

the default resourceGroupId is duplicated with that of DataSourceProxy. In some cases, it may cause problems. Such as use the resourceGroupId as key to cash the Resource in a map.

wangliang181230:
> 
> 
> > @tswstarplanet Can you describe the problem that occurred?
> 
> the default resourceGroupId is duplicated with that of DataSourceProxy. In some cases, it may cause problems. Such as use the resourceGroupId as key to cache the Resource in a map.

Key will only be `resourceId`, not `resourceGroupId`.

Issue #1434, opened 2019-08-09
Time type precision causes post mirroring and current mirroring to match mismatched transactions cannot be rolled back


lirengui:
- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
时间类型精度导致后置镜像和当前镜像比对不匹配事务无法回滚

### Ⅱ. Describe what happened
当业务表字段类为timestamp时
  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅲ. Describe what you expected to happen
支持timestamp类型时间精度

![image](https://user-images.githubusercontent.com/26484302/62761019-0f014100-bab8-11e9-997d-1018e187abbc.png)

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

l81893521:
i will fix this problem

PR #1443, opened 2019-08-11
bugfix: serialize the type of timestamp lost nano value

l81893521:
Jackson would use DateSerializer to handle java.sql.Timestamp(Timestamp extend from Date). So when java.sql.Timestamp has nano second, the digits would lost(java.util.Date dost not support nano second)

I provide customize serializer and deserializer. When serialize java.sql.Timestamp I convert it to a long array include fastTime and nanoSecond, look like this
```
{"@class":"io.seata.rm.datasource.sql.struct.Field","name":"timestamp_type","keyType":"NULL","type":93,"value":["java.sql.Timestamp",[1565405366926,926554000]]}
```
finally use it rebuild java.sql.Timestamp when deserialize.

#1434


rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/JacksonUndoLogParser.java
```diff
@@ -78,4 +105,46 @@ public BranchUndoLog decode(byte[] bytes) {
...
+        @Override
+        public Timestamp deserialize(JsonParser p, DeserializationContext ctxt) {
+            if(p.isExpectedStartArrayToken()){
+                ArrayNode arrayNode = null;
+                try {
+                    arrayNode = p.getCodec().readTree(p);
+                } catch (IOException e) {
+                    LOGGER.error("deserialize java.sql.Timestamp error : {}", e.getMessage(), e);
+                }
+                Timestamp timestamp = new Timestamp(arrayNode.get(0).asLong());
```
jsbxyyx: arrayNode may appear NPE

rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/JacksonUndoLogParser.java
```diff
@@ -78,4 +105,46 @@ public BranchUndoLog decode(byte[] bytes) {
         }
     }
 
+    private static class TimestampSerializer extends JsonSerializer<Timestamp> {
+
+        @Override
+        public void serializeWithType(Timestamp timestamp, JsonGenerator gen, SerializerProvider serializers, TypeSerializer typeSerializer) throws IOException {
+            gen.setCurrentValue(timestamp);
```
slievrly: is necessary？

rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/JacksonUndoLogParser.java
slievrly: should it be in the catch block?

l81893521: yes, thanks for correct my mistake

rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/JacksonUndoLogParser.java
l81893521: I will try to check the api to think is it necessary.

PR #1448, opened 2019-08-12
refactor: executor class duplicate code

andyqian:
remove duplicate code

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/SelectForUpdateExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/SelectForUpdateExecutor.java
@@ -15,18 +15,19 @@
  */
 package io.seata.rm.datasource.exec;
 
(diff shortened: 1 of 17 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/exec/SelectForUpdateExecutor.java
```diff
@@ -73,70 +70,28 @@ public T doExecute(Object... args) throws Throwable {
...
+                    statementProxy.getConnectionProxy().appendLockKey(lockKeys);
+                } else {
+                    throw new RuntimeException("Unknown situation!");
                 }
-            }
 
+                break;
+            }
+        } catch (LockConflictException lce) {
+            conn.rollback(sp);
```
slievrly: should in while block？

andyqian: > should in while block？

done

Issue #1415, opened 2019-08-06
primary key not found when the primary key field used the symbol `


jiajun6416:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
- the primary field is mysql keyword, we have add " ` " to it. TableMeta::containsPK can not found primaryKey. for example
> insert into login_user (\`uid\`, name, \`status\`, createAt)` value (1, 'zhangsan', 1, 2019-08-06)  

![image](https://user-images.githubusercontent.com/29995620/62556676-f8849b00-b8a7-11e9-8328-eb7b4635efb8.png)

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:
I will fix it.

lightClouds917:
> I will fix it.

I see the InsertExecutorTest.java，test colums like 
`    private static final String ID_COLUMN = "id";

    private static final String USER_ID_COLUMN = "user_id";

    private static final String USER_NAME_COLUMN = "user_name";
`
when use "'id'" you will replace  '  to nothing to fix it ?

cmonkey:
PR 1449

slievrly:
@jiajun6416 can you verify this pr？

zjinlei:
fixed in 1.0.0

PR #1449, opened 2019-08-13
bugfix: primary key not found when the primary key field used the symbol (#1415)

cmonkey:


rm-datasource/src/test/java/io/seata/rm/datasource/sql/struct/TableMetaTest.java
```diff
@@ -36,8 +36,7 @@
 import com.alibaba.druid.util.jdbc.ResultSetMetaDataBase;
 import io.seata.rm.datasource.DataSourceProxy;
 
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
```
jsbxyyx: not use *

PR #1452, opened 2019-08-13
bugfix: ensure select for update retry correctly

ggndnn:
In `SelectForUpdateExecutor`, select-for-update statement is not in global locking retry loop, so the data read is not necessarily globally committed. This PR ensue select-for-update retry correctly.


slievrly:
@ggndnn good found. Does it work if the isolation level is set to repeatable read?

ggndnn:
@slievrly I think it works. Because select-for-update statement returns current latest data in db, event set to repeatable read level, not like normal select statement who may query data from snapshot.

ggndnn:
```
catch (LockConflictException lce) {
    conn.rollback(sp);
    lockRetryController.sleep(lce);
}
```

slievrly:
> ```
> catch (LockConflictException lce) {
>    conn.rollback(sp);
>    lockRetryController.sleep(lce);
> }
> ```

So, I think conn.rollback(sp) is executed when the last retry is unsuccessful, instead of going to rollback every time.

slievrly:
@ggndnn After considering I think your plan is more appropriate. Because if I follow my plan, I got the X lock of the database and didn't get the global lock. At the same time, another distributed transaction gets the global lock and releases the database lock. When it wants to perform the rollback, it will cause a deadlock.

ggndnn:
@slievrly Thanks for your review. 
One question, I did not find global lock here, only database lock, any suggestion? I think conn.rollback(sp) is executed every time when `checkLock` throws LockConflictException. Is my understanding correct?

slievrly:
> @slievrly Thanks for your review.
> One question, I did not find global lock here, only database lock, any suggestion? I think conn.rollback(sp) is executed every time when `checkLock` throws LockConflictException. Is my understanding correct?

According to the currently modified code, this understanding is correct. The current `select for update` only checks that the lock does not occupy the lock. In the future, it is necessary to do a lock strategy to keep the behavior consistent with the local transaction.

Issue #1451, opened 2019-08-13
DataCompareUtils.isFieldEquals can't compare byte array.


wang-liqiang:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
Got branchRollback failed reason [2019387415/192.168.0.55:8091:2019387410 Has dirty records when undo.]

### Ⅱ. Describe what happened

Here is the compare log and the actual content of byte arrays, they are identical:

```
[...]

PR #1454, opened 2019-08-13
bugfix: DateCompareUtils can not compare byte array

l81893521:
use deep equals to compare byte array

#1451


rm-datasource/src/main/java/io/seata/rm/datasource/DataCompareUtils.java
```diff
@@ -73,8 +74,11 @@ public static boolean isFieldEquals(Field f0, Field f1) {
                             }
                             boolean result = f0.getValue().equals(f1.getValue());
```
sharajava: why not deepEquals directly?

l81893521: changed. Thanks for correct my mistake.

PR #1460, opened 2019-08-14
feature:gzip compressor

jsbxyyx:
feature gzip compressor


xingfudeshi:
@jsbxyyx CI is failed.

jsbxyyx:
@xingfudeshi 1564 cause CI failed.

l81893521:
I think if you can provide the unit test for CompressorFactory, that s better

core/src/main/java/io/seata/core/compressor/CompressorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+        COMPRESSOR_MAP.put(CompressorType.NONE, new NoneCompressor());
+    }
+
+    /**
+     * Get compressor by code.
+     *
+     * @param code the code
+     * @return the compressor
+     */
+    public static synchronized Compressor getCompressor(byte code) {
```
zjinlei: while use synchronized?

jsbxyyx: avoid repeatedly load Compressor class.

Issue #1438, opened 2019-08-09
LockKeyConflict appears when the concurrency is high.


wudgaby:
- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.
![image](https://user-images.githubusercontent.com/234605/62775397-1afdfa80-bada-11e9-9c29-514f14a36c0f.png)

in seata server
![image](https://user-images.githubusercontent.com/234605/62851266-1cb50180-bd18-11e9-9202-c3f9045bbd3d.png)

### Ⅰ. Issue Description
测试并发数30时, 分布式事务会出现TransactionException LockKeyConflict的问题.

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
io.seata.rm.datasource.exec.LockConflictException: null
	at io.seata.rm.datasource.ConnectionProxy.recognizeLockKeyConflictException(ConnectionProxy.java:128)
	at io.seata.rm.datasource.ConnectionProxy.processGlobalTransactionCommit(ConnectionProxy.java:179)
	at io.seata.rm.datasource.ConnectionProxy.commit(ConnectionProxy.java:156)
	at org.springframework.jdbc.datasource.DataSourceTransactionManager.doCommit(DataSourceTransactionManager.java:329)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:746)
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:714)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:533)
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:304)
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:98)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:688)
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :jdk8
- OS :windows
- Others: seata0.7.1

PR #1469, opened 2019-08-16
optimize: reduce the number of lock conflict exception

ggndnn:
Move lock retry loop into ConnectionProxy.commit, that the number of LockKeyConflict can be significantly reduced whether auto-commit is true or false. Related issue #1438.

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
@@ -15,9 +15,6 @@
  */
 package io.seata.rm.datasource.exec;
 
-import java.sql.SQLException;
-import java.sql.Statement;
-
 import io.seata.rm.datasource.AbstractConnectionProxy;
 import io.seata.rm.datasource.ConnectionProxy;
 import io.seata.rm.datasource.StatementProxy;
@@ -26,13 +23,17 @@ import io.seata.rm.datasource.sql.struct.TableRecords;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.concurrent.Callable;
+
 /**
  * The type Abstract dml base executor.
  *
- * @author sharajava
- *
  * @param <T> the type parameter
  * @param <S> the type parameter
+ * @author sharajava
  */
 public abstract class AbstractDMLBaseExecutor<T, S extends Statement> extends BaseTransactionalExecutor<T, S> {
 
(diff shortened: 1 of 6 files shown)


slievrly:
@ggndnn I am very interested because we always have different ideas. When a LockConflictException occurs, it indicates that there are other distributed transactions that are executing that hold the same data primary key. We define the current distributed transaction as A and another distributed transaction as B. A holds the database lock and wants to get the global lock, and B holds the global lock. If at this point B wants to roll back this data in the second phase of the distributed transaction, it will try to acquire the database lock. According to your code, A will hold the database lock for a longer time. At this time, B may trigger the lock wait timeout exception and perform a rollback retry. We need to evaluate this.

ggndnn:
@slievrly It's indeed interesting to have discussions with you, the atmosphere here is good， different ideas always make things more clear. Thank you. 

Before I make this PR, I referred to seata document in wiki, e.g.  

I think tx2 is A which you just mentioned and tx1 is B. I understand your worries. I also agree that we should make `quick fail` to avoid waiting too much. But `rollback` is not what we preferred, we prefer `commit`, according to issue 1438, we got exception immediately, may be just a moment we can get a success transaction. I think this PR is more in line with the figure in wiki. Is that a formal design?

I agree that we should evaluate this.  What can we do with this PR?

slievrly:
@ggndnn I am very happy to discuss this with you. I understand what you mean above. By the way, is it appropriate for us to communicate through DingTalk?

ggndnn:
@slievrly I joined DingTalk group.

slievrly:
@ggndnn Do you want to add a switch here as we communicate?

ggndnn:
@slievrly Ok, I'll add a switch.

ggndnn:
@slievrly I added a switch. Please have a look whether it's appropriate.

rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
```diff
@@ -83,34 +84,23 @@ protected T executeAutoCommitFalse(Object[] args) throws Exception {
...
-
+            return new LockRetryPolicy(connectionProxy.getTargetConnection()).execute(() -> {
+                T result = executeAutoCommitFalse(args);
+                connectionProxy.commit();
+                return result;
+            });
         } catch (Exception e) {
             // when exception occur in finally,this exception will lost, so just print it here
             LOGGER.error("execute executeAutoCommitTrue error:{}", e.getMessage(), e);
+            connectionProxy.rollback();
```
slievrly: just used for LOCK_RETRY_POLICY_BRANCH_ROLLBACK_ON_CONFLICT=false

ggndnn: Should it be okay to do `ConnectionProxy#rollback` in both cases?

slievrly: > Should it be okay to do `ConnectionProxy#rollback` in both cases?

there is no problem. I just mark ,when   LOCK_RETRY_POLICY_BRANCH_ROLLBACK_ON_CONFLICT=true ,it  already rollback , never run here.

Issue #1480, opened 2019-08-17
TableMetaCache multiple data source table metadata cache problem


zhaoxuguang666:
### Ⅰ. Issue Description
TableMetaCache在多数据源场景下,按数据源刷新所有表的信息，有同名表，TableMeta可能会来回变化，而且会经常在一个数据源上查找其他数据源的表，经常打印找不到表的异常日志

### Ⅱ. Describe what happened
2019-08-17 16:22:38,540 ERROR [TableMetaCache] get table meta error:Table 'test_db_a.test1_table' doesn't exist

l81893521:
Thank you, I will fix this problem.

l81893521:
@zhaoxuguang666; You problem have been fixed.

cbbaaa1989:
@l81893521 i have the same problem ,when the fix will be published? thank you

l81893521:
@cbbaaa1989 merged https://github.com/seata/seata/pull/1481

13567436138:
```
2019-09-27 17:20:59.869 ERROR 8928 --- [leMetaChecker_1] i.s.r.d.sql.struct.TableMetaCache        : get table meta error:Table 'stocklogdb.sku_lockinfos' doesn't exist

com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Table 'stocklogdb.sku_lockinfos' doesn't exist
	at sun.reflect.GeneratedConstructorAccessor49.newInstance(Unknown Source) ~[na:na]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[na:1.8.0_162]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[na:1.8.0_162]
	at com.mysql.jdbc.Util.handleNewInstance(Util.java:425) ~[mysql-connector-java-5.1.48.jar:5.1.48]
	at com.mysql.jdbc.Util.getInstance(Util.java:408) ~[mysql-connector-java-5.1.48.jar:5.1.48]
	at com.mysql.jdbc.SQLError.createSQLException(SQLError.java:944) ~[mysql-connector-java-5.1.48.jar:5.1.48]
	at com.mysql.jdbc.MysqlIO.checkErrorPacket(MysqlIO.java:3933) ~[mysql-connector-java-5.1.48.jar:5.1.48]
[...]

PR #1481, opened 2019-08-17
bugfix: TableMetaCache refresh problem in multiple datasource

l81893521:
Fix TableMetaCache refresh problem in multiple datasource.

#1480

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
@@ -125,8 +125,10 @@ public class TableMetaCache {
         try {
             conn = dataSource.getConnection();
             stmt = conn.createStatement();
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
```diff
@@ -105,7 +103,9 @@ public static void refresh(final DataSourceProxy dataSourceProxy){
                 if (tableMeta == null){
                     LOGGER.error("get table meta error");
                 }
-                if (!tableMeta.equals(entry.getValue())){
+
+                String key = getCacheKey(dataSourceProxy, entry.getValue().getTableName());
```
slievrly: Why not put the conditions ahead?

l81893521: Thanks for correct my mistake.

PR #1482, opened 2019-08-17
bugfix: PreparedStatement insert, params without prepared statement

### LINKED ISSUE #1470: build after-image sql Error

### Ⅰ. Issue Description
95)@T 8~}9{6}_Y`EXKAO2
$Y76D V~$MOBWVA V2M7N0

CREATE TABLE `test1`.`order_tbl`  (
  `order_id` int(11) NOT NULL AUTO_INCREMENT,
  `order_num` int(11) NOT NULL,
  `order_date` timestamp(0) NOT NULL DEFAULT CURRENT_TIMESTAMP,
  PRIMARY KEY (`order_id`) USING BTREE
) ENGINE = InnoDB AUTO_INCREMENT = 4 CHARACTER SET = utf8 COLLATE = utf8_general_ci ROW_FORMAT = Dynamic;

when i use sql 'insert into order_tbl(order_id,order_num,order_date) values (1, 100, ?) ',  it will throw an exception, then i check the exception and i find that if i have no value of primary key, building after-image sql will get an unnecessary 'where' and then throw exception

and what does cycleNums means?

update:
i see that i must use parameter list to build sql and seata use cycleNums to check the value of primary key

how can i just use sql without no parameters? like insert into order_tbl(order_id,order_num,order_date) values (1, 100, '2019-08-16') ?

seata version: 0.8.0-snapshot

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

zjinlei:
…ements

fixes #1470

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
@@ -111,6 +111,7 @@ public class InsertExecutor<T, S extends Statement> extends AbstractDMLBaseExecu
                 if (firstPkValue.equals(QUESTION_MARK)) {
                     pkValues = preparedStatementProxy.getParamsByIndex(firstPkIndex);
                 } else {
+                    pkValues = new ArrayList<>();
                     pkValues.add(firstPkValue);
                 }
             } else {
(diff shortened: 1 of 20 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
```diff
@@ -95,25 +95,28 @@ protected boolean containsPK() {
...
             } else {
-                pkValues = pkIndexs.stream().map(pkIndex -> paramters[pkIndex].get(0)).collect(Collectors.toList());
+                //some parameters are Prepared Statements: (1, 100, ?) or (?, 100, ?)
+                // or all parameters are Immediate Statements
+                List<List<Object>> insertRows = recognizer.getInsertRows();
+                //pk is Prepared Statements
+                if (insertRows.get(0).get(pkIndex).equals("?")) {
+                    pkValues = preparedStatementProxy.getParamsByIndex(pkIndex);
+                } else {
+                    int finalPkIndex = pkIndex;
```
slievrly: How to display the variable named final?

Issue #1483, opened 2019-08-19
Configuring schedule check table meta


l81893521:
## Why you need it?
 I think configuring the schedule check table meta will be better. 

## How it could be?
I will provide a pr and support this.

## Other related information

PR #1493, opened 2019-08-19
optimize: add table meta checker switch

l81893521:
Configuring table meta checker is done. But I m wondering is it necessary configuring the TABLE_META_CHECKER_INTERVAL as well?

#1483

### DIFF
--- a/server/src/main/resources/file.conf
+++ b/server/src/main/resources/file.conf
@@ -47,6 +47,8 @@ client {
     retry.times = 30
   }
   report.retry.count = 5
+  #schedule check table meta
+  table.meta.check.enable = true
 }
 
 ## transaction log store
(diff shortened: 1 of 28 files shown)


server/src/main/resources/file.conf
```diff
@@ -118,6 +118,8 @@ transaction {
   #schedule delete expired undo_log in milliseconds
   undo.log.delete.period = 86400000
   undo.log.table = "undo_log"
+  #schedule check table meta
```
slievrly: This is a client configuration item.

l81893521: Thanks, I m changed

slievrly: > Thanks, I m changed

this file.conf located in server module, but server not use this configuration item.

l81893521: I move it to seata-config-core\src\main\resources\file.conf

l81893521:
> add this configuration item in nacos-config.txt

ok

PR #1523, opened 2019-08-23
Refactor UndoLogManager into AbstractUndoLogManger to support more Db Type in the future

CharmingRabbit:
Add an abstract parent class AbstractUndoLogManager for UndoLogManger and UndoLogManagerOracle to encapsulate common logic in order to support multiple other types of DB in the future.
Add UndoLogManagerFactory as a unified access to UndoLogManager,
Add an resource configuration files(io.seata.rm.datasource.undo.AbstractUndoLogManager), to dynamically loading  concrete UndoLogManager of different Db types.
Remove the duplicate code of the original two UndoLog Managers and the redundant judgment code of each user of the original UndoLog Manager.

Now,Instead of using UndoLogManager (now it is UndoLogManagerMySQL) and UndoLogManagerOracle, just use UndoLogManagerFactory to get the specific UndoLogManager.

Explanation of why we use the resource configuration file: When we use and validate seata, the database we use is PG. Since the community only provides support for MySQL and oracle, we need to implement AT mode support for PG database ourselves, but this part of the code belongs to our extension of db, we think that using the resource configuration file should be able to better decouple our code from the original official code, so we not only abstracted the base class AbstractUndoLogManager, but also added an configuration file under the resources directory.


zjinlei:
Please restore the code changed by the format to highlight your changes.

zjinlei:
Thank you for your contribution!
seata code structure has changed dramatically during our waiting for your feedback, so we will close this pr.

Issue #1529, opened 2019-08-26
NPE when no index in the table


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

NullPointException show in TableMetaCache.java when no any index table

### Ⅱ. Describe what happened

```
```

### Ⅲ. Describe what you expected to happen
I will provide a pr and give some error log to tell the user, table in the database have to give one index or primary index at least.

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

PR #1530, opened 2019-08-26
optimize: throw Exception when no index in the table

l81893521:
git some tips when no index in the table cause NPE in TableMetaCache.java.

#1529

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
@@ -279,7 +279,7 @@ public class TableMetaCache {
 
             }
         }
-        if(tm.getAllIndexes().size() == 0){
+        if(tm.getAllIndexes().isEmpty()){
             throw new ShouldNeverHappenException("Could not found any index in the table: " + tableName);
         }
         IndexMeta index = tm.getAllIndexes().get(indexName);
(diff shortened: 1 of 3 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
```diff
@@ -279,6 +279,9 @@ private static TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseM
 
             }
         }
+        if(tm.getAllIndexes().size() == 0){
```
kezhenxu94:
```suggestion
        if(tm.getAllIndexes().isEmpty()) {
```

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCacheOracle.java
```diff
@@ -191,6 +191,9 @@ private static TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseM
                 index.setIndextype(IndexType.PRIMARY);
             }
         }
+        if(tm.getAllIndexes().size() == 0){
```
kezhenxu94:
```suggestion
        if(tm.getAllIndexes().isEmpty()) {
```

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
l81893521: Thanks

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCacheOracle.java
l81893521: Thanks

PR #1533, opened 2019-08-27
Refactor of the logics of Multi-configuration Isolation

### LINKED ISSUE #1532: Refactor of the logics of Multi-configuration Isolation

### Ⅰ. Issue Description
For the current logics of loading different configuration,it's a bit complicated to understand.So,i'm gonna improve it,and make it clearly.So,the new logics will be like:
At first,will try to read the name from JVM parameters(when use -Dxxx=xx),if reads nothing,will try to read the name from system environment,if reads nothing again,then will ues default configuration:file.conf,registry.conf.
For example,if we have a testing environment called "test",then these conf files should be in the following format:
`
file-test.conf
registry-test.conf
`

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

xingfudeshi:
For the current logics of loading different configuration,it's a bit complicated to understand.So,i'm gonna improve it,and make it clearly.So,the new logics will be like:
At first,will try to read the name from JVM parameters(when use -Dxxx=xx),if reads nothing,will try to read the name from system environment,if reads nothing again,then will ues default configuration:file.conf,registry.conf.
For example,if we have a testing environment called "test",then these conf files should be in the following format:
`
file-test.conf
registry-test.conf
`

fixes #1532

### DIFF
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
@@ -15,10 +15,10 @@
  */
 package io.seata.config;
 
-import java.util.Objects;
-
 import io.seata.common.exception.NotSupportYetException;
 import io.seata.common.loader.EnhancedServiceLoader;
+
+import java.util.Objects;
 
 /**
  * The type Configuration factory.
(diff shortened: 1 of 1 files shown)


config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -31,25 +31,21 @@
     private static final String REGISTRY_CONF_SUFFIX = ".conf";
     private static final String ENV_SYSTEM_KEY = "SEATA_CONFIG_ENV";
```
slievrly: how about `seata_env`?

xingfudeshi: > how about `seata_env`?

Sure,i will unity the name.

xingfudeshi: > how about `seata_env`?

it's done.I will update wiki when this PR merged.

PR #1542, opened 2019-08-27
optimize: AbstractSessionManager debug log is wrong

### LINKED ISSUE #1538: AbstractSessionManager#updateBranchSessionStatus debug log is wrong

### Ⅰ. Issue Description

image

日志输出的内容是否应该分别修改为BRANCH_UPDATE和BRANCH_REMOVE？

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

zjinlei:

fixes #1538

### DIFF
--- a/server/src/main/resources/registry.conf
+++ b/server/src/main/resources/registry.conf
@@ -51,7 +51,6 @@ config {
   nacos {
     serverAddr = "localhost"
     namespace = "public"
-    cluster = "default"
   }
   consul {
     serverAddr = "127.0.0.1:8500"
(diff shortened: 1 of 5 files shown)


server/src/main/resources/registry.conf
```diff
@@ -51,7 +51,6 @@ config {
   nacos {
     serverAddr = "localhost"
     namespace = "public"
-    cluster = "default"
```
xingfudeshi: Please remove this config item from all registry.conf files.

PR #1559, opened 2019-08-29
optimize: delay to delete the expired undo log

### LINKED ISSUE #1464: When the seata-server starts, it should takes a delay to delete the expired undo log

### Ⅰ. Issue Description
When the seata-server starts, it should takes a delay to delete the expired undo log. when the 
delete the expired undo log schedule run the first，rm channel was not registered.
image

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

jsbxyyx:
optimize delay to delete the expired undo log

fixes #1464

### DIFF
--- a/server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
+++ b/server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
@@ -113,6 +113,11 @@ public class DefaultCoordinator extends AbstractTCInboundHandler
      */
     protected static final long UNDOLOG_DELETE_PERIOD = CONFIG.getLong(ConfigurationKeys.TRANSACTION_UNDO_LOG_DELETE_PERIOD, 24 * 60 * 60 * 1000);
 
+    /**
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -459,7 +459,7 @@ public void init() {
             } catch (Exception e) {
                 LOGGER.info("Exception undoLog deleting ... ", e);
             }
-        },0, UNDOLOG_DELETE_PERIOD,TimeUnit.MILLISECONDS);
+        },3 * 60 * 1000, UNDOLOG_DELETE_PERIOD, TimeUnit.MILLISECONDS);
```
xingfudeshi: define a constant.

jsbxyyx: fix done

PR #1569, opened 2019-08-30
bugfix:fix problem when using @TwoPhaseBusiness,

zhangfanghui8:

When registering Dubbo service with annotation @Service and using annotation @TwoPhaseBusiness, the method isTccAutoProxy should have returned true but false.


slievrly:
@zhangfanghui8 the RPC‘s service bean if called by inJVM why not mark as `@LocalTCC`?

zhangfanghui8:
> @zhangfanghui8 the RPC‘s service bean if called by inJVM why not mark as `@LocalTCC`?

yes. if the bean called by inJVM ，mark as @LocalTCC can Implementing TCC function（I tried it）.
But,if not called by in JVM，it‘s a remote dubbo service in other module，I don't think we should use localtcc (even localtcc also can solve the problem). I think remote Dubbo services should  be automatically recognized as TCC beans instead of annotating localtcc.

spring/src/main/java/io/seata/spring/util/TCCBeanParserUtils.java
```diff
@@ -15,14 +15,14 @@
...
 
-import java.lang.reflect.Method;
-
 import io.seata.rm.tcc.api.TwoPhaseBusinessAction;
 import io.seata.rm.tcc.remoting.Protocols;
 import io.seata.rm.tcc.remoting.RemotingDesc;
 import io.seata.rm.tcc.remoting.parser.DefaultRemotingParser;
 import org.springframework.context.ApplicationContext;
 
+import java.lang.reflect.Method;
```
zjinlei: restore, jdk import front
idea: Editor-Code Style-java-import-import layout

zhangfanghui8: You mean to restore the java.lang.reflect.Method's position？

zjinlei: yes

zhangfanghui8: it is ok

zjinlei:
Thanks

Issue #1567, opened 2019-08-30
Seat-server0.8.0 version integrated oracle problem


haoshichong:
mybatis集成oracle时，当表名小写需要加''标识，seata-Client报空指针异常且数据库回滚日志无改变

l81893521:
I will fix this problem

l81893521:
Could you paste you sql here?

slievrly:
insert into "order_tbl"  values  (#{id},#{userId}, #{commodityCode}, #{count}, #{money})

PR #1572, opened 2019-08-30
bugfix: get table meta in oracle when table name was lower case

l81893521:
Fix get table meta in oracle when table name was lower case

#1567


l81893521:
> can‘t get any tableMetaData:
> CREATE TABLE "testX" (
> "id" INT NOT NULL ,
> "name" varchar(32) NOT NULL,
> PRIMARY KEY ("id")
> )
> 
> jdbcTemplate.update("insert into "TEST"."testX" (id,name) values(1,'123')");
> 
> image

fixed, because schema have "double quotes".

PR #1581, opened 2019-09-03
bugfix: missing some length in GlobalSession and FileTransactionStoreManager

AjaxXu:
fix missing length in GlobalSession and FileTransactionStoreManager

No.

No need.

By logical

BUG


server/src/main/java/io/seata/server/store/file/FileTransactionStoreManager.java
```diff
@@ -416,9 +416,9 @@ private boolean writeDataFile(byte[] bs) {
         }
         ByteBuffer byteBuffer = null;
 
-        if (bs.length > MAX_WRITE_BUFFER_SIZE) {
+        if (bs.length + INT_BYTE_SIZE > MAX_WRITE_BUFFER_SIZE) {
```
slievrly: revert this file change，will duplicate with https://github.com/seata/seata/pull/1552/files

PR #1617, opened 2019-09-07
feature: add seata.config.name configuration

ggndnn:
Add seata.config.name configuration, registry.conf file name can be customized.


config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -31,22 +31,29 @@
     private static final String REGISTRY_CONF_SUFFIX = ".conf";
     private static final String ENV_SYSTEM_KEY = "SEATA_ENV";
     private static final String ENV_PROPERTY_KEY = "seataEnv";
-    /**
-     * the name of env
-     */
-    private static String envValue;
+
+    private static final String SYSTEM_PROPERTY_SEATA_CONFIG_NAME = "seata.config.name";
```
xingfudeshi: Can we add this to file.conf as a configuration item?

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -31,22 +31,29 @@
...
+
+    private static final String ENV_SEATA_CONFIG_NAME = "SEATA_CONFIG_NAME";
```
xingfudeshi: I guess our users may feel confuse with this name as there's already a envrionment variable called SEATA_ENV.As i said above,can we use a configuration item to control the file name of registry.conf?

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
ggndnn: I wanted to use a custom name of registry.conf, so added a system property `seata.config.name` here. I understand your meaning, but file.conf is configured in registry.conf. I think it's not possible to do this, is my understanding right?

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
ggndnn: I think SEATA_CONFIG_NAME does not conflict to SEATA_ENV, they have different meaning, SEATA_CONFIG_NAME is file name, and SEATA_ENV is profile name. Before I added this PR, I referred to how spring-boot custom application.properties. As spring did,  `seata.config.name` should be provided as a command line parameter, but system property is easier to implement here.

ggndnn:
@xingfudeshi Thanks for your review. Please have a look whether my explanation appropriate.

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
xingfudeshi: Sure.Your explanation makes sense.

Issue #1623, opened 2019-09-09
RegistryService server address list change events leak


BeiKeJieDeLiuLangMao:
- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
All kinds of RegistryService#lookup implement is：
1. get newest server address list
2. add change listener
3. when listener triggered，change server list

Now consider follow situation：
1. get newest server address list
2. server address list chagned（change events leak）
3. add change listener
4. no change event，do nothing

I know not all underlayer service is supportted to resolve it，but at least etcd3 support.
```java
//1.get all available registries
GetOption getOption = GetOption.newBuilder().withPrefix(buildRegistryKeyPrefix()).build();
GetResponse getResponse = getClient().getKVClient().get(buildRegistryKeyPrefix(), getOption).get();
//2.add to list
List<InetSocketAddress> instanceList = getResponse.getKvs().stream().map(keyValue -> {
    String[] instanceInfo = keyValue.getValue().toString(UTF_8).split(":");
    return new InetSocketAddress(instanceInfo[0], Integer.parseInt(instanceInfo[1]));
}).collect(Collectors.toList());
clusterAddressMap.put(cluster, instanceList);
//3.get revision of previous get request
long revision = getResponse.getHeader().getRevision();
//4.register listener based on previous revision, then all change events after that revision will be sent by etcd server
Watch watchClient = getClient().getWatchClient();
WatchOption watchOption = WatchOption.newBuilder().withPrefix(buildRegistryKeyPrefix()).withRevision(revision).build();
watchClient.watch(buildRegistryKeyPrefix(), watchOption, listener);
```
### Ⅱ. Describe what happened
Server address list change events leak

### Ⅲ. Describe what you expected to happen
At least event leak doesn't happen in etcd impl.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)
1. get newest server address list then start debug break point
2. server address list chagned
3. cancel break point, then add change listener
4. no change event，do nothing

### Ⅴ. Anything else we need to know?
You guys could use etcd3 revision.

### Ⅵ. Environment:

- JDK version :1.9
- OS : MacOS
- Others:

xingfudeshi:
Can you submit a PR to improve this?

BeiKeJieDeLiuLangMao:
@xingfudeshi 
Done.

PR #1629, opened 2019-09-10
bugfix: address list change events leak

BeiKeJieDeLiuLangMao:
Fix etcd registry service's server address list change events leak.

#1623 

No new api.

1. get newest server address list then start debug break point
2. change server address list
3. cancel break point, then add change listener
4. receive change event，refresh address list


xingfudeshi:
@BeiKeJieDeLiuLangMao You need to re-push these commits.

xingfudeshi:
Because you've just associated your github account with email.

PR #1657, opened 2019-09-16
optimize: eliminate the possibility of allocating too much direct memory

ggndnn:
In FileTransactionStoreManager, sometimes, data buffer size was not under control. For example when data size to write was bigger than writerBuffer:

```
...
ByteBuffer byteBuffer = null;

if (bs.length + 4 > MAX_WRITE_BUFFER_SIZE) {
    //allocateNew
    byteBuffer = ByteBuffer.allocateDirect(bs.length + 4);
} else {
...
```
Or in function findTimeoutAndSave:
```
...
ByteBuffer byteBuffer = ByteBuffer.allocateDirect(totalSize);
for (byte[] bytes : listBytes) {
    byteBuffer.putInt(bytes.length);
    byteBuffer.put(bytes);
}
...
```
Those buffers' size was out of control, unlimited, that may be a hidden danger. 

This PR makes fully use of the writerBuffer, eliminate the possibility of allocating too much memory.

Did.

### DIFF
--- a/server/src/main/java/io/seata/server/store/file/FileTransactionStoreManager.java
+++ b/server/src/main/java/io/seata/server/store/file/FileTransactionStoreManager.java
@@ -182,7 +182,7 @@ public class FileTransactionStoreManager extends AbstractTransactionStoreManager
                 return saveHistory();
             }
         } catch (Exception exx) {
-            LOGGER.error("writeSession error," + exx.getMessage());
+            LOGGER.error("writeSession error, {}", exx.getMessage(), exx);
             return false;
         } finally {
             writeSessionLock.unlock();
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/seata/server/store/file/FileTransactionStoreManager.java
```diff
@@ -215,46 +215,87 @@ private boolean saveHistory() throws IOException {
...
+        }
+        writeBuffer.putInt(dataLength);
+        bufferRemainingSize = writeBuffer.remaining();
+        int dataPos = 0;
+        while (dataPos < dataLength) {
+            int dataLengthToWrite = dataLength - dataPos;
+            dataLengthToWrite = Math.min(dataLengthToWrite, bufferRemainingSize);
+            writeBuffer.put(data, dataPos, dataLengthToWrite);
+            bufferRemainingSize = writeBuffer.remaining();
+            if (bufferRemainingSize <= 0) {
```
slievrly: ==0

ggndnn: Yes, it can't be < 0, just in case.

ggndnn: Committed.

server/src/main/java/io/seata/server/store/file/FileTransactionStoreManager.java
```diff
@@ -215,46 +215,87 @@ private boolean saveHistory() throws IOException {
...
-            LOGGER.error("save history data file error," + exx.getMessage());
+            LOGGER.error("save history data file error, {}", exx.getMessage(), exx);
             result = false;
         } finally {
             initFile(currFullFileName);
         }
         return result;
     }
 
+    private boolean writeDataFrame(byte[] data, int dataLength) {
```
jsbxyyx: The dataLength passed in by the caller is data.length, whether to remove the dataLength parameter，use data.length??

ggndnn: Function writeDataFrame is independent, has more functions than just specifying data as parameter, that could be more useful in future. I think no harm done to provide one more parameter dataLength. Do you agree?

jsbxyyx: I am no problem, It’s just that developer are looking a little doubt by now.

ggndnn: Ok, it's fine to remove dataLength parameter by now, clearer and more concise at least.

ggndnn: Committed.

Issue #1662, opened 2019-09-17
MySQL keywords should be supported as table and field names  with in the fetchSchemeInDefaultWay method of TableMetaCache.


jiangbin31:
### Ⅰ. Issue Description
MySQL keywords should be supported as table and field names  with in the fetchSchemeInDefaultWay method of TableMetaCache.

### Ⅱ. Describe what happened
if there is an exception, please attach the exception trace:
if we use MySQL keyword as the filed names or table names,will cause the following exception.

```
java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'order LIMIT 1' at line 1
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:120)
	at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:97)
	at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
	at com.mysql.cj.jdbc.StatementImpl.executeQuery(StatementImpl.java:1218)
	at com.alibaba.druid.filter.FilterChainImpl.statement_executeQuery(FilterChainImpl.java:2831)
	at com.alibaba.druid.filter.FilterAdapter.statement_executeQuery(FilterAdapter.java:2503)
	at com.alibaba.druid.filter.FilterEventAdapter.statement_executeQuery(FilterEventAdapter.java:302)
	at com.alibaba.druid.filter.FilterChainImpl.statement_executeQuery(FilterChainImpl.java:2828)
	at com.alibaba.druid.proxy.jdbc.StatementProxyImpl.executeQuery(StatementProxyImpl.java:221)
	at com.alibaba.druid.pool.DruidPooledStatement.executeQuery(DruidPooledStatement.java:291)
	at io.seata.rm.datasource.sql.struct.TableMetaCache.fetchSchemeInDefaultWay(TableMetaCache.java:131)
	at io.seata.rm.datasource.sql.struct.TableMetaCache.fetchSchema(TableMetaCache.java:119)
	at io.seata.rm.datasource.sql.struct.TableMetaCache.lambda$getTableMeta$0(TableMetaCache.java:75)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2337)
[...]

PR #1663, opened 2019-09-17
bugfix: get table meta fail when table name was keyword

l81893521:
Fix get table meta fail when table name was keyword

#1662

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
@@ -24,11 +24,8 @@ import java.sql.Statement;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
-
 import javax.sql.DataSource;
-
 import com.alibaba.druid.util.JdbcConstants;
-
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
```diff
@@ -121,11 +125,12 @@ private static TableMeta fetchSchemeInDefaultWay(DataSource dataSource, String t
         Connection conn = null;
         Statement stmt = null;
         ResultSet rs = null;
+        KeywordChecker keywordChecker = KeywordCheckerFactory.getKeywordChecker(JdbcConstants.MYSQL);
```
slievrly: best as a static field

l81893521: Thanks

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/TableMetaCache.java
```diff
@@ -53,6 +54,8 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(TableMetaCache.class);
 
+    private static final KeywordChecker keywordChecker = KeywordCheckerFactory.getKeywordChecker(JdbcConstants.MYSQL);
```
slievrly: remove final

l81893521: ok

PR #1664, opened 2019-09-17
bugfix: restore autocommit

zjinlei:
https://github.com/seata/seata-samples/issues/226
After the connection is used, it will be restored to automatic submission.


core/src/main/java/io/seata/core/store/db/LockStoreDataBaseDAO.java
```diff
@@ -183,6 +183,7 @@ public boolean acquireLock(List<LockDO> lockDOs) {
             }
             if (conn != null) {
                 try {
+                    conn.setAutoCommit(true);
```
slievrly: Connection pool can set the connection to default to autocommit=false

PR #1691, opened 2019-09-21
bugfix:fix can not access a member of DruidDataSourceWrapper 

### LINKED ISSUE #1690: GlobalTransactionScanner can not access a member of DruidDataSourceWrapper with modifiers "public"

### Ⅰ. Issue Description
When set support.spring.datasource.autoproxy = true,and druid-spring-boot-starter is used, it creates a DruidDataSourceWrapper instead of creating the original DruidDataSource directly, but DruidDataSourceWrapper does not use the "public" modifier to modify it, it only has access to the package.
image

image

### Ⅱ. Describe what happened

so,when we try to call it's method(or any other members),will cause the following exceptions:
```
Caused by: java.lang.IllegalAccessException: Class io.seata.spring.annotation.GlobalTransactionScanner can not access a member of class com.alibaba.druid.spring.boot.autoconfigure.DruidDataSourceWrapper with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Method.invoke(Method.java:491)
	at io.seata.spring.annotation.GlobalTransactionScanner.lambda$postProcessBeforeInitialization$0(GlobalTransactionScanner.java:321)
	at com.alibaba.druid.spring.boot.autoconfigure.DruidDataSourceWrapper$$EnhancerByCGLIB$$7bd61a2e.afterPropertiesSet(<generated>)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1837)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1774)
	... 121 common frames omitted
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

xingfudeshi:
fix can not access a member of DruidDataSourceWrapper with modifiers "public"

fixes #1690

I've done integration test.

### DIFF
--- a/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
+++ b/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
@@ -317,6 +317,7 @@ public class GlobalTransactionScanner extends AbstractAutoProxyCreator
                 if (null != m) {
                     return m.invoke(dataSourceProxy, args);
                 } else {
+                    method.setAccessible(true);
                     return method.invoke(bean, args);
                 }
             });


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -317,6 +317,7 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
                 if (null != m) {
                     return m.invoke(dataSourceProxy, args);
                 } else {
+                    method.setAccessible(true);
```
ggndnn: Shouldn't we restore the accessible? e.g. 
```
boolean oldAccessible = method.getAccessible();
method.setAccessible(true);
try {
    ...
} finally {
    method.setAccessible(oldAccessible);
}
```

xingfudeshi: Good idea.

xingfudeshi: > Shouldn't we restore the accessible? e.g.
> 
> ```
> boolean oldAccessible = method.getAccessible();
> method.setAccessible(true);
> try {
>     ...
> } finally {
>     method.setAccessible(oldAccessible);
> }
> ```

fixed

PR #1699, opened 2019-09-23
bugfix: fix use 'in' and 'between' in where condition for Oracle and Mysql

l81893521:
fix where condition like
```
delete from table where id in (1)
```
or
```
delete from table where id between 1 and 2
```

#1697 #1714


rm-datasource/src/main/java/io/seata/rm/datasource/sql/druid/oracle/OracleDeleteRecognizer.java
```diff
@@ -83,7 +85,15 @@ public String getWhereCondition(final ParametersHolder parametersHolder, final A
         }
         StringBuilder sb = new StringBuilder();
         OracleOutputVisitor visitor = super.createOracleOutputVisitor(parametersHolder, paramAppenderList, sb);
-        visitor.visit((SQLBinaryOpExpr) where);
+        if (where instanceof SQLBinaryOpExpr) {
```
slievrly: how about move to a common place such as   BaseRecognizer?

l81893521: maybe yes, but have to separate like getMysqlWhereCondition(expr) and getOracleWhereCondition(expr) due to different output visitor.

l81893521: fixed. I will change the unit test later

PR #1706, opened 2019-09-24
optimize:inner class may be static

cmonkey:


discovery/seata-discovery-etcd3/src/test/java/io/seata/discovery/registry/etcd/EtcdRegistryServiceImplTest.java
```diff
@@ -165,7 +165,7 @@ private ByteSequence buildRegistryKeyPrefix() {
     /**
      * etcd listener
      */
-    private class EtcdListener implements Watch.Listener {
+    private static class EtcdListener implements Watch.Listener {
```
xingfudeshi: This is may not required.The EtcdListener is only used for test.

xingfudeshi:
@cmonkey Thanks for your work,but these modifications may not be required so far.

PR #1707, opened 2019-09-24
optimize:change get charset

cmonkey:


core/src/main/java/io/seata/core/lock/LockMode.java
```diff
@@ -31,7 +31,7 @@
     /**
      * store the lock in seata's server
      */
-    REMOTE;
+    REMOTE
```
CoffeeLatte007: keep ;

cmonkey: Has been modified

Issue #1702, opened 2019-09-24
unit test failure on java 11 environment


Justice-love:
- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
unit test class: io.seata.server.event.DefaultCoreForEventBusTest

error message:
```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.042 s <<< FAILURE! - in io.seata.server.event.DefaultCoreForEventBusTest
[ERROR] test  Time elapsed: 1.041 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <1> but was: <2>
	at io.seata.server.event.DefaultCoreForEventBusTest.test(DefaultCoreForEventBusTest.java:82)
```

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

cmonkey:
和jdk 版本关系不大，将@Test 替换成RepeatedTest, 并给RepeatedTest 参数1000/10000, 此异常就容易重现了

ggndnn:
@Justice-love 
Is it based on the latest code of develop branch? Please refer to pr #1576. If so, we should consider more other reasons of this fail.

@cmonkey 
It was easy to reproduce the unit test fail after `RepeatedTest` used, because the test code did not clean objects correctly in each loop. The test case is not applicable to `RepeatedTest`. But I think it's better to release resources correctly in very test case, otherwise, there may be some unexpected situations. 

If add the code below, it's not easy to reproduce the error.

```
public void test() throws IOException, TransactionException, InterruptedException {
    ...
    SessionHolder.init(null);
    DefaultCoordinator coordinator = new DefaultCoordinator(null);
    coordinator.init();

    try {
        ...
    } finally {
        coordinator.destroy();
        SessionHolder.destory();
    }
}
```

ggndnn:
@Justice-love ok, thank you.

cmonkey:
通过在finally 块中调用destroy 方法，通过RepeatedTest执行100次，没有发现此异常，应该是解决掉了

ggndnn:
Please check #1710 for more information.

PR #1710, opened 2019-09-25
feature: add prefix counter for NamedThreadFactory

ggndnn:
Though PR #1576 is one reason of #1702, but it's obviously incomplete. I suspect that there are multiple `AsyncCommitting` threads executing at the same time causing this test failure.

The reason is that maven-surefire-plugin was used, but parameters `reuseForks` and `forkCount ` were left to be their default values. According to [maven-surefire-plugin doc](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html), only one VM was forked and then reused to execute all tests. 

The log below could prove this, notice the UUID used in two test cases:
```
[INFO] Running io.seata.server.UUIDGeneratorOverflowTest
[UUID 0] is: 4022933692
[UUID 1] is: 4022933693
[UUID 2] is: 4022933694
[UUID 3] is: 4022933695
[UUID 4] is: 4022933696
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 s - in io.seata.server.UUIDGeneratorOverflowTest
[INFO] Running io.seata.server.event.DefaultCoreForEventBusTest
2019-09-23 10:28:11.754 INFO [main]io.seata.server.coordinator.DefaultCore.begin:142 -Successfully begin global transaction xid = 172.17.0.1:0:4022933697
```

Because several test case in seata didn't release resources in time, for example, in test case `DefaultCoordinatorMetricsTest`, `DefaultCoordinator#init` was called, but never destroyed, that when `DefaultCoreForEventBusTest` started, there may be more than one asyncCommitting threads running.

It's not so easy to reproduce this test failure, so this PR is opened to verify this suspicion, just waiting for the next failure. I'm not sure whether this suspicion is correct, if it proves right, all test cases should be checked and try to release resources created correctly.

This suspicion proves correct, if we can get log looks like below, please notice the thread name:

```
2019-09-23 10:28:11.850 INFO [AsyncCommitting_1_1]io.seata.server.coordinator.DefaultCore.doGlobalCommit:245 -Global[172.17.0.1:0:4022933697] committing is successfully done.
2019-09-23 10:28:11.851 INFO [AsyncCommitting_2_1]io.seata.server.coordinator.DefaultCore.doGlobalCommit:245 -Global[172.17.0.1:0:4022933697] committing is successfully done.
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.042 s <<< FAILURE! - in io.seata.server.event.DefaultCoreForEventBusTest
[ERROR] test  Time elapsed: 1.041 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <1> but was: <2>
	at io.seata.server.event.DefaultCoreForEventBusTest.test(DefaultCoreForEventBusTest.java:82)
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   DefaultCoreForEventBusTest.test:82 expected: <1> but was: <2>
[INFO] 
[ERROR] Tests run: 90, Failures: 1, Errors: 0, Skipped: 1
```

**Update(2019/10/17)**:

As the issue #1702 has been solved, change to a feature PR, it's helpful to investigate threads related issues.


common/src/main/java/io/seata/common/thread/NamedThreadFactoryEx.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Based on NamedThreadFactory
+ *
+ * @author jimin.jm @alibaba-inc.com
+ * @author ggndnn
+ */
+public class NamedThreadFactoryEx implements ThreadFactory {
```
xingfudeshi: How about extend NamedThreadFactory directly?

ggndnn: It's not easy to implement NamedThreadFactoryEx by extending NamedThreadFactory. I didn't mean to change NamedThreadFactory before, after your reminder, I'd like to modify NamedThreadFactory directly, just add a counter for every prefix.

ggndnn: Committed.

xingfudeshi: actually i mean you can use NamedThreadFactoryEx to extend NamedThreadFactory,and override the newThread method.

xingfudeshi: The code you've committed previously was OK to me.what i mean is you can change the base class to NamedThreadFactory instead of implementing threadFactory directly.so,we can know this NamedThreadFactoryEx is only a extension for NamedThreadFactory for testing purpose.

xingfudeshi: And applying this counter to all places will bring performance costs.

xingfudeshi: And applying this counter to all places will bring performance costs.

ggndnn: I understood your meaning. Your suggestion reminded me. The previous code was not so correct, the prefix counter should not increase in newThread but in constructor, so there is no performance cost. I think no harm done to add a prefix counter to NamedThreadFactory, just like java.util.concurrent.DefautThreadFactory's pool counter, not only for test, it's helpful in debugging thread related issue. What do you think?

xingfudeshi: make sense.

common/src/main/java/io/seata/common/thread/NamedThreadFactory.java
```diff
@@ -24,9 +26,10 @@
  * The type Named thread factory.
  *
  * @author jimin.jm @alibaba-inc.com
- * @date 2018 /9/12
+ * @author ggndnn
  */
 public class NamedThreadFactory implements ThreadFactory {
+    private final static Map<String, AtomicInteger> PREFIX_COUNTER = new ConcurrentHashMap<>();
```
jsbxyyx: whether need specified capacity?

ggndnn: I don' t think it's so necessary, the default value looks ok.

PR #1722, opened 2019-09-29
optimize: simplify to make codes more readable

tjnettech:
Ⅰ. Describe what this PR did
Simplify to make codes more readable.

Ⅱ. Does this pull request fix one issue?
No.

Ⅲ. Why don't you add test cases (unit test/integration test)?

Ⅳ. Describe how to verify it

Ⅴ. Special notes for reviews

### DIFF
--- a/codec/seata-codec-kryo/src/main/java/io/seata/codec/kryo/KryoCodec.java
+++ b/codec/seata-codec-kryo/src/main/java/io/seata/codec/kryo/KryoCodec.java
@@ -27,7 +27,7 @@ public class KryoCodec implements Codec {
 
     @Override
     public <T> byte[] encode(T t) {
-        if (t == null || !(t instanceof AbstractMessage)) {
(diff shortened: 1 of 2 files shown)


codec/seata-codec-kryo/src/main/java/io/seata/codec/kryo/KryoCodec.java
```diff
@@ -27,7 +27,7 @@
 
     @Override
     public <T> byte[] encode(T t) {
-        if (t == null || !(t instanceof AbstractMessage)) {
+        if (!(t instanceof AbstractMessage)) {
             throw new IllegalArgumentException("message is null");
```
xingfudeshi: The exception message should be changed as well.how about "message is illegal".

tjnettech: Yes, I didn't notice the message.

PR #1723, opened 2019-09-30
feature: add benchmark projects

ggndnn:
Performance is critical, this PR was opened to encourage and help contributors to do benchmark test and improve performance. Project `seata-benchmark-undolog` is an example of microbenchmark.


leizhiyuan:
maybe it should be in another project？like the  sample project？

lovepoem:
> maybe it should be in another project？like the sample project？

Agree with you 
Benchmark tests are usually used as performance tests to get benchmark data. Modifications are not frequent, and it's better to be a separate project.

ggndnn:
> > maybe it should be in another project？like the sample project？
> 
> Agree with you
> Benchmark tests are usually used as performance tests to get benchmark data. Modifications are not frequent, and it's better to be a separate project.

I thought about it too, a separate benchmark project. How to add one?

slievrly:
> > > maybe it should be in another project？like the sample project？
> > 
> > 
> > Agree with you
> > Benchmark tests are usually used as performance tests to get benchmark data. Modifications are not frequent, and it's better to be a separate project.
> 
> I thought about it too, a separate benchmark project. How to add one?

https://github.com/seata/seata-benchmark

ggndnn:
Close this PR here and reopen another one [https://github.com/seata/seata-benchmark/pull/1](https://github.com/seata/seata-benchmark/pull/1) in seata-benchmark.

PR #1750, opened 2019-10-11
optimize: abstract table meta cache

github-ygy:
abstract table meta cache


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/AbstractTableMetaCache.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * get scheme from datasource and tableName
+     *
+     * @param dataSource
+     * @param tableName
+     * @return
+     * @throws SQLException
+     */
+    protected abstract TableMeta fetchScheme(DataSource dataSource, String tableName) throws SQLException;
```
slievrly: Scheme?

github-ygy: done ，modified to schema

github-ygy:
> whether need add oracle meta cache test case



> whether need add oracle meta cache test case

I don't know the logic yet. I found it needs to be sorted out from the bug of pr1749. I'll learn later that this logic can provide Oracle tests, but I can provide TableMetaCacheFactoryTest.

jsbxyyx:
> > whether need add oracle meta cache test case
> 
> > whether need add oracle meta cache test case
> 
> I don't know the oracle yet. I found it needs to be sorted out from the bug of pr1749. I'll learn later that can provide Oracle tests, but now I can provide TableMetaCacheFactoryTest.

Sorry, I am not expressing it right.

PR #1761, opened 2019-10-14
bugfix: oracle rollback failed when the table has null Blob Clob value

l81893521:
fix oracle rollback failed when the table that include null value in the Clob and Blob column

#1762


slievrly:
@l81893521 add issue.

l81893521:
> @l81893521 add issue.

ok

Issue #1767, opened 2019-10-14
Get Table meta failed when set useInformationSchema true and table name was keyword


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
Can not get table meta failed when set useInformationSchema true and table name was keyword

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
io.seata.common.exception.ShouldNeverHappenException: [xid:192.168.117.1:8091:2024785002]get tablemeta failed
	at io.seata.rm.datasource.sql.struct.cache.AbstractTableMetaCache.getTableMeta(AbstractTableMetaCache.java:78)
	at io.seata.rm.datasource.AbstractConnectionProxy.prepareStatement(AbstractConnectionProxy.java:113)
	at org.springframework.jdbc.core.JdbcTemplate$SimplePreparedStatementCreator.createPreparedStatement(JdbcTemplate.java:1510)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:602)
	at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:849)
	at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:904)
	at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:914)
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)
1.create a table that name was keyword like
```
CREATE TABLE `table` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT COMMENT 'primary key',
  `name` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=15 DEFAULT CHARSET=utf8mb4;
```
2. and set the useInformationSchema true like
```
jdbc:mysql://127.0.0.1:3306/xxx?useInformationSchema=true
```

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

Issue #1765, opened 2019-10-14
(springcloud+mybatis) Table Name Keyword error (Not repaired) version:0.8.1


warApePoint:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
If the database table name uses a keyword such as order，Mybatis automatically symbolizes the table name when SQL is executed(`order`)，The resulting TableMeta is null

### Ⅱ. Describe what happened

![1571048449(1)](https://user-images.githubusercontent.com/32215349/66745897-debc9100-eeb2-11e9-89b8-7c8d376c4889.jpg)
```
### Error updating database.  Cause: java.sql.SQLException: io.seata.common.exception.ShouldNeverHappenException: [xid:10.20.29.45:8091:2024776247]get tablemeta failed
### The error may exist in com/warape/seatorder/mapper/OrderMapper.java (best guess)
### The error may involve com.warape.seatorder.mapper.OrderMapper.insert-Inline
### The error occurred while setting parameters
### SQL: INSERT INTO `order`  ( user_id, product_id, count, money )  VALUES  ( ?, ?, ?, ? )
### Cause: java.sql.SQLException: io.seata.common.exception.ShouldNeverHappenException: [xid:10.20.29.45:8091:2024776247]get tablemeta failed
; uncategorized SQLException; SQL state [null]; error code [0]; io.seata.common.exception.ShouldNeverHappenException: [xid:10.20.29.45:8091:2024776247]get tablemeta failed; nested exception is java.sql.SQLException: io.seata.common.exception.ShouldNeverHappenException: [xid:10.20.29.45:8091:2024776247]get tablemeta failed] with root cause

io.seata.common.exception.ShouldNeverHappenException: [xid:10.20.29.45:8091:2024776247]get tablemeta failed
	at io.seata.rm.datasource.sql.struct.TableMetaCache.getTableMeta(TableMetaCache.java:91) ~[seata-all-0.8.1.jar:0.8.1]
	at io.seata.rm.datasource.exec.BaseTransactionalExecutor.getTableMeta(BaseTransactionalExecutor.java:194) ~[seata-all-0.8.1.jar:0.8.1]
	at io.seata.rm.datasource.exec.BaseTransactionalExecutor.getTableMeta(BaseTransactionalExecutor.java:178) ~[seata-all-0.8.1.jar:0.8.1]
[...]

PR #1768, opened 2019-10-14
bugfix: fixed problem when set useInformationSchema true and table name was keyword

l81893521:
fix can not get table meta when set useInformationSchema true and table name was keyword

#1765 #1767


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
```diff
@@ -106,6 +106,18 @@ private TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseMetaData
...
         tm.setTableName(tableName);
 
+        //when set the useInformationSchema true just like jdbc:mysql://127.0.0.1:3306/xxx?useInformationSchema=true
+        //mysql will use DatabaseMetaDataUsingInfoSchema instead of DatabaseMetaData
+        //so
+        //the type of get table meta will change from
+        //show full columns from xxx from xxx
+        //to
+        //select xxx from xxx where catalog_name like ? and table_name like ?
+        //in the second type we have to remove the "`"
```
xingfudeshi: Multiline comment please use /***/

l81893521: ok

PR #1770, opened 2019-10-15
optimize: string splice and release lock

zjinlei:
replace + to append
reduce unnecessary method:doReleaseLock


slievrly:
@zjinlei Is there a better description of the pr title?

PR #1788, opened 2019-10-18
bugfix: mysql can not get primary key value

### LINKED ISSUE #1736: insert sql can not getGeneratedKeys

### Ⅰ. Issue Description
insert sql can not rs.getGeneratedKeys()

because genKeys.getGeneratedKeys() operation in InsertExecutor, cause the rs cursor move to last.

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

jsbxyyx:
fix insert sql not get primary key, **only mysql**.

fixes #1736


rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
```diff
@@ -306,6 +306,11 @@ private boolean checkPkValues(List<Object> pkValues) {
             Object v = genKeys.getObject(1);
             pkValues.add(v);
         }
+        try {
+            genKeys.beforeFirst();
+        } catch (SQLException e) {
+            LOGGER.warn("Fail to reset ResultSet cursor. can not get primary key value");
```
xingfudeshi: This execption should be ignore?

jsbxyyx: only user not need get primary key, can be ignore exception.

PR #1796, opened 2019-10-21
fix #1794 unexcepted exception can roll back

### LINKED ISSUE #1794: rollbackFor does not work properly

### Ⅱ. Describe what happened

```
```

方法设置：    @GlobalTransactional(name = "fsp-create-order", rollbackFor = RuntimeException.class)

异常抛出：Exception

结果：应该不回滚，但是回滚了，我觉得rollbackOn方法有问题

1. xxx
2. xxx
3. xxx

public boolean rollbackOn(Throwable ex) {

        RollbackRule winner = null;
        int deepest = Integer.MAX_VALUE;

        if (this.rollbackRules != null) {
            for (RollbackRule rule : this.rollbackRules) {
                int depth = rule.getDepth(ex);
                if (depth >= 0 && depth < deepest) {
                    deepest = depth;
                    winner = rule;
                }
            }
        }

        return winner == null || !(winner instanceof NoRollbackRule);
    }

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

github-ygy:
fix #1794 unexcepted exception can roll back

TransactionInfoTest#testRollBackOn


tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
```diff
@@ -62,6 +62,7 @@ public boolean rollbackOn(Throwable ex) {
         int deepest = Integer.MAX_VALUE;
 
         if (this.rollbackRules != null) {
+            winner = NoRollbackRule.DEFAUL_NO_ROLLBACK_RULE;
```
jsbxyyx: the rollbackRules is not null but it is empty, not do roll back?

tm/src/main/java/io/seata/tm/api/transaction/NoRollbackRule.java
```diff
@@ -22,6 +22,8 @@
  */
 public class NoRollbackRule extends RollbackRule {
 
+    public static final NoRollbackRule DEFAUL_NO_ROLLBACK_RULE = new NoRollbackRule(Throwable.class);
```
xingfudeshi: DEFAUL->DEFAULT

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
xingfudeshi: > the rollbackRules is not null but it is empty, not do roll back?

When might it be empty?

jsbxyyx: image

xingfudeshi: OK.Then this.rollbackRules != null should be replaced with CollectionUtils.isNotEmpty.

tm/src/main/java/io/seata/tm/api/transaction/NoRollbackRule.java
github-ygy: done

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
github-ygy: done ,thanks for @jsbxyyx

PR #1798, opened 2019-10-22
optimize: improving method format performance

yangfuhai:

improving method format performance.

optimize code..


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionalInterceptor.java
```diff
@@ -160,9 +158,17 @@ public TransactionInfo getTransactionInfo() {
...
+
+        Class<?>[] params = method.getParameterTypes();
+        int in = 0;
+        for (Class<?> clazz : params) {
+            sb.append(clazz.getName());
+            if (++in < params.length) {
+                sb.append(", ");
+            }
+        }
+        return sb.append(")").toString();
```
slievrly: @yangfuhai Is the performance difference from the usage of + and append?

yangfuhai: the performance is 10 times higher than before. this is particularly important in high concurrency.

slievrly: Collectors.joining is implemented via StringBuilder.append, so the difference comes from method.getName() + paramTypes and StringBuilder(method.getName()).append?

yangfuhai: the reason for the performance gap maybe is from Arrays.stream.map vs for-loop。

zjinlei: recommended to use StringJoiner

PR #1799, opened 2019-10-22
optimize: reduce unnecessary synchronized

jsbxyyx:
optimize CodecFactory and LockerFactory get synchronized


server/src/main/java/io/seata/server/lock/LockerFactory.java
```diff
@@ -77,7 +77,7 @@ public static synchronized final LockManager getLockManager() {
      * @param branchSession the branch session
      * @return the lock manager
      */
-    public static synchronized final Locker get(BranchSession branchSession) {
+    public static final Locker get(BranchSession branchSession) {
```
zjinlei: see the method getLockManager

PR #1807, opened 2019-10-23
refactor: optimized NettyClientChannelManager and fixed potential memory leaks in it

ggndnn:
* Clarified the responsibilities of channel pool and client manager, the pool should be only responsible for channel management.
* Fixed potential memory leaks in `NettyClientChannelManager`, objects in channelLocks or poolKeyMap had no chance to be removed.

By the way,  considering some situations, e.g. the channels can be shared by threads, cached by `NettyClientChannelManager` most of the time, their life cycles are mainly controlled by netty, I think the pool acts like an object factory not a pool. So the pool's ability is not fully reflected, that could be optimized in future.


core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
```diff
@@ -76,8 +85,17 @@
      *
      * @return channels
      */
-    ConcurrentMap<String, Channel> getChannels() {
-        return channels;
+    Map<String, Channel> getChannels() {
+        return new HashMap<>(channels);
```
slievrly: how about Collections.unmodifiableMap？

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
```diff
@@ -175,38 +172,95 @@ void reconnect(String transactionServiceGroup) {
...
+        NettyPoolKey poolKey;
         try {
             NettyPoolKey currentPoolKey = poolKeyFunction.apply(serverAddress);
             NettyPoolKey previousPoolKey = poolKeyMap.putIfAbsent(serverAddress, currentPoolKey);
             if (null != previousPoolKey && previousPoolKey.getMessage() instanceof RegisterRMRequest) {
                 RegisterRMRequest registerRMRequest = (RegisterRMRequest) currentPoolKey.getMessage();
                 ((RegisterRMRequest) previousPoolKey.getMessage()).setResourceIds(registerRMRequest.getResourceIds());
             }
-            channelFromPool = nettyClientKeyPool.borrowObject(poolKeyMap.get(serverAddress));
+            poolKey = previousPoolKey != null ? previousPoolKey : currentPoolKey;
```
slievrly: previousPoolKey already judgment in if condtion.

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
```diff
@@ -175,38 +172,95 @@ void reconnect(String transactionServiceGroup) {
...
+            channelFromPool = nettyClientKeyPool.borrowObject(poolKey);
             channels.put(serverAddress, channelFromPool);
```
slievrly: when channel put in map, it will available but may be not identify.

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
ggndnn: previousPoolKey could still be null.

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
ggndnn: done.

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
ggndnn: fixed.

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
slievrly: it's my fault.

funky-eyes:

PR #1825, opened 2019-10-27
bugfix: global status inconsistent when rollback and branch register …

### LINKED ISSUE #1812: OrderServer rollback,but StorageServer no rollback
seata version:0.9.0

premise
The seata-server is deployed on windows, using the [springcloud-eureka-feign-mybatis-seata demo](https://github.com/seata/seata-samples/tree/master/springcloud-eureka-feign-mybatis-seata), and the Thread.sleep (2000) simulation request timeout is added to the StorageServer.decrease. Using jmeter, open 10 threads, each thread calls 2 times OrderServer order/create interface.

phenomenon
Since the timeout order table is all rolled back, no data is generated (normal)
Defect inventory appears in the storage table (exception)
Branch_table, lock_table, undo_log have a row of data, indicating that the transaction did not roll back normally (exception)
The global_table table is empty.
Can be reproduced multiple times.

zjinlei:
…are concurrent on db mode

fixes #1812


ymwangzq:
Dose doGlobalCommit have the same problem?

PR #1858, opened 2019-10-31
bugfix:fix Global transaction does not work

### LINKED ISSUE #1857: Global Transaction does not work

### Ⅰ. Issue Description
The globaltransactionscanner overrides the postprocessafterinitialization, so the postprocessafterinitialization of abstractautoproxycreator is not called, resulting in wrapifnecessary not being called, so all transactions are invalid.

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

xingfudeshi:

fixes #1857 


slievrly:
image
check result is ok.

Issue #1865, opened 2019-11-01
Lack of error log when branch register fail.


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
Could not found any error log in the TC when branch register fail.

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

PR #1866, opened 2019-11-01
optimize: server lack of error trace

l81893521:
log the error when branch register fail.

#1865


server/src/main/java/io/seata/server/coordinator/DefaultCore.java
```diff
@@ -88,6 +88,7 @@ public Long branchRegister(BranchType branchType, String resourceId, String clie
                 globalSession.addBranch(branchSession);
             } catch (RuntimeException ex) {
                 branchSession.unlock();
+                LOGGER.error("Failed to store branch xid = {} branchId = {}", globalSession.getXid(), branchSession.getBranchId(), ex);
```
slievrly: Why don't you throw the exception details？

l81893521: fixed.

PR #1869, opened 2019-11-02
performance: add phase one successful reporting switch

zjinlei:
optimize performance.
The successful reporting of the phase one has no practical use, but it does require an rpc and 2 db queries and 1 db write.


server/src/main/resources/file.conf
```diff
@@ -47,6 +47,7 @@ client {
     retry.times = 30
   }
   report.retry.count = 5
+  phase_one_true.report = true
```
slievrly: report.enable?

zjinlei: done

PR #1884, opened 2019-11-09
optimize: auto closeable

jsbxyyx:
optimize: auto closeable


core/src/main/java/io/seata/core/store/db/LockStoreDataBaseDAO.java
```diff
@@ -185,27 +186,7 @@ public boolean acquireLock(List<LockDO> lockDOs) {
...
-            if (ps != null) {
-                try {
-                    ps.close();
-                } catch (SQLException e) {
-                }
-            }
-            if (conn != null) {
-                try {
-                    if (originalAutoCommit) {
-                        conn.setAutoCommit(true);
```
zjinlei: need to be processed originalAutoCommit

common/src/main/java/io/seata/common/util/IOUtil.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author jsbxyyx
+ */
+public class IOUtil {
+
+    /**
+     * close Closeable
+     * @param closeables the closeables
+     */
+    public static void close(AutoCloseable... closeables) {
+        if (closeables != null && closeables.length > 0) {
```
zjinlei: use seata CollectionUtils

echooymxq:
why not use try-with-resources?

helloworlde:
I found all resource in this change could use try-with-resource, and that can be closed automatic. Maybe change to use try-with-resource is better.

jsbxyyx:
@echooymxq @helloworlde Not all places can be used try-with-resources.
image

slievrly:
@jsbxyyx database operation related variable names can be optimized.

jsbxyyx:
@slievrly OK.

helloworlde:
> @echooymxq @helloworlde Not all places can be used try-with-resources.
> image

It's can be closed by multiple try.

jsbxyyx:
> > @echooymxq @helloworlde Not all places can be used try-with-resources.
> > image
> 
> It's can be closed by multiple try.

If use try-with-resources, the `conn` close can not has other logic.

common/src/main/java/io/seata/common/util/IOUtil.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @author jsbxyyx
+ */
+public class IOUtil {
+
+    /**
+     * close Closeable
+     * @param closeables the closeables
+     */
+    public static void close(AutoCloseable... closeables) {
```
xingfudeshi: We can merge these two methods into one.
public static void close(AutoCloseable closeable,AutoCloseable... moreCloseables);

PR #1894, opened 2019-11-13
not use

ruqinhu:


ruqinhu:
clientTest is not use,so to delete

PR #1898, opened 2019-11-13
feature:  support dynamic config

slievrly:

feature:  support dynamic config


config/seata-config-etcd3/src/main/java/io/seata/config/etcd3/EtcdConfiguration.java
```diff
@@ -250,32 +238,57 @@ private static Client getClient() {
...
+                public void onNext(WatchResponse watchResponse) {
+                    try {
+                        GetResponse getResponse = getClient().getKVClient().get(ByteSequence.from(dataId, UTF_8)).get();
+                        List<KeyValue> keyValues = getResponse.getKvs();
+                        if (CollectionUtils.isNotEmpty(keyValues)) {
+                            event.setDataId(dataId).setNewValue(keyValues.get(0).getValue().toString(UTF_8));
+                            listener.onChangeEvent(event);
+                        }
+                    } catch (Exception e) {
+                        LOGGER.error("error occurred while getting value{}", e.getMessage());
```
xingfudeshi: LOGGER.error("error occurred while getting value{}", e.getMessage(),e);

config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
```diff
@@ -229,79 +213,49 @@ private void setFailResult(ConfigFuture configFuture) {
...
+    class FileListener implements ConfigurationChangeListener {
 
-        /**
-         * Instantiates a new Config change runnable.
-         */
-        public ConfigChangeRunnable() {
-        }
+        private final String dataId;
+        private final ConfigurationChangeListener listener;
+        private final ExecutorService executor = new ThreadPoolExecutor(CORE_LISTENER_THREAD, MAX_LISTENER_THREAD, 0L,
```
zjinlei: Generate too many instances of executor and Thread.

config/seata-config-etcd3/src/main/java/io/seata/config/etcd3/EtcdConfiguration.java
slievrly: fixed.

config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
slievrly: This file subscription does not have a subscription function itself. It requires a while to monopolize a thread. For each subscription, a thread pool with a single thread is required.

config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
```diff
@@ -197,6 +183,20 @@ public void run() {
...
+                            Config appConfig = ConfigFactory.parseFileAnySyntax(new File(targetFilePath));
+                            tempConfig = ConfigFactory.load(appConfig);
+                        } else {
+                            tempConfig = ConfigFactory.load(name);
+                        }
+                        if (null != tempConfig) {
+                            fileConfig = tempConfig;
+                            targetFileLastModified = tempLastModified;
+                        }
+                    }
```
zjinlei: FileConfiguration  test well.
But we use FileConfiguration to load registry.conf and file.conf, so both listen to their changes, Is it necessary to monitor them all?

slievrly: registry.conf is that the metadata configuration is not allowed to change at run time. If it is allowed to change, for example, if I use the Nacos registry, I need to reload the entire instance.

slievrly: fixed.

config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
```diff
@@ -197,6 +195,20 @@ public void run() {
                     return;
                 }
                 try {
+                    long tempLastModified = new File(targetFilePath).lastModified();
+                    if (allowDynamicRefresh && tempLastModified > targetFileLastModified) {
```
zjinlei: if(allowDynamicRefresh){
long tempLastModified = new File(targetFilePath).lastModified();
...
}

slievrly: fixed.

PR #1906, opened 2019-11-17
bugfix: add exception system exit code  when rpcServer init.

objcoding:
…cServer init exception.

We need to tell the system something about code(eg:-1) when tc server abnormal exit system


server/src/main/java/io/seata/server/Server.java
```diff
@@ -84,7 +84,12 @@ public static void main(String[] args) throws IOException {
         }
         XID.setPort(rpcServer.getListenPort());
 
-        rpcServer.init();
+        try {
+            rpcServer.init();
+        } catch (Throwable e) {
+            e.printStackTrace();
```
l81893521: Why do not use LOGGER.error instead of e.printStackTrace();

PR #1911, opened 2019-11-19
optimize : javadoc should be used for classes, class variables and methods.

lzf971107:

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionContext.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionContext.java
@@ -33,7 +33,10 @@ public class ConnectionContext {
     private String xid;
     private Long branchId;
     private boolean isGlobalLockRequire;
-    //table and primary key should not be duplicated
+
+    /**
+     * table and primary key should not be duplicated
+     */
     private Set<String> lockKeysBuffer = new HashSet<>();
     private List<SQLUndoLog> sqlUndoItemsBuffer = new ArrayList<>();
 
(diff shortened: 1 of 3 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionContext.java
```diff
@@ -33,7 +33,10 @@
     private String xid;
     private Long branchId;
     private boolean isGlobalLockRequire;
-    //table and primary key should not be duplicated
+
+    /**
+     * table and primary key should not be duplicated
```
helloworlde: Please use capital for the first letter, and use dot as end.

server/src/main/java/io/seata/server/store/StoreConfig.java
```diff
@@ -29,13 +29,19 @@
     private static final Configuration CONFIGURATION = ConfigurationFactory.getInstance();
 
 
-    // default 16kb
+    /**
+     * default 16kb
+     */
```
helloworlde: Please use capital for the first letter, and use dot as end. And description should be more clearly.

server/src/main/java/io/seata/server/store/StoreConfig.java
```diff
@@ -29,13 +29,19 @@
...
     private static final int DEFAULT_MAX_BRANCH_SESSION_SIZE = 1024 * 16;
 
-    // default 512b
+    /**
+     * default 512b
```
helloworlde: Same as above.

server/src/main/java/io/seata/server/store/StoreConfig.java
```diff
@@ -29,13 +29,19 @@
...
+     */
     private static final int DEFAULT_MAX_GLOBAL_SESSION_SIZE = 512;
 
-    // default 16kb
+    /**
+     * default 16kb
```
helloworlde: Same as above.

PR #1914, opened 2019-11-19
bugfix: support where condition exists for Oracle 

yuhuangbin:
support where condition exists for Oracle


rm-datasource/src/main/java/io/seata/rm/datasource/sql/druid/BaseRecognizer.java
```diff
@@ -50,6 +57,20 @@ public BaseRecognizer(String originalSQL) {
 
     }
 
+    public void excuteVisit(SQLExpr where, SQLASTVisitor visitor) {
```
l81893521: Please change the name to executeVisit()

yuhuangbin: fixed

Issue #1916, opened 2019-11-19
NullPointerException in DB mock when mock value was null 


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
When set the mock value null, would cause some NullPointerException during the CI.

### Ⅱ. Describe what happened
```
07:49:53.429 [main] ERROR io.seata.rm.datasource.sql.struct.cache.AbstractTableMetaCache - get table meta error:Failed to fetch schema of t1
java.sql.SQLException: Failed to fetch schema of t1
	at io.seata.rm.datasource.sql.struct.cache.MysqlTableMetaCache.fetchSchema(MysqlTableMetaCache.java:86)
	at io.seata.rm.datasource.sql.struct.cache.AbstractTableMetaCache.refresh(AbstractTableMetaCache.java:90)
	at io.seata.rm.datasource.sql.struct.TableMetaCacheTest.refreshTest_0(TableMetaCacheTest.java:146)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:628)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:117)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:184)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:180)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:127)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:68)
[...]

PR #1917, opened 2019-11-19
bugfix: NullPointerException in DB mock during CI

l81893521:
When set the mock value null, would cause some NullPointerException during the CI.

#1916


rm-datasource/src/test/java/io/seata/rm/datasource/mock/MockDriver.java
```diff
@@ -110,6 +110,9 @@ public MockPreparedStatement createSeataMockPreparedStatement(MockConnection con
      * @param mockReturnValue
      */
     public void setMockReturnValue(Object[][] mockReturnValue) {
+        if (mockReturnValue == null) {
+            this.mockReturnValue = new Object[][]{};
```
helloworlde: Need return here, change `if ... else ...` or `... ? ... : ...` could be better

rm-datasource/src/test/java/io/seata/rm/datasource/mock/MockDriver.java
```diff
@@ -134,6 +141,9 @@ public void setMockColumnsMetasReturnValue(Object[][] mockColumnsMetasReturnValu
      * @param mockIndexMetasReturnValue
      */
     public void setMockIndexMetasReturnValue(Object[][] mockIndexMetasReturnValue) {
+        if (mockIndexMetasReturnValue == null) {
+            this.mockIndexMetasReturnValue = new Object[][]{};
```
helloworlde: Same as above.

rm-datasource/src/test/java/io/seata/rm/datasource/mock/MockDriver.java
l81893521: Fixed. Thanks

rm-datasource/src/test/java/io/seata/rm/datasource/mock/MockDriver.java
l81893521: Fixed. Thanks

PR #1927, opened 2019-11-22
bugfix:A class with private access constructors should not be loaded by SPI.

CharmingRabbit:
Fix a bug that will appear when the class has default private constructor


common/src/main/java/io/seata/common/loader/EnhancedServiceLoader.java
```diff
@@ -352,8 +352,10 @@ private static void loadFile(Class<?> service, String dir, ClassLoader classLoad
             constructor.setAccessible(true);
             s = service.cast(constructor.newInstance(args));
         } else {
-            // default Constructor
-            s = service.cast(implClazz.newInstance());
+            // default Constructor (might be private, use ctor.newInstance() instead of class.newInstance())
+            Constructor<S> constructor = implClazz.getDeclaredConstructor();
+            constructor.setAccessible(true);
+            s = service.cast(constructor.newInstance());
```
xingfudeshi: Should a class with private constructor be loaded with SPI ?

CharmingRabbit: It is true that should not. The scenario I just met is that my service provider class is a singleton class (with a private constructor and a getInstance method). I should not use singleton mode in service provider class. Moreover, I  think this line of code 
`constructor.setAccessible(true);`
should not break the encapsulation of a class either when an argument constructor is called.

xingfudeshi: Yes.I agree with you.

CharmingRabbit:
> LGTM.

Thanks for your approval
I want to know Is there any  other conditions to pass this pr（Ask another person with authority to review?） Thank you for your help.

CharmingRabbit:
> LGTM.

@xingfudeshi 
Thanks for your approval
I want to know Is there any other conditions to pass this pr（Ask another person with authority to review?） Thank you for your help.

CharmingRabbit:
@jsbxyyx @xingfudeshi 
all the checks and reviews passed,please merge into develop.

PR #1931, opened 2019-11-22
feature: nacos-config.py support  namespace

anrror:


slievrly:
image

function test pass.

slievrly:
./nacos-config.py localhost $nacos-namespace-id

PR #1936, opened 2019-11-24
optimize: Not every service needs to initialize tm client.

objcoding:
Not every service needs to initialize tm client, and need to initialize tm client when service use GlobalTransactional annotation.


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -304,6 +313,24 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+            try {
+                Class<?> serviceInterface = SpringProxyUtils.findTargetClass(bean);
+                if (existsAnnotation(new Class[]{serviceInterface})) {
+                    initTmClient = true;
+                    if (LOGGER.isDebugEnabled()) {
+                        LOGGER.debug("Need to init tm client.");
+                    }
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
xingfudeshi: Should any exceptions stop the starting of application?

objcoding: > Should any exceptions stop the starting of application?

findTargetClass method

xingfudeshi: > > Should any exceptions stop the starting of application?
> 
> findTargetClass method

I think any exception here should be ignored. We can display a log to the user and tell them that TM client did not start for some reasons. However, we can not directly prevent the application from starting.What do you think?

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -165,20 +168,29 @@ public void destroy() {
...
-                    + txServiceGroup + "]");
+
+        // Not every service needs to initialize tm client.
+        // Need to initialize tm client when use @GlobalTransactional.
+        if (initTmClient) {
+            //init TM
+            TMClient.init(applicationId, txServiceGroup);
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info(
+                    "Transaction Manager Client is initialized. applicationId[" + applicationId + "] txServiceGroup["
```
xingfudeshi: use {}

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -165,20 +168,29 @@ public void destroy() {
...
         }
-        //init TM
-        TMClient.init(applicationId, txServiceGroup);
-        if (LOGGER.isInfoEnabled()) {
-            LOGGER.info(
-                "Transaction Manager Client is initialized. applicationId[" + applicationId + "] txServiceGroup["
-                    + txServiceGroup + "]");
+
+        // Not every service needs to initialize tm client.
+        // Need to initialize tm client when use @GlobalTransactional.
```
xingfudeshi: and @GlobalLock.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -165,20 +168,29 @@ public void destroy() {
         ShutdownHook.getInstance().destroyAll();
     }
 
-    private void initClient() {
+    public void setInitTmClient(boolean initTmClient) {
+        this.initTmClient = initTmClient;
+    }
+
+    public void initClient() {
```
zjinlei: Exposure to this method is not very good

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > > > Should any exceptions stop the starting of application?
> > 
> > 
> > findTargetClass method
> 
> I think any exception here should be ignored. We can display a log to the user and tell them that TM client did not start for some reasons. However, we can not directly prevent the application from starting.What do you think?

I think ok，
My previous consideration was that if TM did not init, but GlobalTransactional was useful in the service.
I wil fix it, thank you brother.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > use {}

ok

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
```diff
@@ -79,11 +80,12 @@ public void testWrapLocalTccBean(Object bean, String beanName, Object cacheKey)
...
     /**
-     * Test after properties set.
+     * Test initClient.
      */
     @Test
-    public void testAfterPropertiesSet() {
-        globalTransactionScanner.afterPropertiesSet();
+    public void testInitClient() {
+        globalTransactionScanner.setInitTmClient(true);
+        globalTransactionScanner.initClient();
```
xingfudeshi: Can assert something?

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: ok,thanks for your reviewed.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -287,15 +299,16 @@ private MethodDesc makeMethodDesc(GlobalTransactional anno, Method method) {
     }
 
     @Override
-    public void afterPropertiesSet() {
-        if (disableGlobalTransaction) {
-            if (LOGGER.isInfoEnabled()) {
-                LOGGER.info("Global transaction is disabled.");
+    public void onApplicationEvent(ContextRefreshedEvent event) {
+        if (event.getApplicationContext().getParent() == null) {
```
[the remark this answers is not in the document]
xingfudeshi: > Can this judgment be determined to execute the initClient only once?

If the parent is null,that means this context is the root context.So,only once will execute.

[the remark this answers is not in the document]
objcoding: > Can this judgment be determined to execute the initClient only once?

if spring boot application, There will be two  application context, so initClinet method wil be executed twice.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > Exposure to this method is not very good

test那里要用到，暂时还没有什么好方法

xingfudeshi: > Exposure to this method is not very good

Seems for testing purpose.It can be replaced by setAccessible (true);

objcoding: > Seems for testing purpose.It can be replaced by setAccessible (true);

ok

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
zjinlei: > and @GlobalLock.
@xingfudeshi  Why?

xingfudeshi: Sorry.My mistake.Only @GlobalTransactional need to initialize tm client.

xingfudeshi: Then you can't just use existsAnnotation Method,you have to create new one check for @GlobalTransactional only.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > > Exposure to this method is not very good
> 
> Seems for testing purpose.It can be replaced by setAccessible (true);

please review again.

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
objcoding: > Can assert something?

please review again.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > > > > Should any exceptions stop the starting of application?
> > > 
> > > 
> > > findTargetClass method
> > 
> > 
> > I think any exception here should be ignored. We can display a log to the user and tell them that TM client did not start for some reasons. However, we can not directly prevent the application from starting.What do you think?
> 
> I think ok，
> My previous consideration was that if TM did not init, but GlobalTransactional was useful in the service.
> I wil fix it, thank you brother.

please review again.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > > use {}
> 
> ok

please review again.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > ok,thanks for your reviewed.

please review again.

objcoding: > Sorry.My mistake.Only @GlobalTransactional need to initialize tm client.

thank you, please review again.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -276,6 +286,24 @@ private boolean existsAnnotation(Class<?>[] classes) {
         return false;
     }
 
+    private boolean exitsGlobalTransaction(Class<?>[] classes) {
```
zjinlei: exits->exists

zjinlei:
@long187 Will this pr affect saga?

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > exits->exists

ok

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
```diff
@@ -79,11 +81,21 @@ public void testWrapLocalTccBean(Object bean, String beanName, Object cacheKey)
...
+        try {
+            Class<? extends GlobalTransactionScanner> aClass = globalTransactionScanner.getClass();
+            Method setInitTmClientMethod = aClass.getDeclaredMethod("setInitTmClient", Boolean.TYPE);
+            setInitTmClientMethod.setAccessible(true);
+            Method initClientMethod = aClass.getDeclaredMethod("initClient");
+            initClientMethod.setAccessible(true);
+            setInitTmClientMethod.invoke(globalTransactionScanner, true);
+            initClientMethod.invoke(globalTransactionScanner);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
objcoding: > Just throws explicit exception on the method signature.

ok

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
```diff
@@ -79,11 +81,21 @@ public void testWrapLocalTccBean(Object bean, String beanName, Object cacheKey)
...
+     * Test initClient.
      */
     @Test
-    public void testAfterPropertiesSet() {
-        globalTransactionScanner.afterPropertiesSet();
+    public void testInitClient() {
+        try {
+            Class<? extends GlobalTransactionScanner> aClass = globalTransactionScanner.getClass();
+            Method setInitTmClientMethod = aClass.getDeclaredMethod("setInitTmClient", Boolean.TYPE);
+            setInitTmClientMethod.setAccessible(true);
```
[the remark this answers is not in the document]
objcoding: > setInitTmClient is already a public modifier.

ok

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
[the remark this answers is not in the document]
objcoding: > method comments

ok

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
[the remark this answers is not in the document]
zjinlei: @objcoding not see any changes

spring/src/test/java/io/seata/spring/annotation/GlobalTransactionScannerTest.java
zjinlei: @objcoding not see any changes

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -287,15 +318,16 @@ private MethodDesc makeMethodDesc(GlobalTransactional anno, Method method) {
     }
 
     @Override
-    public void afterPropertiesSet() {
-        if (disableGlobalTransaction) {
-            if (LOGGER.isInfoEnabled()) {
-                LOGGER.info("Global transaction is disabled.");
+    public void onApplicationEvent(ContextRefreshedEvent event) {
+        if (event.getApplicationContext().getParent() == null) {
```
slievrly: test on springboot application is ok？

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -173,19 +177,23 @@ private void initClient() {
...
+                    applicationId, txServiceGroup);
+            }
         }
         //init RM
         RMClient.init(applicationId, txServiceGroup);
         if (LOGGER.isInfoEnabled()) {
             LOGGER.info(
-                "Resource Manager is initialized. applicationId[" + applicationId + "] txServiceGroup[" + txServiceGroup
-                    + "]");
+                "Resource Manager is initialized..applicationId[{}],txServiceGroup[{}]", applicationId, txServiceGroup);
```
l81893521: Please remove one of the period.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -233,9 +241,8 @@ protected Object wrapIfNecessary(Object bean, String beanName, Object cacheKey)
                     }
                 }
 
-                LOGGER.info(
-                    "Bean[" + bean.getClass().getName() + "] with name [" + beanName + "] would use interceptor ["
-                        + interceptor.getClass().getName() + "]");
+                LOGGER.info("Bean[{}] with name [{}] would use interceptor [{}]",
```
l81893521: I think leak of space after the Bean.

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -287,15 +318,16 @@ private MethodDesc makeMethodDesc(GlobalTransactional anno, Method method) {
...
+        if (event.getApplicationContext().getParent() == null) {
+            if (disableGlobalTransaction) {
+                if (LOGGER.isInfoEnabled()) {
+                    LOGGER.info("Global transaction is disabled.");
+                }
+                return;
             }
-            return;
+            initClient();
         }
```
l81893521: How do you think if write like this
```
if (event.getApplicationContext().getParent() == null && disableGlobalTransaction) {
        if (LOGGER.isInfoEnabled()) {
                LOGGER.info("Global transaction is disabled.");
        }
} else {
        initClient();
}

objcoding: > How do you think if write like this
> 
> ```
> if (event.getApplicationContext().getParent() == null && disableGlobalTransaction) {
>         if (LOGGER.isInfoEnabled()) {
>                 LOGGER.info("Global transaction is disabled.");
>         }
> } else {
>         initClient();
> }
> ```

two if is easy to understanding

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
objcoding: > test on springboot application is ok？

after testing, is ok.

slievrly:
Temporarily not merged, waiting for feedback

slievrly:
Here are some points to consider: whether the provider directly initiates a globalRollback request when an exception occurs in the provider. If you want to support this feature, you need to enable the transactionManager on the provider.

PR #1946, opened 2019-11-26
optimize: remove variable deepest and reduce duplicate judgment

ruqinhu:
1.(null instanceof Object) always be false
2. Depth has been judged by RollbackRule.getDepth(ex)

### DIFF
--- a/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
+++ b/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
@@ -59,21 +59,16 @@ public final class TransactionInfo implements Serializable {
     }
 
     public boolean rollbackOn(Throwable ex) {
-
         RollbackRule winner = null;
-        int deepest = Integer.MAX_VALUE;
-
         if (CollectionUtils.isNotEmpty(rollbackRules)) {
             winner = NoRollbackRule.DEFAULT_NO_ROLLBACK_RULE;
             for (RollbackRule rule : this.rollbackRules) {
                 int depth = rule.getDepth(ex);
-                if (depth >= 0 && depth < deepest) {
-                    deepest = depth;
+                if (depth >= 0) {
                     winner = rule;
                 }
             }
         }
-
-        return winner == null || !(winner instanceof NoRollbackRule);
+        return !(winner instanceof NoRollbackRule);
     }
 }
--- a/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
+++ b/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
@@ -66,7 +66,6 @@ public final class TransactionInfo implements Serializable {
                 int depth = rule.getDepth(ex);
                 if (depth >= 0) {
                     winner = rule;
-                    break;
                 }
             }
         }
--- a/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
+++ b/tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
@@ -66,6 +66,7 @@ public final class TransactionInfo implements Serializable {
                 int depth = rule.getDepth(ex);
                 if (depth >= 0) {
                     winner = rule;
+                    break;
                 }
             }
         }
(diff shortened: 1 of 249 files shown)


tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
```diff
@@ -59,21 +59,16 @@ public void setRollbackRules(Set<RollbackRule> rollbackRules) {
...
         RollbackRule winner = null;
-        int deepest = Integer.MAX_VALUE;
-
         if (CollectionUtils.isNotEmpty(rollbackRules)) {
             winner = NoRollbackRule.DEFAULT_NO_ROLLBACK_RULE;
             for (RollbackRule rule : this.rollbackRules) {
                 int depth = rule.getDepth(ex);
-                if (depth >= 0 && depth < deepest) {
-                    deepest = depth;
+                if (depth >= 0) {
```
slievrly: rollbackRules contains AException and BException,  AException extend BException.  please valid this scenario.

ruqinhu: add class MyIllegalArgumentException extends IllegalArgumentException,and test is ok

ruqinhu: > rollbackRules contains AException and BException, AException extend BException. please valid this scenario.

add class MyIllegalArgumentException extends IllegalArgumentException,and test is ok

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
```diff
@@ -59,21 +59,16 @@ public void setRollbackRules(Set<RollbackRule> rollbackRules) {
...
                     winner = rule;
```
ruqinhu: rollbackRules is LinkedHashSet .  in this ,if i add break ,RollbackRule will be a Higher priority. 
but now, it just like as it was : Loop will execute to end ,and if RollbackRule and NoRollBackRule has same Exception, noRollBackRule will will take effect.

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
ruqinhu: > rollbackRules contains AException and BException, AException extend BException. please valid this scenario.
add MyIllegalArgumentException extends IllegalArgumentException ,test is ok.
and I have a question below.

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
ruqinhu: same Exception refers to the relationship of inheritance

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
slievrly: It is always the smallest win. You have not compared their priority here. What is the point of the loop just assigning?

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
slievrly: `if RollbackRule and NoRollBackRule has same Exception` Regardless of this scenario, users need to be responsible for their own actions. If you say both yes and no about a thing, it is impossible for the code to determine whether it is or not.

tm/src/main/java/io/seata/tm/api/transaction/TransactionInfo.java
ruqinhu: yes, return value is boolean ,Why compare priorities？if roolback and noroolback has same priority, i think below can add break for reducing cycle times...

ruqinhu: sorry, just thought if roolbackrule and noroolnack has father and son exception, smallest win will useful

ruqinhu:
sorry, just thought if roolbackrule and noroolnack has father and son exception, smallest win will useful

ruqinhu:
update test for showing smallest winner feature

Issue #1952, opened 2019-11-27
seata-0.9.0.1 表名识别异常


azbh111:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
在sql表名前加上数据库名字，会抛异常： java.sql.SQLException: Failed to fetch schema of saas_yibao.`test_model`

seata在查询表的索引时，执行的sql是 SHOW INDEX FROM `saas_yibao.test_model` FROM `saas_yibao`

然而正确的sql应该是 SHOW INDEX FROM `test_model` FROM `saas_yibao`

```sql
insert into saas_yibao.`test_model` (`name`) values (#{name});
```

### Ⅱ. Describe what happened

```
java.sql.SQLException: Failed to fetch schema of saas_yibao.`test_model`
	at io.seata.rm.datasource.sql.struct.cache.MysqlTableMetaCache.fetchSchema(MysqlTableMetaCache.java:86)
	at io.seata.rm.datasource.sql.struct.cache.AbstractTableMetaCache.lambda$getTableMeta$0(AbstractTableMetaCache.java:62)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2039)
	at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1853)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2037)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2020)
	at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:112)
	at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:54)
	at io.seata.rm.datasource.sql.struct.cache.AbstractTableMetaCache.getTableMeta(AbstractTableMetaCache.java:60)
	at io.seata.rm.datasource.exec.BaseTransactionalExecutor.getTableMeta(BaseTransactionalExecutor.java:191)
	at io.seata.rm.datasource.exec.BaseTransactionalExecutor.getTableMeta(BaseTransactionalExecutor.java:176)
	at io.seata.rm.datasource.exec.InsertExecutor.beforeImage(InsertExecutor.java:72)
[...]

PR #1953, opened 2019-11-28
bugfix: fix get table meta failed with catalog

l81893521:
Can not get table meta when the SQL include catalog, like
```
select * from account.account_tbl
```

#1952


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
```diff
@@ -102,6 +102,7 @@ protected TableMeta fetchSchema(DataSource dataSource, String tableName) throws
 
     private TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseMetaData dbmd, String tableName)
         throws SQLException {
+        //always "" for mysql
```
zjinlei: What does this sentence mean?

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
```diff
@@ -122,6 +123,9 @@ private TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseMetaData
             tableName = tableName.replace("`", "");
         }
 
+        String[] catalogTable = tableName.split("\\.");
```
slievrly: why not use rsmd.getTableName()?

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
l81893521: That means this method always return "" in Mysql. Mysql does not implement this interface.

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
l81893521: Thanks a lot.

PR #1956, opened 2019-11-28
optimize: modify no available server error more clearly, and fixed NP…

helloworlde:
1. Print expect server service name instant of `no available server to connect.` when could found server in registry center
2. Fixed issue of NPE when get server service instances return empty.

1. Modify server service group name not `default`.
2. Start a service with using registry center.
3. When service can't find `default` service, will print log like `no available service 'default' found, please make sure registry config correct`


core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
```diff
@@ -209,15 +212,15 @@ private Channel doConnect(String serverAddress) {
...
-        List<InetSocketAddress> availInetSocketAddressList = RegistryFactory.getInstance().lookup(
-            transactionServiceGroup);
-        if (!CollectionUtils.isEmpty(availInetSocketAddressList)) {
-            for (InetSocketAddress address : availInetSocketAddressList) {
-                availList.add(NetUtil.toStringAddress(address));
-            }
+        List<InetSocketAddress> availInetSocketAddressList = RegistryFactory.getInstance()
+                                                                            .lookup(transactionServiceGroup);
+        if (CollectionUtils.isEmpty(availInetSocketAddressList)) {
+            return null;
```
xingfudeshi: please return a emtpy list to avoid NPE.

helloworlde: Had modified, the caller method has judgement for null or empty too.

discovery/seata-discovery-core/src/main/java/io/seata/discovery/registry/RegistryService.java
```diff
@@ -88,4 +88,12 @@
...
      */
     void close() throws Exception;
+
+    /**
+     * Get current service group name
+     *
+     * @param key service group
+     * @return the service group name
+     */
+    String getServiceGroup(String key);
```
slievrly: I think you should define a default method.

helloworlde: Had replaced.

PR #1957, opened 2019-11-29
refactor:load keywordChecker through SPI

CharmingRabbit:
(1) Make the constructor of KeywordChecker(msql,oracle) public so that they can be loaded by SPI(EnhancedServiceLoader)
(2) Add keywordChecker resource file
(3) KeywordCheckerFactory no longer recognize specific db type,when more database types are supported in the future, its code does not need to be changed


CharmingRabbit:
> LGTM

thanks for your approval ,could you help to notify other reviewer to review this pr?

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
```diff
@@ -28,20 +29,31 @@
...
+                    Map<String, KeywordChecker> initializedMap = new HashMap<>();
+                    List<KeywordChecker> checkerList = EnhancedServiceLoader.loadAll(KeywordChecker.class);
+                    for (KeywordChecker checker : checkerList) {
+                        initializedMap.put(checker.getDbType().toLowerCase(), checker);
+                    }
+                    keywordCheckerMap = initializedMap;
+                }
+            }
+        }
+        if (keywordCheckerMap.containsKey(dbType)) {
```
slievrly: dbType. toLowerCase()?

rm-datasource/src/main/java/io/seata/rm/datasource/undo/oracle/keyword/OracleKeywordChecker.java
```diff
@@ -28,29 +29,12 @@
  * @date 2019/3/25 oracle keyword checker
  */
 public class OracleKeywordChecker implements KeywordChecker {
-    private static volatile KeywordChecker keywordChecker = null;
-    private Set<String> keywordSet;
+    private static Set<String> keywordSet;
 
-    private OracleKeywordChecker() {
+    static {
```
jsbxyyx: why is constructor update to static

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
```diff
@@ -28,20 +29,31 @@
...
-            return MySQLKeywordChecker.getInstance();
-        } else  if (dbType.equals(JdbcConstants.ORACLE)) {
-            return OracleKeywordChecker.getInstance();
-        } else {
-            throw new NotSupportYetException(dbType);
+        if (keywordCheckerMap == null) {
+            synchronized (KeywordCheckerFactory.class) {
+                if (keywordCheckerMap == null) {
+                    Map<String, KeywordChecker> initializedMap = new HashMap<>();
+                    List<KeywordChecker> checkerList = EnhancedServiceLoader.loadAll(KeywordChecker.class);
```
xingfudeshi: Why load all types of keyword checker?I think only one type of keyword is required.
You can use @LoadLevel annotation to load a explict implementation of keywordChecker.
image

xingfudeshi:
@CharmingRabbit Is it possible to add any test case to veriy this feature?

rm-datasource/src/main/java/io/seata/rm/datasource/undo/oracle/keyword/OracleKeywordChecker.java
CharmingRabbit: In old way. the keywordSet will be initialized when the private ctor executed and will be initialized only once(by the lock). 
Now the constructor is set to public, so the initialization of the keywordset is put into the static block to ensure that it is executed only once.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
CharmingRabbit: This is because I noticed that all database types in jdbcconstants are lowercase, and what is not sure is whether the provider of keywordchecker will report its own database types in the correct case, but it seems a little strange to do so(call toLowerCase()).
In fact, when the specific keywordchecker provides the wrong dbtype（just like POSTGRES not postgres）, to throw an exception in Factory is also a reasonable way, so I removed tower case () finally

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
zjinlei: @xingfudeshi support for multiple data sources

CharmingRabbit: @xingfudeshi 
thanks for your advise,it is true that only one checker is used once.

but when a not supported dbType is loaded second or more times,using loadAll(), we can throw unsupported exceptions directly. while using load(), we need to do the following:
(1) obtain the lock(this will also block other thread to abtain the checker of other dbType)
(2) reload all extensions of KeywordChecker,
(3) use reflect to obtain annotation and judge whether the dbtype is supported
(4) throw exception. 

Moreover,getting a explict checker at a time will not bring any performance improvement. reviewing the source code of EnhancedServiceLoader, I found that the difference between loadall() and load() is whether to get comments through reflection and filter by activeName (they both need to load all the extensions)  so I think it will be better to use loadAll()

CharmingRabbit: @xingfudeshi 
thanks for your advise,it is true that only one checker is used once.

but when a not supported dbType is loaded second or more times,using loadAll(), we can throw unsupported exceptions directly. while using load(), we need to do the following:
(1) obtain the lock(this will also block other thread to abtain the checker of other dbType)
(2) reload all extensions of KeywordChecker,
(3) use reflect to obtain annotation and judge whether the dbtype is supported
(4) throw exception.

Moreover,getting a explict checker at a time will not bring any performance improvement. reviewing the source code of EnhancedServiceLoader, I found that the difference between loadall() and load() is whether to get annotation through reflection and filter by activeName (they both need to load all the extensions) so I think it will be better to use loadAll()

CharmingRabbit:
> @CharmingRabbit Is it possible to add any test case to veriy this feature?

@xingfudeshi 
This refactoring only changes the internal logic of the Factory and the creation logic of the specific keywordChecker. The usage scenario does not change. 
Therefore, the existed test cases (KeywordCheckerFactorTest, OracleKeywordCheckerTest, MySQLKeywordCheckerTest) can completely cover the code after refactoring.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
xingfudeshi: OK.Can we use concurrentHashMap.computeIfAbsent to instead of Double check?
And it seems our SPI loader should be optimized.

zjinlei: https://www.yuque.com/cgbcac/aoni5y/uw0n5d
The 50th mission: spi optimize

CharmingRabbit: @xingfudeshi 
I am afraid not now,computeIfAbsent is suit for the situation that a single element was load by using double-checked locking,but now we load all the elements at once,the return value of these two situation mismatchs.
we might use it to simplify the code after optimizing the SPI loader.

zjinlei: @CharmingRabbit Can you recieve the task 50 to optimize spi?
Not limited to 50 descriptions.

CharmingRabbit: @zjinlei 
Yes, I'd like to. Is there a deadline for this task?

zjinlei: @CharmingRabbit please add my dingding i8vx5rv

rm-datasource/src/main/java/io/seata/rm/datasource/undo/KeywordCheckerFactory.java
slievrly: BaseTransactionalExecutor#getDbType，it depends on the jdbc interface to return. Is this guaranteed to return lowercase? Here I suggest lowercase or change the parameters to known enum types.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/oracle/keyword/OracleKeywordChecker.java
```diff
@@ -522,4 +506,10 @@ public String checkAndReplace(String fieldOrTableName) {
         return check(fieldOrTableName) ? fieldOrTableName : fieldOrTableName;
         //        return check(fieldOrTableName)?"`" + fieldOrTableName + "`":fieldOrTableName;
     }
+
+    @Override
+    public String getDbType()
+    {
```
jsbxyyx: code style

rm-datasource/src/main/java/io/seata/rm/datasource/undo/mysql/keyword/MySQLKeywordChecker.java
```diff
@@ -1135,4 +1119,10 @@ public String checkAndReplace(String fieldOrTableName) {
         return check(fieldOrTableName) ? "`" + fieldOrTableName + "`" : fieldOrTableName;
     }
 
+    @Override
+    public String getDbType()
+    {
```
jsbxyyx: code style

PR #1980, opened 2019-12-03
optimize:show the applicationId when register TM

lightClouds917:
optimize:show the applicationId when register TM
when register TM,we do not know which application register,we just know the ip and port like this:
`io.seata.core.rpc.DefaultServerMessageListenerImpl.onRegTmMessage:133 -checkAuth for client:192.168.xx.xx:10984 vgroup:fsp_tx_group ok
`
When we monitor the server log，it is very unfriendly,we want to know which application had regiested,or remove，like this:
`io.seata.core.rpc.DefaultServerMessageListenerImpl.onRegTmMessage:133 -checkAuth for client:192.168.xx.xx:10984 vgroup:fsp_tx_group  applicationId:user-server ok`

### DIFF
--- a/core/src/main/java/io/seata/core/rpc/DefaultServerMessageListenerImpl.java
+++ b/core/src/main/java/io/seata/core/rpc/DefaultServerMessageListenerImpl.java
@@ -133,8 +133,8 @@ public class DefaultServerMessageListenerImpl implements ServerMessageListener {
                 Version.putChannelVersion(ctx.channel(), message.getVersion());
                 isSuccess = true;
                 if (LOGGER.isInfoEnabled()) {
-                    LOGGER.info(String.format("checkAuth for client:%s vgroup:%s ok", ipAndPort,
-                        message.getTransactionServiceGroup()));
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/seata/core/rpc/DefaultServerMessageListenerImpl.java
```diff
@@ -133,8 +133,8 @@ public void onRegTmMessage(RpcMessage request, ChannelHandlerContext ctx, Server
                 Version.putChannelVersion(ctx.channel(), message.getVersion());
                 isSuccess = true;
                 if (LOGGER.isInfoEnabled()) {
-                    LOGGER.info(String.format("checkAuth for client:%s vgroup:%s ok", ipAndPort,
-                        message.getTransactionServiceGroup()));
+                    LOGGER.info("checkAuth for client:{},vgroup:{},applicationId:{}",
```
slievrly: Here you need a status to indicate that it is authenticated `OK`.

PR #1994, opened 2019-12-05
optimize: rename zk configuration root path.

objcoding:
There are many projects for 'config' node in zk, such as kafka.

### DIFF
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigurationKeys.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigurationKeys.java
@@ -29,7 +29,11 @@ public final class ConfigurationKeys {
     /**
      * The constant FILE_ROOT_CONFIG.
      */
-    public static final String FILE_ROOT_CONFIG = "seata";
+    public static final String FILE_ROOT_CONFIG = "config";
+    /**
+     * The constant SEATA_FILE_ROOT_CONFIG
+     */
+    public static final String SEATA_FILE_ROOT_CONFIG = "seata";
     /**
      * The constant FILE_CONFIG_SPLIT_CHAR.
      */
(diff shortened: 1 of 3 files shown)


config/seata-config-core/src/main/java/io/seata/config/ConfigurationKeys.java
```diff
@@ -29,7 +29,7 @@
     /**
      * The constant FILE_ROOT_CONFIG.
      */
-    public static final String FILE_ROOT_CONFIG = "config";
+    public static final String FILE_ROOT_CONFIG = "seata";
```
xingfudeshi: There are some places need to be changed too.

objcoding: > There are some places need to be changed too.

Has been modified.

objcoding:
> There are some places need to be changed too.

Has been modified

config/seata-config-zk/src/main/java/io/seata/config/zk/ZookeeperConfiguration.java
```diff
@@ -47,7 +48,7 @@
 
     private static final String CONFIG_TYPE = "zk";
     private static final String ZK_PATH_SPLIT_CHAR = "/";
-    private static final String ROOT_PATH = ZK_PATH_SPLIT_CHAR + FILE_ROOT_CONFIG;
+    private static final String ROOT_PATH = ZK_PATH_SPLIT_CHAR + SEATA_FILE_ROOT_CONFIG;
```
zjinlei: please check other Configuration Centers and assess if modifications are needed,such as etcd3.

objcoding: > please check other Configuration Centers and assess if modifications are needed,such as etcd3.

only zk

PR #1996, opened 2019-12-06
bugfix:fix auto proxying of datasource which has final modifier

### LINKED ISSUE #1995: problem about auto proxying of datasource which has final modifier

### Ⅰ. Issue Description
When enable auto proxying of datasouce which has final modifier,it will cause a Exception means the class can't be proxied by using cglib.

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

I will fix it.

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

xingfudeshi:
When enable auto proxying of datasouce which has final modifier,it will cause a Exception means the class can't be proxied by using cglib.This PR will use JDK proxy instead of cglib to avoid this problem.

fixes #1995 

I've done integration testing with following scenes:
1.Specific init-method for a datasource bean defination.
2.A datasource class that with final modifier.
3.A datasource that extends AbstractRoutingDataSource.

### DIFF
--- a/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
+++ b/spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
@@ -42,11 +42,9 @@ import org.springframework.aop.TargetSource;
 import org.springframework.aop.framework.AdvisedSupport;
 import org.springframework.aop.framework.autoproxy.AbstractAutoProxyCreator;
 import org.springframework.aop.support.AopUtils;
-import org.springframework.beans.BeanUtils;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.cglib.proxy.Enhancer;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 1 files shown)


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -311,21 +312,25 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
...
-                    return m.invoke(dataSourceProxy, args);
-                } else {
-                    boolean oldAccessible = method.isAccessible();
-                    try {
-                        method.setAccessible(true);
-                        return method.invoke(bean, args);
-                    } finally {
-                        //recover the original accessible for security reason
-                        method.setAccessible(oldAccessible);
+            return Proxy.newProxyInstance(Thread.currentThread().getContextClassLoader(), bean.getClass().getInterfaces(), new InvocationHandler() {
```
zjinlei: If the user configures DataSourceProxy and auto-proxy is enabled, DataSourceProxy's DataSource will be proxied.
It is recommended to check beforehand, prompt the user and terminate.

xingfudeshi: fixed

spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -304,28 +306,41 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
         this.setBeanFactory(applicationContext);
     }
 
+    @Override
+    public Object postProcessBeforeInitialization(Object bean, String beanName) {
+        if (bean instanceof DataSourceProxy && ConfigurationFactory.getInstance().getBoolean(DATASOURCE_AUTOPROXY, false)) {
```
jsbxyyx: I think you judge datasourceproxy targetDataSource is DataSourceProxy，If true, notice user.

jsbxyyx: Deal with later

PR #2001, opened 2019-12-07
bugfix: replace deprecated jvm args

helloworlde:

1. replace deprecated jvm args

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -183,7 +183,7 @@
                             <jvmFlag>-server</jvmFlag>
                             <jvmFlag>-Xss512k</jvmFlag>
                             <jvmFlag>-XX:+UnlockExperimentalVMOptions</jvmFlag>
-                            <jvmFlag>-XX:+UseCGroupMemoryLimitForHeap</jvmFlag>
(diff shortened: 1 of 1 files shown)


server/pom.xml
```diff
@@ -183,7 +183,7 @@
                             <jvmFlag>-server</jvmFlag>
                             <jvmFlag>-Xss512k</jvmFlag>
                             <jvmFlag>-XX:+UnlockExperimentalVMOptions</jvmFlag>
-                            <jvmFlag>-XX:+UseCGroupMemoryLimitForHeap</jvmFlag>
+                            <jvmFlag>-XX:+UseContainerSupport</jvmFlag>
```
slievrly: since  jdk11+

helloworlde: Actually, the arg is added in Java 8u191 also, you can reference [
JDK 8u191 Update Release Notes](https://www.oracle.com/technetwork/java/javase/8u191-relnotes-5032181.html), our image is based on 8u212, so it can execute correctly.  And I had a try of image, it works well.

image
image

Issue #2058, opened 2019-12-18
Schedule retry rollback cause repeated rollback


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
schedule retry rollback cause repeated rollback

### Ⅱ. Describe what happened

```
2019-12-18 18:20:37.456 INFO [ServerHandlerThread_1_500]io.seata.server.coordinator.DefaultCore.begin:154 -Successfully begin global transaction xid = 192.168.117.1:8091:2030392429
2019-12-18 18:20:37.849 WARN [configOperate_2_2]io.seata.config.FileConfiguration.run:265 -Could not found property store.file.max-branch-session-size, try to use default value instead.
2019-12-18 18:20:37.852 WARN [configOperate_2_2]io.seata.config.FileConfiguration.run:265 -Could not found property store.db.min-conn, try to use default value instead.
2019-12-18 18:20:37.853 WARN [configOperate_2_2]io.seata.config.FileConfiguration.run:265 -Could not found property store.db.max-conn, try to use default value instead.
2019-12-18 18:20:37.853 WARN [configOperate_2_2]io.seata.config.FileConfiguration.run:265 -Could not found property store.db.min-conn, try to use default value instead.
2019-12-18 18:20:37.853 WARN [configOperate_2_2]io.seata.config.FileConfiguration.run:265 -Could not found property store.db.min-conn, try to use default value instead.
2019-12-18 18:20:37.861 INFO [ServerHandlerThread_1_500]io.seata.common.loader.EnhancedServiceLoader.loadFile:237 -load LockStore[DB] extension by class[io.seata.core.store.db.LockStoreDataBaseDAO]
2019-12-18 18:20:37.861 INFO [ServerHandlerThread_1_500]io.seata.common.loader.EnhancedServiceLoader.loadFile:237 -load Locker[db] extension by class[io.seata.server.lock.db.DataBaseLocker]
2019-12-18 18:20:38.045 INFO [ServerHandlerThread_1_500]io.seata.server.coordinator.DefaultCore.lambda$branchRegister$0:98 -Successfully register branch xid = 192.168.117.1:8091:2030392429, branchId = 2030392431
[...]

PR #2059, opened 2019-12-18
bugfix: fix repeated rollback caused by asynchronous rollback thread

l81893521:
schedule retry rollback thread pool cause repeated rollback

#2058


server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -381,6 +381,10 @@ protected void handleRetryRollbacking() {
         long now = System.currentTimeMillis();
         for (GlobalSession rollbackingSession : rollbackingSessions) {
             try {
+                //prevent repeated rollback
+                if (rollbackingSession.getStatus().equals(GlobalStatus.Rollbacking) && !rollbackingSession.isTimeout()) {
```
slievrly: if throw RunTimeException it will no longer rollback.

server/src/main/java/io/seata/server/session/GlobalSession.java
```diff
@@ -122,6 +122,14 @@ public boolean isTimeout() {
         return (System.currentTimeMillis() - beginTime) > timeout;
     }
 
+    /**
+     * prevent could not handle rollbacking transaction
+     * @return if true force roll back
+     */
+    public boolean isRollbackingDead() {
+        return (System.currentTimeMillis() - beginTime) > (3 * 6000);
```
zjinlei: (System.currentTimeMillis() - beginTime) > (3 * 6000 + timeout);
Suggested overload isTimeout()

zjinlei: My mistake, you are right!

server/src/main/java/io/seata/server/coordinator/DefaultCoordinator.java
l81893521: I think it will keep rollback

PR #2082, opened 2019-12-24
change version to 1.1.0-SNAPSHOT and add logs

slievrly:
change version to 1.1.0-SNAPSHOT and add logs


lovepoem:
Should be 1.0.1-snapshot

PR #2086, opened 2019-12-24
feature: add apollo configuration namespace.

### LINKED ISSUE #1555: seata config with apoll is not easy to uses enough, can you optimize it?
1. In actual usage scenarios, many (even that all) applications or serivice machines use the same apollo settings (apollo settings file: _/opt/settings/server.properties_), this settings file contains '_apollo.meta_' and '_env_' elements, can seata load this  settings file instead of config '_apollo.meta_' property in every application?
2. Likewise, in actual usage scenarios, many TM/RM applications use the same seata settings (such as _transport_, _service_, _client_ etc...), can seata support the **namespace** function of apollo ? If support, seata settings are configurated in a public namespace, many appilcations share these settings, instead of all applications duplicate them. In this case 'app.id' config item is optional.

objcoding:

fixes #1555


script/client/spring/application.properties
```diff
@@ -58,6 +58,7 @@ seata.config.consul.server-addr=127.0.0.1:8500
 
 seata.config.apollo.apollo-meta=http://192.168.1.204:8801
 seata.config.apollo.app-id=seata-server
+seata.config.apollo.namespace=application
```
slievrly: please review  `ConfigApolloProperties`, maybe you want to add an attribute?

objcoding:
> Is ConfigApolloProperties missing the namespace field?

resolved.

PR #2103, opened 2019-12-28
optimize: tcc module code style

lzf971107:


tcc/src/main/java/io/seata/rm/tcc/interceptor/ActionContextUtil.java
```diff
@@ -96,9 +98,7 @@ public static void getAllField(Class<?> interFace, List<Field> fields) {
             return;
         }
         Field[] field = interFace.getDeclaredFields();
-        if (field != null) {
-            fields.addAll(Arrays.asList(field));
-        }
+        fields.addAll(Arrays.asList(field));
```
zjinlei: Why remove field != null?

tcc/src/main/java/io/seata/rm/tcc/api/BusinessActionContext.java
```diff
@@ -145,11 +145,6 @@ public void setActionName(String actionName) {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder();
-        sb.append("[xid:").append(xid)
-            .append(",branch_Id:").append(branchId).append(",action_name:").append(actionName)
-            .append(",action_context:")
-            .append(actionContext).append("]");
-        return sb.toString();
+        return "[xid:" + xid + ",branch_Id:" + branchId + ",action_name:" + actionName + ",action_context:" + actionContext + "]";
```
zjinlei: ?

tcc/src/main/java/io/seata/rm/tcc/interceptor/ActionContextUtil.java
lzf971107: Because interFace.getDeclaredFields() can never empty, if there is no field in the interFace default returned for an empty array, instead of null.

tcc/src/main/java/io/seata/rm/tcc/api/BusinessActionContext.java
lzf971107: I think the use of StringBuilder is generally in a loop. In fact, simple string splicing will be optimized by JVM to StringBuilder. Personally, the efficiency is not much different, and the code is more concise. [Stackoverflow](https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java) has related questions and answers on this question.

tcc/src/main/java/io/seata/rm/tcc/interceptor/ActionContextUtil.java
```diff
@@ -15,25 +15,28 @@
...
+import io.seata.common.exception.FrameworkException;
+import io.seata.common.util.StringUtils;
+import io.seata.rm.tcc.api.BusinessActionContextParameter;
+
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
```
a364176773: 尽量把java的包放在最上边

lzf971107: resolved

tcc/src/main/java/io/seata/rm/tcc/api/BusinessActionContext.java
slievrly: @lzf971107 will there be more memory fragmentation here?

lzf971107: > @lzf971107 will there be more memory fragmentation here?

They have the same code after decompilation, and the effect they cause is actually the same. If memory fragmentation occurs, then StringBuilder is also the same.

PR #2113, opened 2019-12-30
bugfix: when branchRollback failed, it will trigger retry of multi-tc.

objcoding:
https://github.com/seata/seata/issues/2089


zjinlei:
Not a solution to issue 2089

PR #2118, opened 2019-12-31
Update OracleTableMetaCache.java

dadatou20089:
schema info in sql statement may be lower,

Oracle sql的语法支持schema为小写，但是通过getColumns的schemaPattern传参必须为大写，否则无法获取表的字段信息 。

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
@@ -100,7 +100,7 @@ public class OracleTableMetaCache extends AbstractTableMetaCache {
         TableMeta tm = new TableMeta();
         tm.setTableName(tableName);
         String[] schemaTable = tableName.split("\\.");
-        String schemaName = schemaTable.length > 1 ? schemaTable[0] : dbmd.getUserName();
+        String schemaName = schemaTable.length > 1 ? schemaTable[0].toUpperCase() : dbmd.getUserName();
         tableName = schemaTable.length > 1 ? schemaTable[1] : tableName;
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
```diff
@@ -100,7 +100,7 @@ private TableMeta resultSetMetaToSchema(DatabaseMetaData dbmd, String tableName)
         TableMeta tm = new TableMeta();
         tm.setTableName(tableName);
         String[] schemaTable = tableName.split("\\.");
-        String schemaName = schemaTable.length > 1 ? schemaTable[0] : dbmd.getUserName();
+        String schemaName = schemaTable.length > 1 ? schemaTable[0].toUpperCase() : dbmd.getUserName();
```
slievrly: If the table is created case sensitively, for example "test" converted to uppercase is wrong.

l81893521: You are right, the username of Oracle also case sensitive, but I think you can not toUpperCase directly(check out from here:https://docs.oracle.com/en/database/oracle/oracle-database/19/dbseg/managing-security-for-oracle-database-users.html#GUID-4723A344-402C-423C-A864-9D7975A02B49). 
You should consider this situation below
```
select * from "TEST".ACCOUNT_TBL;
select * from TEST.ACCOUNT_TBL;
select * from "Test".ACCOUNT_TBL;
```

l81893521:
Waiting for your feedback for a long time. The bug already fixed, please check #2283.

l81893521:
The bug already fix. So I will close this pr, waiting for your contribution next time.

PR #2140, opened 2020-01-06
optimize: GzipUtil code style

lzf971107:


compressor/seata-compressor-gzip/src/main/java/io/seata/compressor/gzip/GzipUtil.java
```diff
@@ -28,48 +26,43 @@
...
             throw new NullPointerException("bytes is null");
         }
-        ByteArrayOutputStream out = null;
-        GZIPOutputStream gzip = null;
-        try {
-            out = new ByteArrayOutputStream();
-            gzip = new GZIPOutputStream(out);
+
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream();
+             GZIPOutputStream gzip = new GZIPOutputStream(out)) {
```
xingfudeshi: The close method of GZIPOutputStream should be called.And the close method of ByteArrayOutputStream actually does nothing.

compressor/seata-compressor-gzip/src/main/java/io/seata/compressor/gzip/GzipUtil.java
```diff
@@ -28,48 +26,43 @@
...
             throw new NullPointerException("bytes is null");
         }
-        ByteArrayOutputStream out = null;
-        GZIPInputStream gunzip = null;
-        try {
-            out = new ByteArrayOutputStream();
-            gunzip = new GZIPInputStream(new ByteArrayInputStream(bytes));
+
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream();
+             GZIPInputStream gunzip = new GZIPInputStream(new ByteArrayInputStream(bytes))) {
```
xingfudeshi: The close method of GZIPInputStream should be called.

lzf971107:
@xingfudeshi 
Hello, the following code is the decompiled code. It can be clearly seen that he uses the close method of gzip.

```java
public static byte[] compress(byte[] bytes) {
        if (bytes == null) {
            throw new NullPointerException("bytes is null");
        } else {
            try {
                ByteArrayOutputStream out = new ByteArrayOutputStream();
                Throwable var2 = null;

                Object var5;
                try {
                    GZIPOutputStream gzip = new GZIPOutputStream(out);
                    Throwable var4 = null;

                    try {
                        gzip.write(bytes);
                        gzip.flush();
                        gzip.finish();
                        var5 = out.toByteArray();
                    } catch (Throwable var30) {
                        var5 = var30;
                        var4 = var30;
                        throw var30;
                    } finally {
                        if (gzip != null) {
                            if (var4 != null) {
                                try {
                                    gzip.close();
                                } catch (Throwable var29) {
                                    var4.addSuppressed(var29);
                                }
                            } else {
                                gzip.close();
                            }
                        }

                    }
                } catch (Throwable var32) {
                    var2 = var32;
                    throw var32;
                } finally {
                    if (out != null) {
                        if (var2 != null) {
                            try {
                                out.close();
                            } catch (Throwable var28) {
                                var2.addSuppressed(var28);
                            }
                        } else {
                            out.close();
                        }
                    }

                }

                return (byte[])var5;
            } catch (IOException var34) {
                throw new RuntimeException("gzip compress error", var34);
            }
        }
    }
```

lzf971107:
@xingfudeshi 
decompress method is the same as compress method.
```java
public static byte[] decompress(byte[] bytes) {
        if (bytes == null) {
            throw new NullPointerException("bytes is null");
        } else {
            try {
                ByteArrayOutputStream out = new ByteArrayOutputStream();
                Throwable var2 = null;

                try {
                    GZIPInputStream gunzip = new GZIPInputStream(new ByteArrayInputStream(bytes));
                    Throwable var4 = null;

                    try {
                        byte[] buffer = new byte[8192];

                        int n;
                        while((n = gunzip.read(buffer)) > -1) {
                            out.write(buffer, 0, n);
                        }

                        byte[] var7 = out.toByteArray();
                        return var7;
                    } catch (Throwable var32) {
                        var4 = var32;
                        throw var32;
                    } finally {
                        if (gunzip != null) {
                            if (var4 != null) {
                                try {
                                    gunzip.close();
                                } catch (Throwable var31) {
                                    var4.addSuppressed(var31);
                                }
                            } else {
                                gunzip.close();
                            }
                        }

                    }
                } catch (Throwable var34) {
                    var2 = var34;
                    throw var34;
                } finally {
                    if (out != null) {
                        if (var2 != null) {
                            try {
                                out.close();
                            } catch (Throwable var30) {
                                var2.addSuppressed(var30);
                            }
                        } else {
                            out.close();
                        }
                    }

                }
            } catch (IOException var36) {
                throw new RuntimeException("gzip decompress error", var36);
            }
        }
    }
```

xingfudeshi:
What i mean is that the use of try with resource statement is redundant for ByteArrayOutputStream or ByteArrayInputStream.Because the method close does not do anything.

lzf971107:
> What i mean is that the use of try with resource statement is redundant for ByteArrayOutputStream or ByteArrayInputStream.Because the method close does not do anything.

Oh, Im sorry, I will change my code.

compressor/seata-compressor-gzip/src/main/java/io/seata/compressor/gzip/GzipUtil.java
```diff
@@ -28,48 +26,42 @@
...
-        GZIPOutputStream gzip = null;
-        try {
-            out = new ByteArrayOutputStream();
-            gzip = new GZIPOutputStream(out);
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        try (GZIPOutputStream gzip = new GZIPOutputStream(out)) {
             gzip.write(bytes);
-            gzip.close();
+            gzip.flush();
+            gzip.finish();
```
slievrly: Is it necessary here? close will call the finish method.

lzf971107: > Is it necessary here? close will call the finish method.

Of course not. If you do not call the flush method and the finish method before the close method, you will not be able to write data into the stream. Since the return statement will return before the finally block, the data was not written into the stream at that time, which caused compression failure. Further decompression failed, and I wrote a new unit test method for compression and decompression testing.

PR #1703, opened 2019-09-24
refactor: create sql parser spi and a druid type sql parser

ggndnn:
### Ⅰ. Describe what this PR did
This PR extracted sql parser part from rm-datasource to create sql parser SPI (seata-sqlparser-core), all imports from druid in rm-datasource were removed except unit test code. Druid related code was moved to a new project seata-sqlparser-druid where a druid type sql parser was implemented.

PR #2156, opened 2020-01-08
refactor: add sql parser type druid as constant

ggndnn:

According to the last review comment in #1703, added sql parser type druid as constant.


zjinlei:
Could it be an enumeration class that put more sql parsers type?

ggndnn:
> Could it be an enumeration class that put more sql parser types?

How about an abstract class contains sql parser type constants? Because `@LoadLevel(name = "druid")` needs constants. Any suggestion?

zjinlei:
> > Could it be an enumeration class that put more sql parser types?
> 
> How about an abstract class contains sql parser type constants? Because `@LoadLevel(name = "druid")` needs constants. Any suggestion?

Enumeration attributes are constants, so why not consider it?
I think it is possible to build a SqlParserType class in seata-sqlparser-core.

ggndnn:
> > > Could it be an enumeration class that put more sql parser types?
> > 
> > 
> > How about an abstract class contains sql parser type constants? Because `@LoadLevel(name = "druid")` needs constants. Any suggestion?
> 
> Enumeration attributes are constants, so why not consider it?
> I think it is possible to build a SqlParserType class in seata-sqlparser-core.

`@LoadLevel(name=` needs `String` type constants. I'm not sure how to add enum to it, so my suggestion is to create an abstract class `SqlParserType` contains string type constants or just use `ConfigurationKeys`.

```
public @interface LoadLevel {
    /**
     * Name string.
     *
     * @return the string
     */
    String name();

    /**
     * Order int.
     *
     * @return the int
     */
    int order() default 0;
}
```

ggndnn:
> LGTM. We may need to move ConfigurationKeys to the common module.

This can be considered.

ggndnn:
> > LGTM. We may need to move ConfigurationKeys to the common module.
> 
> This can be considered.

@slievrly 
Need to do this in this pr or in future pr after more extensive discussion?

zjinlei:
> ConfigurationKeys

Sorry, My low-level mistake.
But it is only an constant, not an configuration key, and there will be more sql parser types, I don't think it should be put in ConfigurationKeys.

ggndnn:
> > ConfigurationKeys
> 
> Sorry, My low-level mistake.
> But it is only a constant, not a configuration key, and there will be more sql parser types, I don't think it should be put in ConfigurationKeys.
> Hope that there is a better way, can list more sql parser later.

@zjinlei 

SqlParserType added.

Issue #2165, opened 2020-01-10
Seata saga high concurrent asynchronous execution with duplicate primary key XID


wenminglei:
@long187  Seata saga high concurrent asynchronous execution with duplicate primary key XID。

PR #2167, opened 2020-01-10
bugfix: saga concurrent asynchronous execution with duplicate primary key xid. 

long187:
bugfix: saga concurrent asynchronous execution with duplicate primary key xid. 

#2165


saga/seata-saga-tm/src/main/java/io/seata/saga/tm/DefaultSagaTransactionalTemplate.java
```diff
@@ -84,6 +85,7 @@ public void rollbackTransaction(GlobalTransaction tx, Throwable ex)
 
     @Override
     public GlobalTransaction beginTransaction(TransactionInfo txInfo) throws TransactionalExecutor.ExecutionException {
+        RootContext.unbind();
```
zjinlei: Please clean up after thread execution

PR #2180, opened 2020-01-14
optimize: use Lambda to initialize ThreadLocal

CvShrimp:
optimize the usage of ThreadLocal with lambda

no


tm/src/main/java/io/seata/tm/api/transaction/TransactionHookManager.java
```diff
@@ -38,10 +38,6 @@ private TransactionHookManager() {
      */
     public static List<TransactionHook> getHooks() throws IllegalStateException {
         List<TransactionHook> hooks = LOCAL_HOOKS.get();
-
```
slievrly: emptyList size is 0, ArrayList default size is 10; emptyList will not put in the `LOCAL_HOOKS `, but ArrayList always put in the  `LOCAL_HOOKS `.

CvShrimp:
i have modify the commit

PR #2184, opened 2020-01-15
refactor: make seata-config more readable

koonchen:

I found some optimization points when i was reading source code in seata-config module.

### DIFF
--- a/config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
@@ -53,7 +53,7 @@ public class FileConfiguration extends AbstractConfiguration {
 
     private static final int MAX_CONFIG_OPERATE_THREAD = 2;
 
-    private static final long LISTENER_CONFIG_INTERNAL = 1000;
+    private static final long LISTENER_CONFIG_INTERNAL = 1 * 1000;
 
     private static final String REGISTRY_TYPE = "file";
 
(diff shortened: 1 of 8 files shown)


config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
```diff
@@ -53,7 +53,7 @@
 
     private static final int MAX_CONFIG_OPERATE_THREAD = 2;
 
-    private static final long LISTENER_CONFIG_INTERNAL = 1 * 1000;
+    private static final long LISTENER_CONFIG_INTERNAL = 1000;
```
zjinlei: 1 * 1000 is more readable

config/seata-config-etcd3/src/main/java/io/seata/config/etcd3/EtcdConfiguration.java
```diff
@@ -240,7 +238,7 @@ private static Client getClient() {
     /**
      * the type config change notifier
      */
-    private class EtcdListener implements ConfigurationChangeListener {
+    private static class EtcdListener implements ConfigurationChangeListener {
```
zjinlei: Why use static?

config/seata-config-core/src/main/java/io/seata/config/FileConfiguration.java
koonchen: ok

config/seata-config-etcd3/src/main/java/io/seata/config/etcd3/EtcdConfiguration.java
koonchen: EtcdListener/NacosListener/ZKListener does not have access to the enclosing instance. It uses less space.

PR #2185, opened 2020-01-15
bugfix: fixed issue of judgement container in kubernetes

helloworlde:
1. Fixed issue of judgement container error in older version kubernetes


server/src/main/java/io/seata/server/ParameterParser.java
```diff
@@ -132,7 +133,7 @@ private Boolean isRunningInContainer() {
         Path path = Paths.get(C_GROUP_PATH);
         if (Files.exists(path)) {
             try (Stream<String> stream = Files.lines(path)) {
-                return stream.anyMatch(line -> line.contains(DOCKER_PATH));
+                return stream.anyMatch(line -> line.contains(DOCKER_PATH) || line.contains(KUBEPODS_PATH));
```
zjinlei: Is it necessary to make a configuration?

helloworlde: I think doesn't need, the only different is the priority of configuration form environment.

Issue #1515, opened 2019-08-22
Is there Transactional Propagation like spring in seata?


kaixinbaba:
We'd like to know,  Is there any **Transactional Propagation** like 
```
public enum Propagation {
    REQUIRED(0),
    SUPPORTS(1),
    MANDATORY(2),
    REQUIRES_NEW(3),
    NOT_SUPPORTED(4),
    NEVER(5),
    NESTED(6);

    private final int value;

    private Propagation(int value) {
        this.value = value;
    }

    public int value() {
        return this.value;
    }
}
```
in seata?
Thank you~

PR #2206, opened 2020-01-19
feature: support to set propagation of global transaction

CharmingRabbit:
Supports 4 propagations of transaction:
(1) REQUIRED: Use existing transaction context or if not, begin a new transaction  (the default propagation)
(2) REQUIRES_NEW: Begin a new transaction context.
(3) NOT_SUPPORTED: exclude the existing transaction context.
(4) SUPPORTS: If there is a transaction, join the transaction; if there is no transaction, execute it in a non transactional way

fixed [#1515](https://github.com/seata/seata/issues/1515)


tm/src/main/java/io/seata/tm/api/Propagation.java
```diff
@@ -0,0 +1,64 @@
...
+    Propagation(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public static Propagation get(String name) {
+        for (Propagation propagation : Propagation.class.getEnumConstants()) {
```
zjinlei: Propagation.values() is it enough?

tm/src/main/java/io/seata/tm/api/Propagation.java
```diff
@@ -0,0 +1,64 @@
...
+ *  limitations under the License.
+ */
+package io.seata.tm.api;
+
+/**
+ * Propagation level of global transactions.
+ *
+ * @author haozhibei
+ */
+public enum Propagation {
```
zjinlei: Propagation is a property of the transaction, can be at the same directory as TransactionInfo.java.

tm/src/main/java/io/seata/tm/api/Propagation.java
CharmingRabbit: I've  simplified the propagation to a pure enum class

tm/src/main/java/io/seata/tm/api/Propagation.java
CharmingRabbit: ACK

PR #2288, opened 2020-02-23
codecov: codecov ignore mock

lovepoem:

Codecov ignore mock

ci success


lovepoem:
> https://github.com/seata/seata/pull/2288/files#diff-8bc9db6adba7208e25c96be322351063L14 needs to be modified because the codec module has been renamed as serializer.

image

The generated  codes is in target ,I think it  doesn't affect coverage

PR #2301, opened 2020-02-26
feature: postgresql add default expr and nextval

jsbxyyx:
feature: postgresql add default expr and convert nextval to sequence expr.


sqlparser/seata-sqlparser-druid/src/main/java/io/seata/sqlparser/druid/postgresql/PostgresqlInsertRecognizer.java
```diff
@@ -113,12 +116,21 @@ public boolean visit(SQLExprTableSource x) {
                 } else if (expr instanceof SQLVariantRefExpr) {
                     row.add(((SQLVariantRefExpr) expr).getName());
                 } else if (expr instanceof SQLMethodInvokeExpr) {
```
jaspercloud, line 118: InsertExecutor.getPkValuesBySequence逻辑没有实现

sqlparser/seata-sqlparser-core/src/main/java/io/seata/sqlparser/struct/SqlDefaultExpr.java
```diff
@@ -0,0 +1,42 @@
+/*
```
l81893521, line 1: why not use com.alibaba.druid.sql.ast.expr.SQLSequenceExpr

jsbxyyx: sqlparser module can be multi implemention. druid just one of them.

rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
```diff
@@ -172,31 +173,65 @@ protected boolean containsColumns() {
...
+     */
+    private List<Object> getPkValuesByDefault() throws SQLException {
+        Map<String, ColumnMeta> pkMetaMap = checkPrimaryKeyMap();
+        ColumnMeta pkMeta = pkMetaMap.values().iterator().next();
+        String columnDef = pkMeta.getColumnDef();
+        // sample: nextval('test_id_seq'::regclass)
+        String seq = org.apache.commons.lang.StringUtils.substringBetween(columnDef, "'", "'");
+        String function = org.apache.commons.lang.StringUtils.substringBetween(columnDef, "", "(");
+        if (StringUtils.isBlank(seq)) {
+            throw new ShouldNeverHappenException("columnDef is " + columnDef);
```
l81893521, line 199: I think the Exception should give more detail. like
```
throw new ShouldNeverHappenException("get primary key value failed, cause columnDef is " + columnDef);
```

rm-datasource/src/main/java/io/seata/rm/datasource/exec/InsertExecutor.java
```diff
@@ -359,4 +388,12 @@ protected boolean checkPkValues(List<Object> pkValues) {
         return pkValues;
     }
 
+    private Map<String, ColumnMeta> checkPrimaryKeyMap() {
```
l81893521, line 391: How about remove this method? Because multi primary key already check in TableMeta.getPrimaryKeyMap()

slievrly:
default as value check valid.
image

PR #2348, opened 2020-03-04
optimize:remove redundant configuration

a364176773:
remove redundant configuration


zjinlei:
io.seata.spring.boot.autoconfigure.properties.registry is also required

Issue #2350, opened 2020-03-04
optimize: Generated instances that were not actually used when the class was loaded


zjinlei:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
![image](https://user-images.githubusercontent.com/20024955/75879946-4de15480-5e57-11ea-9c28-3a216c65aa1f.png)
MysqlTableMetaCache,PostgresqlTableMetaCache ,JacksonUndoLogParser, etc. classes with static variables.

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

ph3636:
Please assign it to me.

PR #2364, opened 2020-03-06
optimize: Generated instances that were not actually used when the class was loaded

ph3636:
…aded

optimize: Generated instances that were not actually used when the class was loaded

#2350


rm-datasource/src/main/java/io/seata/rm/datasource/undo/parser/FastjsonUndoLogParser.java
```diff
@@ -30,14 +30,15 @@
  * @author sharajava
  */
 @LoadLevel(name = FastjsonUndoLogParser.NAME)
-public class FastjsonUndoLogParser implements UndoLogParser {
+public class FastjsonUndoLogParser implements UndoLogParser, Initialize {
 
     public static final String NAME = "fastjson";
 
-    private static final SimplePropertyPreFilter FILTER = new SimplePropertyPreFilter();
+    private SimplePropertyPreFilter filter = new SimplePropertyPreFilter();
```
zjinlei: 'final' can be considered reserved, as are other classes.

ph3636: done

Issue #2353, opened 2020-03-04
feature: FailureHandler implement can be read from the container


zjinlei:
## Why you need it?
 Is your feature request related to a problem? Please describe in details
Currently, it can only be injected through the constructor of GlobalTransactionScanner. A more deterministic way is needed.
Reference ObjectHolder.java

## How it could be?
A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

ph3636:
Please assign it to me.

PR #2370, opened 2020-03-06
feature: failureHandler implement can be read from the container

ph3636:
FailureHandler implement can be read from the container

#2353


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -287,6 +283,17 @@ public void afterPropertiesSet() {
         }
         initClient();
 
+        setCustomFailureHandler();
+    }
+
+    private void setCustomFailureHandler() {
+        if (failureHandlerHook == null) {
+            try {
+                failureHandlerHook = applicationContext.getBean(FailureHandler.class);
```
zjinlei: test failed, FailureHandler has not yet been instantiated.

ph3636: Unit tests have passed,Is the custom FailureHandler in the container?If there is no customization, FailureHandler  is null,Because GlobalTransactionalInterceptor has a default

zjinlei: Custom FailureHandler implementation need be instantiated before GlobalTransactionScanner initialization.
Unit tests initialize the GlobalTransactionScanner first.

ph3636:
> globalTransactionScanner

ph3636:
> Why not load it from the globalTransactionScanner in starter ?

You mean it's more appropriate?
【
@Bean
    @DependsOn({BEAN_NAME_SPRING_APPLICATION_CONTEXT_PROVIDER})
    @ConditionalOnMissingBean(GlobalTransactionScanner.class)
    public GlobalTransactionScanner globalTransactionScanner(SeataProperties seataProperties) {
        if (LOGGER.isInfoEnabled()) {
            LOGGER.info("Automatically configure Seata");
        }
        FailureHandler customFailureHandler = null;
        try {
            customFailureHandler = ObjectHolder.INSTANCE.getObject(ApplicationContext.class).getBean(FailureHandler.class);
        } catch (NoSuchBeanDefinitionException ignore) {
        }
        return new GlobalTransactionScanner(seataProperties.getApplicationId(), seataProperties.getTxServiceGroup(), customFailureHandler);
    }
】

I think my approach is at the bottom of the way, no matter which way he chooses, it can ensure loading.

PR #2407, opened 2020-03-16
optimize: Inaccurate judgment will lead to NPE.

threefish:
Inaccurate judgment will lead to NullPointerException. After the global web exception is caught, the successful return result will be misjudged as a failure.

### DIFF
--- a/integration/http/src/main/java/io/seata/integration/http/XidResource.java
+++ b/integration/http/src/main/java/io/seata/integration/http/XidResource.java
@@ -15,6 +15,7 @@
  */
 package io.seata.integration.http;
 
+import io.seata.common.util.StringUtils;
 import io.seata.core.context.RootContext;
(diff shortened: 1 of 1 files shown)


integration/http/src/main/java/io/seata/integration/http/XidResource.java
```diff
@@ -36,7 +36,7 @@ public static void cleanXid(String rpcXid) {
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug("unbind[{}] from RootContext", unbindXid);
             }
-            if (!rpcXid.equalsIgnoreCase(unbindXid)) {
+            if (rpcXid != null && !rpcXid.equalsIgnoreCase(unbindXid)) {
```
jsbxyyx: !StringUtils.equalsIgnoreCase ?

PR #2409, opened 2020-03-16
optimize: reduce the db and network request when undoLog  or lockKey is empty

lightClouds917:

[issue](https://github.com/seata/seata/issues/2388)
optimize: unnecessary to register branch transaction when undo is empty


lightClouds917:
@zjinlei  PTAL,THS

rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
```diff
@@ -211,23 +211,32 @@ private void processLocalCommitWithGlobalLocks() throws SQLException {
...
         } catch (Throwable ex) {
-            LOGGER.error("process connectionProxy commit error: {}", ex.getMessage(), ex);
-            report(false);
-            throw new SQLException(ex);
+            if (hasUndoLog) {
+                LOGGER.error("process connectionProxy commit error: {}", ex.getMessage(), ex);
+                report(false);
+                throw new SQLException(ex);
+            } else {
+                LOGGER.error("process connectionProxy commit error: {},and the undolog is null", ex.getMessage(), ex);
```
zjinlei, line 236: lack throw new SQLException(ex);

lightClouds917: ok,i will fix

rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
```diff
@@ -211,24 +211,30 @@ private void processLocalCommitWithGlobalLocks() throws SQLException {
...
-            register();
-        } catch (TransactionException e) {
-            recognizeLockKeyConflictException(e, context.buildLockKeys());
-        }
-
-        try {
-            if (context.hasUndoLog()) {
+        if (context.hasUndoLog()) {
+            try {
+                register();
```
slievrly, line 216: I think it's better to put it in registers and flushUndoLogs.
if there is no undolog why flushUndoLogs method execution without judgment

lightClouds917: I have optimized the code, PTAL

lightClouds917: I have finish it.PTAL

PR #2439, opened 2020-03-22
bugfix: it will cause memory overflow when client reconnect frequently.

objcoding:
1. remove channel from identified channels map if expired;
2. optimize ChannelManager class code style.
fix https://github.com/seata/seata/issues/2531


core/src/main/java/io/seata/core/rpc/ChannelManager.java
```diff
@@ -43,22 +43,32 @@
...
-        = new ConcurrentHashMap<Channel, RpcContext>();
+
+    private static final ConcurrentMap<Channel, RpcContext> IDENTIFIED_CHANNELS = new ConcurrentHashMap<>();
+
+    /**
+     * the tm client active channel: clientIdentified(applicationId:ip) -> channel
+     * Used to store the active channel of the tm client,
+     * When the tm client reconnects, the value will be updated.
+     */
+    private static final ConcurrentMap<String, Channel> TM_CLIENT_ACTIVE_CHANNELS = new ConcurrentHashMap<>();
```
l81893521: what TM_CLIENT_ACTIVE_CHANNELS and RM_CLIENT_ACTIVE_CHANNELS  use for?

objcoding: Used to record the latest channel of the current client.

objcoding: When the customer re-registers, it will check whether TM_CLIENT_ACTIVE_CHANNELS and RM_CLIENT_ACTIVE_CHANNELS have channels, if so, it will be removed from ss to avoid IDENTIFIED_CHANNELS cache accumulation.

core/src/main/java/io/seata/core/rpc/ChannelManager.java
```diff
@@ -118,54 +128,80 @@ private static RpcContext buildChannelHolder(NettyPoolKey.TransactionRole client
...
+
+        RpcContext rpcContext;
+        if (!IDENTIFIED_CHANNELS.containsKey(newChannel)) {
+            rpcContext = buildChannelHolder(NettyPoolKey.TransactionRole.TMROLE, request.getVersion(),
+                request.getApplicationId(),
+                request.getTransactionServiceGroup(),
+                null, newChannel);
+            rpcContext.holdInIdentifiedChannels(IDENTIFIED_CHANNELS);
+        } else {
+            rpcContext = IDENTIFIED_CHANNELS.get(newChannel);
```
l81893521: I do not think it will be execute here.

objcoding: Seems to be such a truth.

objcoding: resolved it.

core/src/main/java/io/seata/core/rpc/ChannelManager.java
```diff
@@ -118,54 +128,72 @@ private static RpcContext buildChannelHolder(NettyPoolKey.TransactionRole client
...
      * @throws IncompatibleVersionException the incompatible version exception
      */
-    public static void registerTMChannel(RegisterTMRequest request, Channel channel)
-        throws IncompatibleVersionException {
+    public static void registerTMChannel(RegisterTMRequest request, Channel newChannel) throws IncompatibleVersionException {
         Version.checkVersion(request.getVersion());
+
+        // remove channel from identified channels if expired.
+        String clientIdentified = request.getApplicationId() + Constants.CLIENT_ID_SPLIT_CHAR
+            + ChannelUtil.getClientIpFromChannel(newChannel);
```
ph3636: Can be changed to a general method

core/src/main/java/io/seata/core/rpc/ChannelManager.java
```diff
@@ -118,54 +128,72 @@ private static RpcContext buildChannelHolder(NettyPoolKey.TransactionRole client
...
+        Version.checkVersion(request.getVersion());
+        Set<String> dbkeySet = dbKeytoSet(request.getResourceIds());
+
+        // remove channel from identified channels if expired.
+        String clientIdentified = request.getApplicationId() + Constants.CLIENT_ID_SPLIT_CHAR
+            + ChannelUtil.getClientIpFromChannel(newChannel);
+        Channel expiredChannel = RM_CLIENT_ACTIVE_CHANNELS.get(clientIdentified);
+        if (expiredChannel != null) {
+            releaseRpcContext(expiredChannel);
+            IDENTIFIED_CHANNELS.remove(expiredChannel);
```
ph3636: 【IDENTIFIED_CHANNELS.remove(expiredChannel);】Can be put into method releaseRpcContext

objcoding: > 【IDENTIFIED_CHANNELS.remove(expiredChannel);】Can be put into method releaseRpcContext

IDENTIFIED_CHANNELS.remove (expiredChannel) will only be executed for channels with expired rr, so this operation cannot be put into releaseRpcContext (expiredChannel).

core/src/main/java/io/seata/core/rpc/ChannelManager.java
objcoding: > Can be changed to a general method

Done.

core/src/main/java/io/seata/core/rpc/ChannelManager.java
objcoding: The getRoleFromChannel method is no longer useful.

core/src/main/java/io/seata/core/rpc/ChannelManager.java
l81893521, line 54: How about remove these two map, and made the IDENTIFIED_CHANNELS correct at
```
io.seata.core.rpc.netty.AbstractRpcRemotingServer.ServerHandler#handleDisconnect
```

objcoding: When the client re-registers, it will update the map value.

core/src/main/java/io/seata/core/rpc/ChannelManager.java
```diff
@@ -118,54 +128,74 @@ private static RpcContext buildChannelHolder(NettyPoolKey.TransactionRole client
...
-            rpcContext.holdInIdentifiedChannels(IDENTIFIED_CHANNELS);
-        } else {
-            rpcContext = IDENTIFIED_CHANNELS.get(channel);
-            rpcContext.addResources(dbkeySet);
+        Version.checkVersion(request.getVersion());
+        Set<String> dbkeySet = dbKeytoSet(request.getResourceIds());
+
+        // remove channel from identified channels if expired.
+        String clientIdentified = getClientIdentified(request.getApplicationId(), newChannel);
+        Channel expiredChannel = RM_CLIENT_ACTIVE_CHANNELS.get(clientIdentified);
```
slievrly, line 176: If more than one client IP has the same applicationId application how to distinguish?

PR #2440, opened 2020-03-23
optimize: optimize contact us and startup log

slievrly:

optimize: optimize contact us and startup log

1. The QR code of dingTalk into group is invalid

2. WeChat group cannot be actively entered into the group and replaced with WeChat official account

3. Optimize the problem that the output log is null at startup

4. change the log level when startup delete undolog.


xingfudeshi:
LGTM.

PR #2452, opened 2020-03-25
CollectionUtils.decodeMap method variable ConcurrentHashMap refactor to HashMap

xianlaioy:
CollectionUtils.decodeMap method variable  ConcurrentHashMap refactor to HashMap

### 2. Local variables are stored in each thread's own stack. That means that local variables are never shared between threads. That also means that all local primitive variables are thread safe.

Local references to objects are a bit different. The reference itself is not shared. The object referenced however, is not stored in each threads's local stack. All objects are stored in the shared heap. If an object created locally never escapes the method it was created in, it is thread safe. In fact you can also pass it on to other methods and objects as long as none of these methods or objects make the passed object available to other threads.
https://stackoverflow.com/questions/11535237/local-variables-in-static-method-and-thread-safety

### DIFF
--- a/common/src/main/java/io/seata/common/util/CollectionUtils.java
+++ b/common/src/main/java/io/seata/common/util/CollectionUtils.java
@@ -17,9 +17,9 @@ package io.seata.common.util;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * The type Collection utils.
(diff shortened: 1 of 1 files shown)


common/src/main/java/io/seata/common/util/CollectionUtils.java
```diff
@@ -82,8 +82,7 @@ public static String toString(Collection<?> col) {
         if (isEmpty(col)) {
             return "";
         }
-        StringBuilder sb = new StringBuilder();
-        sb.append("[");
+        StringBuilder sb = new StringBuilder("]");
```
slievrly: `[`

slievrly:
@xianlaioy 
image

PR #2466, opened 2020-03-27
bugfix: memory visibility of active attribute in file mode

ph3636:

bugfix: memory visibility of active attribute in file mode

close the session, then no more branch can be registered.


zjinlei:
retry CI

PR #2473, opened 2020-03-28
bugfix: flush condition of disk in file mode

ph3636:
bugfix: flush condition of disk in file mode


slievrly:
image
image
In the past it was just flush when idle.

PR #2476, opened 2020-03-28
Optimizing sqlType actuator based on static factory pattern

pengzhengfa:

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/ExecuteTemplate.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/ExecuteTemplate.java
@@ -19,6 +19,7 @@ import io.seata.core.context.RootContext;
 import io.seata.rm.datasource.StatementProxy;
 import io.seata.rm.datasource.sql.SQLVisitorFactory;
 import io.seata.sqlparser.SQLRecognizer;
+import io.seata.sqlparser.SQLType;
 
 import java.sql.SQLException;
 import java.sql.Statement;
@@ -26,6 +27,7 @@ import java.sql.Statement;
 /**
  * The type Execute template.
  *
+ * @author pengzhengfa
  * @author sharajava
  */
 public class ExecuteTemplate {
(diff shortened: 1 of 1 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/exec/ExecuteTemplate.java
```diff
@@ -78,23 +80,7 @@
...
-                    executor = new DeleteExecutor<>(statementProxy, statementCallback, sqlRecognizer);
-                    break;
-                case SELECT_FOR_UPDATE:
-                    executor = new SelectForUpdateExecutor<>(statementProxy, statementCallback, sqlRecognizer);
-                    break;
-                default:
-                    executor = new PlainExecutor<>(statementProxy, statementCallback);
-                    break;
-            }
+            executor = createSqlExecutor(sqlRecognizer.getSQLType(), sqlRecognizer, statementProxy, statementCallback);
```
l81893521: I think
```
createSqlExecutor(sqlRecognizer, statementProxy, statementCallback)
```
that is enough.

pengzhengfa: @l81893521 Thank you for your advice

pengzhengfa: @l81893521 Createsqlexecutor (sqlrecognizer, statementproxy, statementcallback) needs to return parameters to execute SQL for the executor

l81893521: I m apologize that I did not clear my point. I mean 3 parameters that s enough. just like that below.
```
executor = createSqlExecutor(sqlRecognizer, statementProxy, statementCallback);
```

pengzhengfa: @l81893521 OK, I'll change it

PR #2490, opened 2020-04-01
bugfix: fix RpcContext.addResource when resource is null

dk-lockdown:
RpcContext.addResource(String resource),when resource is blank,should do nothing,instead of reset resourceSets

the original code is
```
public void addResource(String resource) {
        if (null == resource) {
            this.resourceSets = new HashSet<String>();
        }
        this.resourceSets.add(resource);
    }
```
obviously, there is a bug


jsbxyyx:
please git relation github user.name and user.email

slievrly:
I made sure that this method is only used in UT.

PR #2495, opened 2020-04-01
bugfix:  fix the NPE and reduce the request when lockkey is null

lightClouds917:
1.fix: Fix the NPE when lockkey is null;
2.optimize:  Reduce once db request when lockkey is null;

yes.
[issue](https://github.com/seata/seata/issues/2496)

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
@@ -15,8 +15,10 @@
  */
 package io.seata.rm.datasource;
 
+import io.seata.common.util.CollectionUtils;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
 import io.seata.common.util.StringUtils;
@@ -198,8 +200,10 @@ public class ConnectionProxy extends AbstractConnectionProxy {
     }
 
     private void processLocalCommitWithGlobalLocks() throws SQLException {
-
-        checkLock(context.buildLockKeys());
+        Set<String> lockKeys = context.getLockKeysBuffer();
+        if (CollectionUtils.isNotEmpty(lockKeys)) {
+            checkLock(context.buildLockKeys());
+        }
         try {
             targetConnection.commit();
         } catch (Throwable ex) {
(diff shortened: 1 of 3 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
```diff
@@ -198,8 +200,10 @@ private void doCommit() throws SQLException {
     }
 
     private void processLocalCommitWithGlobalLocks() throws SQLException {
-
-        checkLock(context.buildLockKeys());
+        Set<String> lockKeys = context.getLockKeysBuffer();
+        if (CollectionUtils.isNotEmpty(lockKeys)) {
```
slievrly, line 204: Is it more appropriate to put logic with lockkey as empty in the checkLock method?

lightClouds917: ok,i will change

server/src/main/java/io/seata/server/lock/AbstractLockManager.java
```diff
@@ -153,6 +153,9 @@ protected Locker getLocker() {
                                             Long branchID) {
         List<RowLock> locks = new ArrayList<RowLock>();
 
+        if (StringUtils.isEmpty(lockKey)) {
+            return null;
+        }
```
zjinlei, line 158: Incorrect

slievrly: check  `isLockable` method.

server/src/main/java/io/seata/server/lock/AbstractLockManager.java
```diff
@@ -153,6 +153,9 @@ protected Locker getLocker() {
                                             Long branchID) {
         List<RowLock> locks = new ArrayList<RowLock>();
 
+        if (StringUtils.isEmpty(lockKey)) {
```
l81893521, line 156: That s ok. And I think can remove the code
```
if (StringUtils.isNullOrEmpty(lockKey)) {
            // no lock
            return true;
}
```
at 
```
acquireLock(BranchSession branchSession)
```

lightClouds917: `isLockable `will call this also，and it has no NEP judgement;

rm-datasource/src/main/java/io/seata/rm/datasource/ConnectionProxy.java
Garlichead: image
the master code

PR #2523, opened 2020-04-09
optimize: abnormal global transactions, output logs according to frequency

a364176773:
optimize log output when rm client disconnects abnormally


core/src/main/java/io/seata/core/rpc/netty/RpcServer.java
```diff
@@ -109,9 +111,9 @@ public Object sendSyncRequest(String resourceId, String clientId, Object message
                                   long timeout) throws TimeoutException {
         Channel clientChannel = ChannelManager.getChannel(resourceId, clientId);
         if (clientChannel == null) {
-            throw new RuntimeException("rm client is not connected. dbkey:" + resourceId
-                + ",clientId:" + clientId);
-
+            StackTraceLogger.warn(LOGGER,
+                new RuntimeException("rm client is not connected. dbkey:" + resourceId + ",clientId:" + clientId),
```
slievrly, line 115: If these two rows are the same, why don't we just format them once?

core/src/main/java/io/seata/core/rpc/netty/RpcServer.java
```diff
@@ -109,9 +111,9 @@ public Object sendSyncRequest(String resourceId, String clientId, Object message
...
+                "rm client is not connected. dbkey: {},clientId: {}", new String[] {resourceId, clientId});
```
slievrly, line 116: can this logic continue?

a364176773: PTAL @slievrly

PR #2530, opened 2020-04-09
optimize: remove use connPool.

objcoding:
remove use connPool, Because apache object-pool is already used.


core/src/main/java/io/seata/core/rpc/netty/RpcClientBootstrap.java
```diff
@@ -129,57 +128,29 @@ public void start() {
                     .option(EpollChannelOption.TCP_QUICKACK, true);
             }
         }
-        if (nettyClientConfig.isUseConnPool()) {
-            clientChannelPool = new AbstractChannelPoolMap<InetSocketAddress, FixedChannelPool>() {
```
l81893521, line 133: clientChannelPool is never used, except shutdown.

objcoding: Done.

Issue #2548, opened 2020-04-14
使用zk作为配置中心，seata能连接zk但无法读取zk配置，且无法查看Exception 异常堆栈信息


luojiachengaccp:
2020-04-14 11:16:45.323 INFO [main]io.seata.config.FileConfiguration.<init>:101 -The file name of the operation is registry.conf
2020-04-14 11:16:45.443 WARN [main]io.seata.config.ConfigurationFactory.<clinit>:68 -failed to load extConfiguration:not found service provider for : io.seata.config.ExtConfigurationProvider[null] and classloader : sun.misc.Launcher$AppClassLoader@77459877
io.seata.common.loader.EnhancedServiceNotFoundException: not found service provider for : io.seata.config.ExtConfigurationProvider[null] and classloader : sun.misc.Launcher$AppClassLoader@77459877
2020-04-14 11:16:45.458 INFO [main]io.seata.common.loader.EnhancedServiceLoader.loadFile:247 -load ConfigurationProvider[ZK] extension by class[io.seata.config.zk.ZookeeperConfigurationProvider]
2020-04-14 11:16:45.471 INFO [ZkClient-EventThread-11-10.72.27.60:2181,10.72.27.160:2181,10.72.27.150:2081]org.I0Itec.zkclient.ZkEventThread.run:65 -Starting ZkClient event thread.
2020-04-14 11:16:45.477 INFO [main]org.apache.zookeeper.ZooKeeper.logEnv:100 -Client environment:zookeeper.version=3.4.14-4c25d480e66aadd371de8bd2fd8da255ac140bcf, built on 03/06/2019 16:18 GMT
2020-04-14 11:16:45.477 INFO [main]org.apache.zookeeper.ZooKeeper.logEnv:100 -Client environment:host.name=tclshop-pc2-green
2020-04-14 11:16:45.477 INFO [main]org.apache.zookeeper.ZooKeeper.logEnv:100 -Client environment:java.version=1.8.0_111
2020-04-14 11:16:45.477 INFO [main]org.apache.zookeeper.ZooKeeper.logEnv:100 -Client environment:java.vendor=Oracle Corporation
2020-04-14 11:16:45.477 INFO [main]org.apache.zookeeper.ZooKeeper.logEnv:100 -Client environment:java.home=/kuyu/soft/jdk1.8.0_111/jre
[...]

PR #2549, opened 2020-04-14
optimize: optimize ZookeeperConfiguration log

lightClouds917:
optimize:print the error message when getConfig from zk

optimize #2548


config/seata-config-zk/src/main/java/io/seata/config/zk/ZookeeperConfiguration.java
```diff
@@ -117,7 +117,8 @@ public String getConfig(String dataId, String defaultValue, long timeoutMills) {
         try {
             return future.get(timeoutMills, TimeUnit.MILLISECONDS);
         } catch (Exception e) {
-            LOGGER.error("getConfig {} is error or timeout,return defaultValue {}", dataId, defaultValue);
+            LOGGER.error("getConfig {} is error or timeout,return defaultValue {},error : {}",
```
jsbxyyx, line 120: after comma add space?

lightClouds917: > after comma add space?

ok,I will fix

PR #2554, opened 2020-04-15
bugfix: add zk serializer.

objcoding:
add zk serializer.

fixes https://github.com/seata/seata/issues/2337


a364176773:
server/src/main/resources/registry.conf need to add?

config/seata-config-zk/src/main/java/io/seata/config/zk/ZookeeperConfiguration.java
```diff
@@ -245,4 +250,23 @@ public void handleDataDeleted(String s) {
...
+    private ZkSerializer getZkSerializer() {
+        ZkSerializer zkSerializer = null;
+        String serializer = FILE_CONFIG.getConfig(FILE_CONFIG_KEY_PREFIX + SERIALIZER_KEY);
+        if (StringUtils.isNotBlank(serializer)) {
+            try {
+                Class<?> aClass = Class.forName(serializer);
+                zkSerializer = (ZkSerializer) aClass.newInstance();
+            } catch (ClassNotFoundException cfe) {
+                LOGGER.error("No zk serializer class found, serializer:{}", serializer, cfe);
+            } catch (Exception e) {
```
jsbxyyx, line 262: Exception replace to Throwable ? Class.forName can be appears LinkageError

objcoding: Done.

objcoding:
> registry.c

Yes, it needs to be added here.

server/src/main/resources/registry.conf
```diff
@@ -54,7 +54,7 @@ registry {
 
 config {
   # file、nacos 、apollo、zk、consul、etcd3
-  type = "file"
+  type = "zk"
```
l81893521, line 57: Default config type is file.

objcoding: Done.

config/seata-config-zk/src/main/java/io/seata/config/zk/ZookeeperConfiguration.java
```diff
@@ -247,4 +252,23 @@ public void handleDataDeleted(String s) {
...
+                Class<?> aClass = Class.forName(serializer);
+                zkSerializer = (ZkSerializer) aClass.newInstance();
+            } catch (ClassNotFoundException cfe) {
+                LOGGER.error("No zk serializer class found, serializer:{}", serializer, cfe);
+            } catch (Throwable cause) {
+                LOGGER.error("found zk serializer encountered an unknown exception", cause);
+            }
+        }
+        if (zkSerializer == null) {
+            zkSerializer = new DefaultZkSerializer();
```
l81893521, line 269: How about use log to tell users, zkSerializer would use DefaultZkSerializer

objcoding: Done.

config/seata-config-core/src/main/resources/registry.conf
```diff
@@ -66,6 +66,7 @@ config {
     serverAddr = "127.0.0.1:2181"
     sessionTimeout = 6000
     connectTimeout = 2000
+    serializer = ""
```
slievrly, line 69: It is not recommended that the serializer be exposed to the user to increase the cost of understanding what code is actually defined for the configuration item.

objcoding: This suggestion is very good.

PR #2616, opened 2020-04-26
feature: TCC adapter for Dubbo And Sofa reference annotation

q294881866:
Support Dubbo @Reference and sofa @SofaReference  annotation for TCC mode


spring/src/main/java/io/seata/spring/tcc/TccAnnotationProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+     * @param beanName
+     * @param annotation
+     */
+    protected void process(Object bean, String beanName, Class<? extends Annotation> annotation) {
+        if (Objects.isNull(annotation) || PROXIED_SET.contains(beanName)) {
+            return;
+        }
+
+        ReflectionUtils.doWithFields(bean.getClass(), field -> {
+            Annotation reference = field.getAnnotation(annotation);
```
slievrly, line 75: if necessary? FieldFilter check the filed contains annotation.

spring/src/main/java/io/seata/spring/tcc/TccAnnotationProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+            if (reference == null) {
+                return;
+            }
+
+            addTccAdvise(bean, beanName, field, field.getType());
```
slievrly, line 80: why not getInterface()?

q294881866: Previously, it was implemented with an interface, but there were many repetitions. In fact, it's the only one that's different. There's no need to use an interface
` static {
        ANNOTATIONS.add(loadAnnotation("org.apache.dubbo.config.annotation.Reference"));
        ANNOTATIONS.add(loadAnnotation("com.alipay.sofa.runtime.api.annotation.SofaReference"));
        ANNOTATIONS.add(loadAnnotation( ...
    }`

spring/src/main/java/io/seata/spring/tcc/TccAnnotationProcessor.java
q294881866: The code impossible to reuse.

q294881866: > if necessary? FieldFilter check the filed contains annotation.
The code impossible to reuse.

slievrly: What I mean is that reference can never null， `field.isAnnotationPresent(annotation)`

PR #2622, opened 2020-04-27
feature: support version valid check

slievrly:

feature: support version valid check


core/src/main/java/io/seata/core/protocol/Version.java
```diff
@@ -69,8 +76,35 @@ public static String getChannelVersion(Channel c) {
...
      * @throws IncompatibleVersionException the incompatible version exception
      */
-    public static String checkVersion(String version) throws IncompatibleVersionException {
-        // TODO: check
-        return version;
+    public static void checkVersion(String version) throws IncompatibleVersionException {
+        long current = convertVersion(CURRENT);
+        long clientVersion = convertVersion(version);
+        long divideVersion = convertVersion(VERSION_0_7_1);
+        if ((current - divideVersion > 0 && clientVersion - divideVersion < 0) || (current - divideVersion < 0 && clientVersion - divideVersion > 0)) {
```
ph3636, line 83: Is it clearer to change 【current - divideVersion > 0】 to 【current > divideVersion】

core/src/main/java/io/seata/core/protocol/Version.java
```diff
@@ -69,8 +76,35 @@ public static String getChannelVersion(Channel c) {
...
+            throw new IncompatibleVersionException("incompatible client version:" + version);
+        }
+    }
+
+    private static long convertVersion(String version) throws IncompatibleVersionException {
+        String parts[] = StringUtils.split(version, '.');
+        Long result = 0L;
+        int i = 1;
+        int size = parts.length;
+        if (size > MAX_VERSION_DOT + 1) {
```
ph3636, line 93: Why not fix it to 3?  
if (size != 3 ) {
throw ...
}

slievrly: some version such as： 1.3.0，1.3.0.1，1.3.0-SNAPSHOT ...

core/src/main/java/io/seata/core/protocol/Version.java
slievrly: fixed

core/src/main/java/io/seata/core/protocol/Version.java
ph3636: If so, The value of result is that 1.3.0.1 is larger than 1.4.0?

slievrly: get what you mean. fix later

slievrly: @ph3636 how about this one?

ph3636: good

core/src/main/java/io/seata/core/protocol/Version.java
```diff
@@ -69,8 +76,36 @@ public static String getChannelVersion(Channel c) {
...
      * @throws IncompatibleVersionException the incompatible version exception
      */
-    public static String checkVersion(String version) throws IncompatibleVersionException {
-        // TODO: check
-        return version;
+    public static void checkVersion(String version) throws IncompatibleVersionException {
+        long current = convertVersion(CURRENT);
+        long clientVersion = convertVersion(version);
+        long divideVersion = convertVersion(VERSION_0_7_1);
+        if ((current > divideVersion && clientVersion < divideVersion) || (current < divideVersion && clientVersion > divideVersion)) {
```
zjinlei, line 83: The verification function has not yet been reached and an exception has been thrown: io.netty.handler.codec.TooLongFrameException: Adjusted frame length exceeds 8388608: 2415945472-discarded

PR #2629, opened 2020-04-28
bugfix: fix duplicated resource id in PostgreSQL

l81893521:
In this case
```
jdbc:postgresql://127.0.0.1:5432/seata?currentSchema=public
jdbc:postgresql://127.0.0.1:5432/seata?currentSchema=seata
```

it will cause the same resource id. and the error below occur.
1.Get file lock fail.
2.Cause the wrong table meta in pg sql.

#2623

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
@@ -141,12 +141,20 @@ public class DataSourceProxy extends AbstractDataSourceProxy implements Resource
 
     @Override
     public String getResourceId() {
+        if (DBType.POSTGRESQL.name().equalsIgnoreCase(getDbType())) {
+            return getPGResourceId();
+        } else {
+            return getDefaultResourceId();
+        }
+    }
+
+    /**
+     * get the default resource id
+     * @return resource id
+     */
+    private String getDefaultResourceId() {
         if (jdbcUrl.contains("?")) {
-            if (DBType.POSTGRESQL.name().equalsIgnoreCase(getDbType())) {
-                return getPGResourceId();
-            } else {
-                return jdbcUrl.substring(0, jdbcUrl.indexOf('?'));
-            }
+            return jdbcUrl.substring(0, jdbcUrl.indexOf('?'));
         } else {
             return jdbcUrl;
         }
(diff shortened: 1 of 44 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -141,7 +141,32 @@ public String getResourceGroupId() {
...
+             * 1.get file lock fail
+             * 2.error table meta cache
+             */
+            StringBuilder paramsBuilder = new StringBuilder();
+            String paramUrl = jdbcUrl.substring(jdbcUrl.indexOf('?') + 1, jdbcUrl.length());
+            String[] urlParams = paramUrl.split("&");
+            for (String urlParam : urlParams) {
+                if (urlParam.contains("currentSchema")) {
+                    paramsBuilder.append(urlParam);
+                }
```
zjinlei, line 161: 1.currentSchema is magic number
2.break?

l81893521: ok. Already add break

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -141,7 +141,33 @@ public String getResourceGroupId() {
     @Override
     public String getResourceId() {
         if (jdbcUrl.contains("?")) {
-            return jdbcUrl.substring(0, jdbcUrl.indexOf('?'));
+            StringBuilder jdbcUrlBuilder = new StringBuilder();
```
zjinlei, line 144: According to dbType, non-pg remains the same

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -141,7 +141,33 @@ public String getResourceGroupId() {
...
+                }
+            }
+
+            if (paramsBuilder.length() > 0) {
+                jdbcUrlBuilder.append("?");
+                jdbcUrlBuilder.append(paramsBuilder);
+            }
+
+            return jdbcUrlBuilder.toString();
         } else {
```
jsbxyyx, line 171: remove else, direct return?

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
l81893521: fixed

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
l81893521: I think
```
If() {

} else {

}
```
is more readable.

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -141,7 +142,35 @@ public String getResourceGroupId() {
     @Override
     public String getResourceId() {
         if (jdbcUrl.contains("?")) {
-            return jdbcUrl.substring(0, jdbcUrl.indexOf('?'));
```
zjinlei, line 144: Keep the original clear, suggest:
return DBType.POSTGRESQL.name().equalsIgnoreCase(getDbType())? getPgResourceId() : jdbcUrl.substring(0,jdbcUrl.indexOf('?'));

l81893521: fixed.

zjinlei: Only separate the different parts.
if (!jdbcUrl.contains("?")) is the same

l81893521: as you say, the code would be like this 
```
if (jdbcUrl.contains("?")) {
    if (DBType.POSTGRESQL.name().equalsIgnoreCase(getDbType())) {
        getPGResourceId();
    } else {
      do something....
    }
} else {
    do something
}
```
the code is less, but the logic is more complicated.

PR #2642, opened 2020-04-30
bugfix: unsafe double-checked locking in SofaRegistryServiceImpl

chenxi-null:
This usage of double check locking is not thread-safe, the reason is present in this PR [#1599](https://github.com/seata/seata/pull/1599) which I submitted before.

How to fix:

Because `DefaultRegistryClient` is in third-part jar, the `init` method invocation can't be moved to the constructor.

So the local variable is used to correct DCL and minimize the number of field reads.

See also: https://lgtm.com/rules/1507076816054/

Quote from the article:
```java
// a local variable can be used to avoid reading the field more times than neccessary.

private Object lock = new Object();
private volatile MyObject f = null;

public MyObject getMyObject() {
  MyObject result = f;
  if (result == null) {
    synchronized(lock) {
      result = f;
      if (result == null) {
        result = new MyObject();
        result.init();
        f = result; // GOOD
      }
    }
  }
  return result;
}
```

### DIFF
--- a/discovery/seata-discovery-sofa/src/main/java/io/seata/discovery/registry/sofa/SofaRegistryServiceImpl.java
+++ b/discovery/seata-discovery-sofa/src/main/java/io/seata/discovery/registry/sofa/SofaRegistryServiceImpl.java
@@ -118,9 +118,11 @@ public class SofaRegistryServiceImpl implements RegistryService<SubscriberDataOb
     }
 
     private RegistryClient getRegistryInstance() {
-        if (null == registryClient) {
+        RegistryClient result = registryClient;
+        if (null == result) {
             synchronized (SofaRegistryServiceImpl.class) {
-                if (null == registryClient) {
+                result = registryClient;
+                if (null == result) {
                     String address = registryProps.getProperty(PRO_SERVER_ADDR_KEY);
                     final String portStr = StringUtils.substringAfter(address, HOST_SEPERATOR);
 
@@ -131,12 +133,13 @@ public class SofaRegistryServiceImpl implements RegistryService<SubscriberDataOb
                         .setRegistryEndpoint(StringUtils.substringBefore(address, HOST_SEPERATOR))
                         .setRegistryEndpointPort(Integer.parseInt(portStr)).build();
 
-                    registryClient = new DefaultRegistryClient(config);
-                    ((DefaultRegistryClient)registryClient).init();
+                    result = new DefaultRegistryClient(config);
+                    ((DefaultRegistryClient) result).init();
+                    registryClient = result;
                 }
             }
         }
-        return registryClient;
+        return result;
     }
 
     @Override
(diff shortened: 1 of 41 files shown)


discovery/seata-discovery-sofa/src/main/java/io/seata/discovery/registry/sofa/SofaRegistryServiceImpl.java
```diff
@@ -131,12 +133,13 @@ private RegistryClient getRegistryInstance() {
                         .setRegistryEndpoint(StringUtils.substringBefore(address, HOST_SEPERATOR))
                         .setRegistryEndpointPort(Integer.parseInt(portStr)).build();
 
-                    registryClient = new DefaultRegistryClient(config);
-                    ((DefaultRegistryClient)registryClient).init();
+                    result = new DefaultRegistryClient(config);
+                    ((DefaultRegistryClient) result).init();
+                    registryClient = result;
```
zjinlei, line 138: You can continue to return registryClient, and result is used as a local transit variable, keeping the minimum changes.

xingfudeshi:
Please associate your GitHub account with your email.@chenxi-null

discovery/seata-discovery-sofa/src/main/java/io/seata/discovery/registry/sofa/SofaRegistryServiceImpl.java
chenxi-null: @zjinlei 
This way minimizes the number of volatile field reads, which improves performance, but makes the code a bit complicated. 

If minimum change or simple code is preferred, I will change the code to what you recommend ～

zjinlei:
> > Please associate your GitHub account with your email.@chenxi-null
> 
> @xingfudeshi Is there anything wrong with my email ?
> I have already associated with my GitHub account with the email `chenxiCoder@gmail.com`, and it can receive notification from Github.

https://blog.csdn.net/idealcitier/article/details/82697268

discovery/seata-discovery-sofa/src/main/java/io/seata/discovery/registry/sofa/SofaRegistryServiceImpl.java
zjinlei: [RegistryClient result = registryClient] == [null == registryClient]
[result = registryClient] == [null == registryClient]
Same reading times, and here is singleton initialization, code readability is more important.

chenxi-null: @zjinlei 
Just to confirm. Do you mean to change it like this?
```java
if (null == registryClient) {
    synchronized (SofaRegistryServiceImpl.class) {
        if (null == registryClient) {
            // ...
            DefaultRegistryClient result = new DefaultRegistryClient(config);
            result.init();
            registryClient = result;
        }
    }
}
return registryClient;
```

The original code;
```java
RegistryClient result = registryClient;
if (null == result) {
    synchronized (SofaRegistryServiceImpl.class) {
        result = registryClient;
        if (null == result) {
            // ...    
            result = new DefaultRegistryClient(config);
            ((DefaultRegistryClient) result).init();
            registryClient = result;
        }
    }
}
return result;
```

PR #2659, opened 2020-05-06
bugfix: fix mysql insert use select last insert id is undo_log id value

jsbxyyx:
mysql business insert use select last insert id is undo_log id's value.

fixes: #2655

### DIFF
--- a/script/client/at/db/mysql.sql
+++ b/script/client/at/db/mysql.sql
@@ -8,7 +8,7 @@ CREATE TABLE IF NOT EXISTS `undo_log`
     `log_status`    INT(11)      NOT NULL COMMENT '0:normal status,1:defense status',
     `log_created`   DATETIME(6)  NOT NULL COMMENT 'create datetime',
     `log_modified`  DATETIME(6)  NOT NULL COMMENT 'modify datetime',
-    PRIMARY KEY (`xid`, `branch_id`)
+    UNIQUE KEY `ux_undo_log` (`xid`, `branch_id`)
 ) ENGINE = InnoDB
   AUTO_INCREMENT = 1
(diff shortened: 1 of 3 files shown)


a364176773:
is mysql the only one that need to handle ?

jsbxyyx:
@a364176773 oracle, postgresql use sequence.

script/client/at/db/mysql.sql
```diff
@@ -1,16 +1,14 @@
...
     `branch_id`     BIGINT(20)   NOT NULL COMMENT 'branch transaction id',
     `xid`           VARCHAR(100) NOT NULL COMMENT 'global transaction id',
     `context`       VARCHAR(128) NOT NULL COMMENT 'undo_log context,such as serialization',
     `rollback_info` LONGBLOB     NOT NULL COMMENT 'rollback info',
     `log_status`    INT(11)      NOT NULL COMMENT '0:normal status,1:defense status',
-    `log_created`   DATETIME     NOT NULL COMMENT 'create datetime',
-    `log_modified`  DATETIME     NOT NULL COMMENT 'modify datetime',
-    PRIMARY KEY (`id`),
-    UNIQUE KEY `ux_undo_log` (`xid`, `branch_id`)
+    `log_created`   DATETIME(6)  NOT NULL COMMENT 'create datetime',
```
slievrly, line 9: Where is used the precision of time compared?

jsbxyyx: the log_created and log_modified field precision not use in current version. I think the datetime necessary precision to millisecond above.

slievrly: If that code needs to be changed to the same precision, now(6)

jsbxyyx: OK

PR #2660, opened 2020-05-06
optimize: modify IdWorker position to make it reasonable

a364176773:
modify IdWorker position to make it reasonable


a364176773:
@xingfudeshi @zjinlei PTAL

PR #2675, opened 2020-05-09
feature: freely choose context core.

objcoding:
1. freely choose context core;
2. add TransmittableThreadLocal context core.


a364176773:
i think there is a problem with the introduction of ttl. Even the child thread is involved in the global transaction.At this time, the parent thread has submitted the global transaction, and the child thread will commit again, which will trigger the previous problem. Asynchronous, the initiator of the global transaction can't perceive it. Unless they all use callable and get the result of the child thread to commit or roll back the transaction.

maybe asynchronous global transactions can come in handy

funky-eyes:

PR #2676, opened 2020-05-11
feature : support least active load balance

ph3636:

support LeastActiveLoadBalance


common/src/main/java/io/seata/common/rpc/RpcStatus.java
```diff
@@ -0,0 +1,93 @@
...
+    public static void removeStatus(String service) {
+        SERVICE_STATUS_MAP.remove(service);
+    }
+
+    /**
+     * begin count
+     *
+     * @param service the service
+     */
+    public static void beginCount(String service) {
```
l81893521, line 61: Never used except the test case, is the active always equals 1 for each address ?

discovery/seata-discovery-core/src/main/java/io/seata/discovery/loadbalance/LeastActiveLoadBalance.java
```diff
@@ -0,0 +1,53 @@
...
+package io.seata.discovery.loadbalance;
+
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+
+import io.seata.common.loader.LoadLevel;
+import io.seata.common.rpc.RpcStatus;
+
+/**
+ * The type Random load balance.
```
l81893521, line 25: Please change the note.

common/src/main/java/io/seata/common/rpc/RpcStatus.java
ph3636: This PR is incomplete. Please refer to the details

core/src/main/java/io/seata/core/rpc/netty/AbstractNettyRemoting.java
```diff
@@ -209,6 +219,9 @@ protected void sendAsync(Channel channel, RpcMessage rpcMessage) {
             LOGGER.debug("write message:" + rpcMessage.getBody() + ", channel:" + channel + ",active?"
                 + channel.isActive() + ",writable?" + channel.isWritable() + ",isopen?" + channel.isOpen());
         }
+
+        doBeforeRpcHooks(ChannelUtil.getAddressFromChannel(channel), rpcMessage);
```
slievrly, line 223: Where is the corresponding `doAfterRpcHooks ` ?

ph3636: There is no way to locate the result of the asynchronous message, so it is not added. Of course, only the send is added. It is the same for each address and does not affect the result.

PR #2680, opened 2020-05-12
optimize : Change GlobalTransactionalInterceptor to singleton

ph3636:


spring/src/main/java/io/seata/spring/annotation/GlobalTransactionScanner.java
```diff
@@ -150,6 +151,8 @@ public GlobalTransactionScanner(String applicationId, String txServiceGroup, int
         this.txServiceGroup = txServiceGroup;
         this.mode = mode;
         this.failureHandlerHook = failureHandlerHook;
+        this.globalTransactionalInterceptor = new GlobalTransactionalInterceptor(this.failureHandlerHook);
```
a364176773, line 154: if there is no GlobalTransactional, would it be too wasteful?

mingyifei:
> LGTM



> LGTM

image

大佬你好，我想了解一下，这个改变 解决了什么问题

Issue #2705, opened 2020-05-20
AT 模式使用PGSQL数据库


YangQiang-ChengDu:
环境：
seata1.2、jackson 2.9.5、PG42.2.10
错误如下：
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: No serializer found for class org.postgresql.core.v3.replication.V3ReplicationProtocol and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: io.seata.rm.datasource.undo.BranchUndoLog["sqlUndoLogs"]->java.util.ArrayList[0]->io.seata.rm.datasource.undo.SQLUndoLog["beforeImage"]->io.seata.rm.datasource.sql.struct.TableRecords["rows"]->java.util.ArrayList[0]->io.seata.rm.datasource.sql.struct.Row["fields"]->java.util.ArrayList[117]->io.seata.rm.datasource.sql.struct.Field["value"]->org.postgresql.jdbc.PgArray["resultSet"]->org.postgresql.jdbc.PgResultSet["statement"]->org.postgresql.jdbc.PgStatement["connection"]->org.postgresql.jdbc.PgConnection["queryExecutor"]->org.postgresql.core.v3.QueryExecutorImpl["replicationProtocol"])

Garlichead:
栈堆还可以更多点吗

jsbxyyx:
@YangQiang-ChengDu please verify 2708 pr.

PR #2708, opened 2020-05-21
bugfix: support jdbc type array, datalink etc.

jsbxyyx:
support jdbc type array, datalink, java_object, ref.

fixes: #1139 #2705


rm-datasource/src/main/java/io/seata/rm/datasource/sql/serial/SerialArray.java
```diff
@@ -0,0 +1,190 @@
...
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * used for jdbc type is JDBCType.ARRAY serialize.
+ *
+ * @author jsbxyyx
+ */
+public class SerialArray implements java.sql.Array, java.io.Serializable {
```
l81893521, line 37: Why not use javax.sql.rowset.serial.SerialArray?

jsbxyyx: use javax.sql.rowset.serial.SerialArray to serializer framework may be throws exception.
details see: getResultSet method, no-constructor method.

PR #2723, opened 2020-05-25
optimize: optimize get table meta in postgre sql

l81893521:
optimize get table meta in postgre sql


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/PostgresqlTableMetaCache.java
```diff
@@ -64,7 +62,7 @@ protected String getCacheKey(Connection connection, String tableName, String res
 
     @Override
     protected TableMeta fetchSchema(Connection connection, String tableName) throws SQLException {
-        try (Statement stmt = connection.createStatement()) {
+        try {
```
jsbxyyx, line 65: need close statement?

PR #2730, opened 2020-05-25
optimize: optimize get config type from configuration

### LINKED ISSUE #2724: get configTypeName a little flaw

### Ⅰ. Issue Description
when building a configuration, there are insufficient prompts for configuration items that do not exist or are not supported
当配置不存在或者不支持时,无法从异常提示中捕捉到对应的问题

question 1: not support register type should change to not support config type
问题1:not support register type: 应该改为not support config type:

question 2: the prompt that the configuration type that cannot be read and the configuration type is empty should be separated
问题2:应该将不支持配置中心类型,跟读不到配置中心类型的异常提示区分开

有兴趣处理的同学,可以回复你要认领即可
if you are interested in dealing with this problem, you can reply, please assign it to me

image

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

PeineLiang:
fix: get configTypeName exception

fixes #2724

### DIFF
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigType.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigType.java
@@ -66,6 +66,6 @@ public enum ConfigType {
                 return configType;
             }
         }
-        throw new IllegalArgumentException("illegal type:" + name);
+        throw new IllegalArgumentException("not support config type: " + name);
     }
 }
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
@@ -104,7 +104,7 @@ public final class ConfigurationFactory {
                     + ConfigurationKeys.FILE_ROOT_TYPE);
 
             if (StringUtils.isBlank(configTypeName)) {
-                throw new NotSupportYetException("can not get config type: " + configTypeName);
+                throw new NotSupportYetException("config type can not be null");
             }
 
             configType = ConfigType.getType(configTypeName);
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import io.seata.common.exception.NotSupportYetException;
 import io.seata.common.loader.EnhancedServiceLoader;
 import io.seata.common.loader.EnhancedServiceNotFoundException;
+import io.seata.common.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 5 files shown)


config/seata-config-core/src/main/java/io/seata/config/ConfigType.java
```diff
@@ -66,6 +66,6 @@ public static ConfigType getType(String name) {
                 return configType;
             }
         }
-        throw new IllegalArgumentException("illegal type:" + name);
+        throw new IllegalArgumentException("not support config type: " + name);
```
l81893521, line 69: How about do not change here?

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -101,9 +102,14 @@ private static Configuration buildConfiguration() {
             configTypeName = CURRENT_FILE_INSTANCE.getConfig(
                 ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
                     + ConfigurationKeys.FILE_ROOT_TYPE);
+
+            if (StringUtils.isEmpty(configTypeName)) {
```
l81893521, line 106:
```
if (configTypeName == null)
```
That s enough.

l81893521: Because if
```
configTypeName.equals("")
```
ConfigType.getType would throw exception.

config/seata-config-core/src/main/java/io/seata/config/ConfigType.java
PeineLiang: Because this is the getType method of the Enum class, the reason for the exception here is that the input parameter is illegal.

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -101,9 +102,14 @@ private static Configuration buildConfiguration() {
...
                     + ConfigurationKeys.FILE_ROOT_TYPE);
+
+            if (StringUtils.isBlank(configTypeName)) {
+                throw new NotSupportYetException("can not get config type: " + configTypeName);
+            }
+
             configType = ConfigType.getType(configTypeName);
         } catch (Exception e) {
-            throw new NotSupportYetException("not support register type: " + configTypeName, e);
+            throw e;
```
l81893521, line 112: If throw it directly, is it would throw other exception?

config/seata-config-core/src/main/java/io/seata/config/ConfigurationFactory.java
```diff
@@ -101,9 +102,14 @@ private static Configuration buildConfiguration() {
             configTypeName = CURRENT_FILE_INSTANCE.getConfig(
                 ConfigurationKeys.FILE_ROOT_CONFIG + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
                     + ConfigurationKeys.FILE_ROOT_TYPE);
+
+            if (StringUtils.isBlank(configTypeName)) {
+                throw new NotSupportYetException("can not get config type: " + configTypeName);
```
slievrly, line 107: configTypeName is blank，why not print it out explicitly？

PeineLiang: you are right ,i change the message to "config type can not be null"

PR #2731, opened 2020-05-25
optimize: consul supports ACL Token

jiehai:


a364176773:
please don't destroy the original format

jiehai:
consul supports ACL Token

Issue #2716, opened 2020-05-22
duplicate items when registering the branch


HanKaiqiang:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
lockKeys have the duplicate items, like this:
_lockKeys = ci_73_978_njrs:307;ci_73_978_njrs:307,308;ci_73_978_njrs:307,308,309,310,311;ci_73_978_njrs:307,308,309,310;ci_73_978_njrs:307,308,309,310,311,312;ci_73_978_njrs:307,308,309,310,311,312,313,314;ci_73_978_njrs:307,308,309,310,311,312,313;ci_73_799_pjct:14;trigger_tasks_73:167470;ci_73_978_njrs:307,308,309;trigger_task_change_logs_73:150222_

### Ⅱ. Describe what happened
when the @Transactional is under the @GlocalTransactional, and the method is called recursively in many hierarchies

### Ⅲ. Describe what you expected to happen
lockKeys should be:
_lockKeys = ci_73_978_njrs:309;ci_73_978_njrs:307;ci_73_978_njrs:308;ci_73_978_njrs:314;ci_73_978_njrs:312;trigger_tasks_73:167471;ci_73_978_njrs:313;ci_73_978_njrs:310;ci_73_799_pjct:14;ci_73_978_njrs:311;trigger_task_change_logs_73:150223_

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

slievrly:
@HanKaiqiang what version is used? This is a bug in the lower version.

HanKaiqiang:
> @HanKaiqiang使用什么版本？这是较低版本的错误。

mybatis-spring-boot-starter 2.0.1

HanKaiqiang:
mybatis-spring-boot-starter: 2.1.0

HanKaiqiang:
mybatis:
    mapperLocations: classpath:mapper/*.xml
    typeAliasesPackage: io.seata.sample.entity
    configuration-properties:
        blobType: BLOB
        boolValue: true
        prefix:
    configuration:
        cache-enabled: false
        lazy-loading-enabled: false
        multiple-result-sets-enabled: true
        use-column-label: true
        default-statement-timeout: 25000
        use-generated-keys: true
        **default-executor-type: reuse**
[...]

PR #2745, opened 2020-05-28
remove Duplicate lockKey

HanKaiqiang:
 remove Duplicate lockKey

#2716

mybatis
configuration:
default-executor-type: reuse

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/BaseTransactionalExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/BaseTransactionalExecutor.java
@@ -35,6 +35,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.StringJoiner;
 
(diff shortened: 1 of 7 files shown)


rm-datasource/src/main/java/io/seata/rm/datasource/exec/BaseTransactionalExecutor.java
```diff
@@ -246,8 +244,30 @@ protected void prepareUndoLog(TableRecords beforeImage, TableRecords afterImage)
         }
 
         ConnectionProxy connectionProxy = statementProxy.getConnectionProxy();
-
         TableRecords lockKeyRecords = sqlRecognizer.getSQLType() == SQLType.DELETE ? beforeImage : afterImage;
+
```
l81893521, line 248: This method is use for prepare undo log, so you put the remove duplicate lock key, that s look like unreasonable.

rm-datasource/src/main/java/io/seata/rm/datasource/exec/BaseTransactionalExecutor.java
```diff
@@ -246,8 +244,30 @@ protected void prepareUndoLog(TableRecords beforeImage, TableRecords afterImage)
...
+                            }
+                        }
+                    }
+                }
+            }
+            lockKeyRecords.setRows(rows);
+        }
+
         String lockKeys = buildLockKey(lockKeyRecords);
         connectionProxy.appendLockKey(lockKeys);
```
l81893521, line 272: Do you have any good idea when append the lock key, it will remove duplicated lock key automatically?

PR #2777, opened 2020-06-08
bugfix: fix can't rollback when set rollback retry count was zero.

l81893521:
fix could not rollback when set rollback retry count was zero.


tm/src/main/java/io/seata/tm/api/DefaultGlobalTransaction.java
```diff
@@ -120,7 +120,7 @@ public void commit() throws TransactionException {
             return;
         }
         assertXIDNotNull();
-        int retry = COMMIT_RETRY_COUNT;
+        int retry = COMMIT_RETRY_COUNT <= 0 ? 1 : COMMIT_RETRY_COUNT;;
```
slievrly, line 123: Is it better to use the default value when <=0?

l81893521: fixed.

PR #2782, opened 2020-06-09
Reduce hashmap one expansion

xianlaioy:
0.75 is the default load factor,if the maps size equals length need divided by 0.75.

### DIFF
--- a/core/src/main/java/io/seata/core/model/BranchStatus.java
+++ b/core/src/main/java/io/seata/core/model/BranchStatus.java
@@ -15,7 +15,7 @@
  */
 package io.seata.core.model;
 
-import com.google.common.collect.Maps;
+import java.util.HashMap;
 import java.util.Map;
 
 import io.seata.common.exception.ShouldNeverHappenException;
@@ -108,7 +108,7 @@ public enum BranchStatus {
         return code;
     }
 
-    private static final Map<Integer, BranchStatus> MAP = Maps.newHashMapWithExpectedSize(values().length);
+    private static final Map<Integer, BranchStatus> MAP = new HashMap<>((int) ((float) values().length / 0.75F + 1.0F));
 
     static {
         for (BranchStatus status : values()) {
--- a/core/src/main/java/io/seata/core/model/BranchStatus.java
+++ b/core/src/main/java/io/seata/core/model/BranchStatus.java
@@ -15,7 +15,7 @@
  */
 package io.seata.core.model;
 
-import java.util.HashMap;
+import com.google.common.collect.Maps;
 import java.util.Map;
 
 import io.seata.common.exception.ShouldNeverHappenException;
@@ -108,7 +108,7 @@ public enum BranchStatus {
         return code;
     }
 
-    private static final Map<Integer, BranchStatus> MAP = new HashMap<>(values().length);
+    private static final Map<Integer, BranchStatus> MAP = Maps.newHashMapWithExpectedSize(values().length);
 
     static {
         for (BranchStatus status : values()) {


core/src/main/java/io/seata/core/model/BranchStatus.java
```diff
@@ -108,7 +108,7 @@ public int getCode() {
         return code;
     }
 
-    private static final Map<Integer, BranchStatus> MAP = new HashMap<>(values().length);
+    private static final Map<Integer, BranchStatus> MAP = Maps.newHashMapWithExpectedSize(values().length);
```
l81893521, line 111: I think that s same and branch status was fixed, it would not let the map resize.

xianlaioy: This way of writing is bound to expand capacity

l81893521: You re right, thanks.

xianlaioy:
trigger a new build

l81893521:
> trigger a new build

check here:
https://travis-ci.com/github/seata/seata/jobs/346914071

xianlaioy:
> > trigger a new build
> 
> check here:
> https://travis-ci.com/github/seata/seata/jobs/346914071

seata-serializer-seata 1.3.0-SNAPSHOT .............. FAILURE

Strangely, I didn't change the code for this module

l81893521:
> > > trigger a new build
> > 
> > 
> > check here:
> > https://travis-ci.com/github/seata/seata/jobs/346914071
> 
> seata-serializer-seata 1.3.0-SNAPSHOT .............. FAILURE
> 
> Strangely, I didn't change the code for this module

maybe you can check the dependency of guava in core/pom.xml

xianlaioy:
> > > > trigger a new build
> > > 
> > > 
> > > check here:
> > > https://travis-ci.com/github/seata/seata/jobs/346914071
> > 
> > 
> > seata-serializer-seata 1.3.0-SNAPSHOT .............. FAILURE
> > Strangely, I didn't change the code for this module
> 
> maybe you can check the dependency of guava in core/pom.xml

tks,

core/src/main/java/io/seata/core/model/BranchStatus.java
```diff
@@ -108,7 +108,7 @@ public int getCode() {
         return code;
     }
 
-    private static final Map<Integer, BranchStatus> MAP = new HashMap<>(values().length);
+    private static final Map<Integer, BranchStatus> MAP = new HashMap<>((int) ((float) values().length / 0.75F + 1.0F));
```
slievrly, line 111: The capacity of the map is always 2 ^ n.

xianlaioy: tks，min value is 2^n。So 10 won't expand, if it's 12 it will.

xianlaioy: `(int) ((float) values().length / 0.75F + 1.0F)  `
It should be written in a more correct way

slievrly: I don't recommend that. 0.75 is a hashMap default factor, should not directly write this value here,  if 0.75 becomes a different value in a later JDK version, what shall we do?

xianlaioy: It's supposed to be a constant, in theory it doesn't change。JDK Hashmap Source Code is optimized all the time, but this load factor has not changed

xianlaioy: We encapsulate a method to handle internal logic, and default load factors can also be defined as variables. Forget it. I close it.

PR #2783, opened 2020-06-09
optimize: optimize trace logger

l81893521:
optimize the trace logger.

### DIFF
--- a/core/src/main/java/io/seata/core/logger/StackTraceLogger.java
+++ b/core/src/main/java/io/seata/core/logger/StackTraceLogger.java
@@ -19,8 +19,6 @@ import io.seata.config.Configuration;
 import io.seata.config.ConfigurationFactory;
 import io.seata.core.constants.ConfigurationKeys;
 import org.slf4j.Logger;
-
-import java.util.Date;
 import java.util.concurrent.ThreadLocalRandom;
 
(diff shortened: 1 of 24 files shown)


core/src/main/java/io/seata/core/logger/StackTraceLogger.java
```diff
@@ -30,10 +32,12 @@
 
     private static final String STACK_TRACE_LOGGER_PREFIX = "[stacktrace]";
 
+    private static final int PERCENT = 100;
+
     public static void info(Logger logger, Throwable cause, String format, Object[] args) {
         if (logger.isInfoEnabled()) {
             int rate = getRate();
-            if (System.currentTimeMillis() % rate == 0) {
+            if (ThreadLocalRandom.current().nextInt(PERCENT / rate) == 0) {
```
slievrly, line 40: How to understand rate? If I set the rate to 1000？

l81893521: Look like I misunderstand the mean of rate in the origin logic. I will change it.

l81893521: Fixed. But now the meaning of 
```
client.log.exception
```
was changed.
If set the rate to 1000, It means "show 1 log per 1000 times"

l81893521:
The code already change, so close.

PR #2846, opened 2020-06-29
optimize: Optimize LogStoreDataBaseDAO, and added count method to LogStore

wangliang1986:
1. 优化LogStoreDataBaseDAO，可通过condition中的多个查询条件一起筛选数据。
2. 添加查询数量的方法到LogStore接口中。

从PR #2828 拆分出来的。

### DIFF
--- a/core/src/main/java/io/seata/core/store/GlobalTransactionDO.java
+++ b/core/src/main/java/io/seata/core/store/GlobalTransactionDO.java
@@ -28,9 +28,9 @@ public class GlobalTransactionDO {
 
     private String xid;
 
-    private Long transactionId;
+    private long transactionId;
 
-    private Integer status;
+    private int status;
 
     private String applicationId;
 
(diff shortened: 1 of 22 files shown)


core/src/main/java/io/seata/core/store/GlobalTransactionDO.java
```diff
@@ -24,23 +25,23 @@
  *
  * @author zhangsen
  */
-public class GlobalTransactionDO {
+public class GlobalTransactionDO implements GlobalTransactionModel {
 
     private String xid;
 
-    private Long transactionId;
+    private long transactionId;
```
lj2018110133, line 32: why changed to basic data types? the transactionId may be null

wangliang1986: the transactionId will not be null unless manually modified to null

PR #2887, opened 2020-07-14
bugfix: fix rm client receive response logic.

objcoding:
The previous logic client side did not have onResponse logic. The TransactionMessageHandler interface has an implementation class AbstractRMHandler. The onRequest method is specifically used to handle rm’s commit, rollback, and deleteUndoLog. After putting these logics in the processor, commit, rollback, and deleteUndoLog are all independent. The respective processor, the logic of onResponse is placed in ClientOnResponseProcessor, so the relevant logic of the client response is placed here, unified and centralized processing

### DIFF
--- a/rm/src/main/java/io/seata/rm/AbstractRMHandler.java
+++ b/rm/src/main/java/io/seata/rm/AbstractRMHandler.java
@@ -152,7 +152,7 @@ public abstract class AbstractRMHandler extends AbstractExceptionHandler
 
     @Override
     public void onResponse(AbstractResultMessage response, RpcContext context) {
-        LOGGER.info("the rm client received response msg from tc server, bug without any processing logic.");
+        LOGGER.info("the rm client received response msg [{}] from tc server.", response.toString());
     }
 
     public abstract BranchType getBranchType();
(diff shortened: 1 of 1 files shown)


slievrly:
image

rm/src/main/java/io/seata/rm/AbstractRMHandler.java
```diff
@@ -153,7 +152,6 @@ public AbstractResultMessage onRequest(AbstractMessage request, RpcContext conte
 
     @Override
     public void onResponse(AbstractResultMessage response, RpcContext context) {
-        throw new ShouldNeverHappenException();
```
slievrly, line 156: add receive response log.

objcoding: done.

rm/src/main/java/io/seata/rm/AbstractRMHandler.java
```diff
@@ -153,7 +152,7 @@ public AbstractResultMessage onRequest(AbstractMessage request, RpcContext conte
...
+        LOGGER.warn("client received response msg [{}] from tc server, bug without any processing logic.", response.getMsg());
```
slievrly: if it's normal，response.getMsg() is null.

objcoding: ok, done

PR #2914, opened 2020-07-21
bugfix: fix branchType not cleaned when consumer is in TCC mode

wangliang1986:
bugfix: fix branchType not cleaned when consumer is in TCC mode.

added some test.


core/src/test/java/io/seata/core/context/RootContextTest.java
```diff
@@ -57,6 +57,45 @@ public void testGetXID() {
...
+        assertThat(RootContext.getBranchType()).isNull();
+        //after bind xid, branchType is not null
+        RootContext.bind(DEFAULT_XID);
+        assertThat(RootContext.getBranchType()).isEqualTo(DEFAULT_BRANCH_TYPE.name());
+
+        RootContext.unbind();
+        assertThat(RootContext.unbindBranchType()).isEqualTo(DEFAULT_BRANCH_TYPE.name());
+        assertThat(RootContext.getBranchType()).isNull();
+    }
+
```
CharmingRabbit, line 98: I think we should merge these two cases into the existing method testBindBranchType_And_UnbindBranchType().

wangliang1986: done

PR #2935, opened 2020-07-27
bugfix：fix saga designer bug that the property box does not switch when switching nodes

mxszs:

1. bugfix：fix edit node details
When one task is selected and then another task is selected, the property box on the right remains unchanged, which will cause the properties of the two tasks to be modified at the same time

2. solve the same problem in some cases


a364176773:
please modify the title  ，such as bugfix: fix saga designer

slievrly (approved):
LGTM

PR #2936, opened 2020-07-28
[close] [WIP:再拆分一些模块] refactor: Some classes moved from seata-server to seata-tc

wangliang181230:
refactor: Some classes moved from seata-server to seata-tc
重构：部分类从seata-server转移到新建的model seata-tc中，为直连模式的PR #2937 做准备。


wangliang181230:
> 
> 
> What exactly do TC and Server mean? I think we can split out the functional modules like storage and lock.

The direct connection mode needs to use the functions of `LogStore` and `LockStore`

PR #2987, opened 2020-08-06
bugfix: fix repeat execute sql

hecyxy:
when execute retry policy, it will execute sql repeatly. it will cause dirty data. we just need to retry the commit action.

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
@@ -107,8 +107,8 @@ public abstract class AbstractDMLBaseExecutor<T, S extends Statement> extends Ba
         ConnectionProxy connectionProxy = statementProxy.getConnectionProxy();
         try {
             connectionProxy.setAutoCommit(false);
+            final T result = executeAutoCommitFalse(args);
             return new LockRetryPolicy(connectionProxy).execute(() -> {
-                T result = executeAutoCommitFalse(args);
                 connectionProxy.commit();
                 return result;
             });


ggndnn:
@hecyxy Thanks for your contribution. There is a retry policy that branch will rollback on conflict, so sql statement should be executed each time in this case. Please refer to configuration `client.rm.lock.retryPolicyBranchRollbackOnConflict` (https://seata.io/zh-cn/docs/user/configurations.html).

hecyxy:
> https://seata.io/zh-cn/docs/user/configurations.html

thanks for your replay. i will give you a example. In our application, when i open the retryPolicyBranchRollbackOnConflict , it will execute  ' executeAutoCommitFalse(args) ' and  'connectionProxy.commit()' repeatly if there is a lock conflict int current status , then it will cause to executeAutoCommitFalse repeatly, if it doesn't include primary key or unique key,,the lock disappers,it may insert the same record serval times.then in the next.

1. insert without primary key or unique key
2. commit -> lock conflict occurs
3. insert again 
3. commit ->  lock conflict occurs
4. insert again 
5. commit successfully (lock disappear)
6. cause dirty data

ggndnn:
> > https://seata.io/zh-cn/docs/user/configurations.html
> 
> thanks for your reply. i will give you a example. In our application, when i open the retryPolicyBranchRollbackOnConflict , it will execute ' executeAutoCommitFalse(args) ' and 'connectionProxy.commit()' repeatly if there is a lock conflict int current status , then it will cause to executeAutoCommitFalse repeatly, if it doesn't include primary key or unique key,the lock disappers,it may insert the same record serval times.
> the following 'insert' is the same sql, user only wants to execute once.
> 
> 1. insert (without primary key or unique key)
> 2. commit -> lock conflict occurs
> 3. insert again
> 4. commit ->  lock conflict occurs
> 5. insert again
> 6. commit successfully (lock disappear)
> 7. cause dirty data

My suggestion is to close this pr which is not a property solution, and open an issue describing the environment that you used, such as seata version, db type, isolation level, your use case, some example code and so on. I think the primary key is mandatory as global lock key, that I am not sure how your situation happened. Please try the latest version seata first.

hecyxy:
> > > https://seata.io/zh-cn/docs/user/configurations.html
> > 
> > 
> > thanks for your reply. i will give you a example. In our application, when i open the retryPolicyBranchRollbackOnConflict , it will execute ' executeAutoCommitFalse(args) ' and 'connectionProxy.commit()' repeatly if there is a lock conflict int current status , then it will cause to executeAutoCommitFalse repeatly, if it doesn't include primary key or unique key,the lock disappers,it may insert the same record serval times.
> > the following 'insert' is the same sql, user only wants to execute once.
> > 
> > 1. insert (without primary key or unique key)
> > 2. commit -> lock conflict occurs
> > 3. insert again
> > 4. commit ->  lock conflict occurs
> > 5. insert again
> > 6. commit successfully (lock disappear)
> > 7. cause dirty data
> 
> My suggestion is to close this pr which is not a property solution, and open an issue describing the environment that you used, such as seata version, db type, isolation level, your use case, some example code and so on. I think the primary key is mandatory as global lock key, that I am not sure how your situation happened. Please try the latest version seata first.



> > > https://seata.io/zh-cn/docs/user/configurations.html
> > 
> > 
> > thanks for your reply. i will give you a example. In our application, when i open the retryPolicyBranchRollbackOnConflict , it will execute ' executeAutoCommitFalse(args) ' and 'connectionProxy.commit()' repeatly if there is a lock conflict int current status , then it will cause to executeAutoCommitFalse repeatly, if it doesn't include primary key or unique key,the lock disappers,it may insert the same record serval times.
> > the following 'insert' is the same sql, user only wants to execute once.
> > 
> > 1. insert (without primary key or unique key)
> > 2. commit -> lock conflict occurs
> > 3. insert again
> > 4. commit ->  lock conflict occurs
> > 5. insert again
> > 6. commit successfully (lock disappear)
> > 7. cause dirty data
> 
> My suggestion is to close this pr which is not a property solution, and open an issue describing the environment that you used, such as seata version, db type, isolation level, your use case, some example code and so on. I think the primary key is mandatory as global lock key, that I am not sure how your situation happened. Please try the latest version seata first.

OK

PR #2964, opened 2020-08-03
optimize: make configuration automation and rationalization

wangliang181230:
### Ⅰ. Describe what this PR did
optimize: make configuration automation and rationalization
优化：使配置自动化、合理化。

### 当前PR所调整的内容：
1. 自动生成配置默认值：
1.1. `seata.client.rm.saga-json-parser`，自动判断依赖生成相应的配置默认值，判断顺序：`fastjson(原默认值)`、`jackson`
1.2. `seata.client.undo.log-serialization`，自动判断依赖生成相应的配置默认值，判断顺序：`jackson(原默认值)`、`fastjson`、`kryo`、`protostuff`.
2. 合理化的配置：
2.1. 修改了`eureka`和`sofa`的`application`配置默认值: `seata-server`。原来的注册应用名为`default`，让别人看到会不明白是什么应用。
2.2. 所有的`localhost`改为`127.0.0.1`
3. 顺便添加了一些测试用例。

PR #3146, opened 2020-09-24
optimize: change exceptionRate in PROPERTIES and YML files to exception-rate

lizhuoxuan:
Because the naming style of PROPERTIES and YML files was a combination of dot + underscore, I thought 'exceptionRate' should be changed to 'exception-rate' would be more appropriate.


wangliang181230:
#2964 This PR has fixed this problem

lizhuoxuan:
> #2964 This PR has fixed this problem

In script/client/spring/application.yml and script/client/spring/application.properties,it's still 'exceptionRate'.I didn't solve any problem, I just changed the format. :smiley:

lizhuoxuan:
> #2964 This PR has fixed this problem

Copy that.:smile:

PR #3170, opened 2020-10-06
bugfix: the disposables tree set won't accept another Disposable with the same priority.

jujinghao:
Fix the following issue:
Disposable TmNettyRemotingClient and RmNettyRemotingClient added by GlobalTransactionScanner have the same priority by default, so the RmNettyRemotingClient won't be closed gracefully during shutdown.


slievrly (approved):
LGTM, great found.

core/src/main/java/io/seata/core/rpc/ShutdownHook.java
```diff
@@ -91,35 +91,63 @@ public static void removeRuntimeShutdownHook() {
...
+
         public DisposablePriorityWrapper(Disposable disposable, int priority) {
             this.disposable = disposable;
             this.priority = priority;
+            this.seqId = seq.incrementAndGet();
         }
 
         @Override
         public int compareTo(DisposablePriorityWrapper disposablePriorityWrapper) {
-            return priority - disposablePriorityWrapper.priority;
```
selfishlover, line 105: use PriorityQueue to replace TreeSet will solve this problem once and forever. No need to add a "seqId" to fix the probem which is actually caused by TreeSet.

core/src/main/java/io/seata/core/rpc/ShutdownHook.java
```diff
@@ -91,35 +91,63 @@ public static void removeRuntimeShutdownHook() {
...
+            int cmp = priority - disposablePriorityWrapper.priority;
+            if (cmp == 0) {
+                if (seqId > disposablePriorityWrapper.seqId) {
+                    cmp = 1;
+                } else if (seqId < disposablePriorityWrapper.seqId) {
+                    cmp = -1;
+                } else {
+                    cmp = 0;
+                }
+            }
```
jsbxyyx, line 119: Long.compare(x, y) ?

core/src/main/java/io/seata/core/rpc/ShutdownHook.java
```diff
@@ -91,35 +91,63 @@ public static void removeRuntimeShutdownHook() {
...
         }
 
         @Override
         public int hashCode() {
-            return Objects.hashCode(this.priority);
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + priority;
+            result = prime * result + (int) (seqId ^ (seqId >>> 32));
+            return result;
```
jsbxyyx, line 129: Objects.hash(...values)?

l81893521 (approved):
LGTM

core/src/main/java/io/seata/core/rpc/ShutdownHook.java
a364176773: +1

core/src/main/java/io/seata/core/rpc/ShutdownHook.java
a364176773: +1

a364176773 (approved):
LGTM

PR #3175, opened 2020-10-10
optimize: improve UUIDGenerator using "history time" version of snowflake algorithm

### LINKED ISSUE #3126: seata更改服务器时间后增删改报错回滚
测试环境下经常会更改时间做测试，当更改时间后再将服务器时间更改回来，在seata 1.3.0 AT 模式下会出现报错回滚问题，server的错误提示就是超时回滚

selfishlover:
Improve UUIDGenerator base on an improved snowflake algorithm which use "history time" instead of current timestamp.
[algorithm doc](https://www.yuque.com/simonalong/butterfly/mn26oy)
So that seata-server can support system time move backward during run time.
Beside, implement is base on CAS instead of Lock, so it would be faster~

fixes #3126 

run seata-server in a server, then change machine's system time to a previous time, if seata-server can still generate normal UUID, then it proved worked.


common/src/test/java/io/seata/common/util/IdWorkerTest.java
```diff
@@ -0,0 +1,45 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package io.seata.common.util;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
selfishlover: idea auto merge these in one *

a364176773 (approved):
LGTM

common/src/main/java/io/seata/common/util/IdWorker.java
```diff
@@ -14,161 +14,173 @@
...
+     * block current thread if the QPS of acquiring UUID is too high
+     * that current sequence space is exhausted
      */
-    protected long tilNextMillis(long lastTimestamp) {
-        long timestamp = timeGen();
-        while (timestamp <= lastTimestamp) {
-            timestamp = timeGen();
+    private void waitIfNecessary() {
+        long currentWithSequence = timestampAndSequence.get();
+        long current = currentWithSequence >>> sequenceBits;
```
l81893521, line 129: if currentWithSequence will never be negative number, I think user
```
>>
```
is enought.

selfishlover: currentWithSequence could be a negative number(though it would be a long long time to reach this~). What ever, >>> and >> have the same speed for cpu.

common/src/main/java/io/seata/common/util/IdWorker.java
```diff
@@ -14,161 +14,174 @@
...
+        return System.currentTimeMillis() - twepoch;
     }
 
-    public static long initWorkerId() {
-        InetAddress address;
+    /**
+     * auto generate workerId, try using mac first, if failed, then randomly generate one
+     * @return workerId
+     */
+    private long generateWorkerId() {
```
l81893521, line 152: I think origin generate worker id is better than this.

l81893521 (approved):
LGTM

PR #3201, opened 2020-10-16
optimize: no longer substring the message in rpc transmission

l81893521:
User Feedback: Misunderstand the message when transaction error occurred in TC. Because if the message longer than 128 byte, and TC would substring the message before rpc transmission.

no longer substring the message in rpc transmission and show the hold message especially in transaction error.


wangliang181230 (approved):
LGTM

lightClouds917 (approved):
LGTM

jsbxyyx:
We recommend keeping the original situation. for example: adding `go to seata-server view log` in message.

a364176773 (approved):
LGTM

serializer/seata-serializer-seata/src/main/java/io/seata/serializer/seata/protocol/AbstractResultMessageCodec.java
```diff
@@ -45,13 +43,7 @@
...
             if (StringUtils.isNotEmpty(resultMsg)) {
-                String msg;
-                if (resultMsg.length() > MAX_ERR_MSG_LEN) {
-                    msg = resultMsg.substring(0, MAX_ERR_MSG_LEN) + "...";
-                } else {
-                    msg = resultMsg;
-                }
-                byte[] bs = msg.getBytes(UTF8);
+                byte[] bs = resultMsg.getBytes(UTF8);
                 out.writeShort((short)bs.length);
```
slievrly, line 47: short  should need to be less  2^15-1

l81893521: fixed.

slievrly (approved):
LGTM

PR #3243, opened 2020-10-29
Solve the problem of context passing during asynchronous execution. 

### LINKED ISSUE #166: is support "A call B and C, call B by sync, but call C by async"
## Why you need it?
A call B and C, call B by sync, but call C by async

## How it could be?
Thread Context

about-cloud:
Solve the problem of context passing during asynchronous execution.

fixes #166

Added, detailed PR code.

1. Unit testing of new code；
2. Start A, B, C three services, A calls B, A calls C asynchronously. (At least one service is called asynchronously)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -125,6 +125,7 @@
         <hessian.version>4.0.63</hessian.version>
         <fst.version>2.57</fst.version>
         <groovy.version>2.4.4</groovy.version>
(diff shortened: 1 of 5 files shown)


a364176773:
I think it has been implemented in #2675

a364176773:
Thank you for your contribute

PR #3129, opened 2020-09-18
bugfix: forbidding execute SQL which update pk value

81519434:
### Ⅰ. Describe what this PR did
Forbidding execute SQL which updates pk value.

### Ⅱ. Does this pull request fix one issue?
#3066

PR #3287, opened 2020-11-16
bugfix:throw exception when update pk 

81519434:
optimize this pr. #3129


rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
```diff
@@ -35,10 +24,19 @@
 import io.seata.rm.datasource.sql.struct.TableRecords;
 import io.seata.sqlparser.SQLRecognizer;
 import io.seata.sqlparser.util.JdbcConstants;
-import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.SQLException;
```
a364176773, line 30: java packages should be at the top

a364176773 (changes requested):
please use p3C to format the code

caohdgege:
关联一下#3289

rm-datasource/src/main/java/io/seata/rm/datasource/exec/AbstractDMLBaseExecutor.java
81519434: done

xingfudeshi:
This problem is that the primary key should not be changed. If the new primary key value is consistent with the old primary key value, can it be ignored directly?@81519434

81519434:
> This problem is that the primary key should not be changed. If the new primary key value is consistent with the old primary key value, can it be ignored directly?@81519434

Yes, this pr is doing exactly what you said!

xingfudeshi (approved):
LGTM.

slievrly (approved):
LGTM

a364176773 (approved):
LGTM

PR #3411, opened 2020-12-25
feature: add config support for seata server thread pool

lvxianzheng:
新功能:增加seata server thread pool 可配置参数功能

增加参数配置，可以性能调优,tcc 模式下，当客户端并发超过50的时候，服务器和客户端都会Hang住，可以扩大此参数暂时解决问题，这是因为2阶段同步提交引起的，改为异步提交可以修复此问题。

集成测试才可以测试

在file.conf中修改配置，运行性能测试，观察线程数量变化

无


a364176773:
标题请用英文，可参考其他pr

server/src/main/resources/file.conf
```diff
@@ -1,4 +1,15 @@
 ## transaction log store, only used in seata-server
```
slievrly, line 1: move to line 13？

lvxianzheng: As these parameters are for server thread pool instead of store, so we'd better put them out of the store section. 
这些参数是设置线程池的，最好还是放到server的段落，或者删除这一段，采用默认设置也是可以的。

server/src/main/resources/file.conf
```diff
@@ -1,4 +1,15 @@
...
+server {
```
l81893521, line 2: I have some question about this pr.
1.The configuration ofNetty we already had the prefix name "transport.xxx.xxx", how about change to transport.xxx.xxx instead of server.xxx.xxx
2.How about move this configuration in the NettyServerConfig.java ?
3.We do not show the configuration of Netty in file.conf. How about remove it and add to
```
file.conf.example
script/server/file.conf
script/server/file.properties
script/server/file.yml
```

a364176773: +1

lvxianzheng: 1 and 2:
I think the thread pool configurations are related to the server itself instead of transport.
So the server.xxx.xxx is more better than transport.
If the configuration is set for client and server communication, it is better using transport.xxx.xxx

3. In fact, I can remove the default value in file.conf.
I put the default value into the file.conf, because it is the responsibility to reminder we can change the value for performance problems.

l81893521: Thanks for you replay. For now we already put the config of client in the NettyServerConfig and the server in the NettyServerConfig, and they begin with 'transport.xxx.xxx'. How do you think move the config of server to the NettyServerConfig and let it begin with 'transport.xxx.xxx'. May be you are right and I hope can keep in touch with after this. But for now I think the consistency is more important than the correctness.

l81893521 (approved):
LGTM

l81893521 (commented):
By the way. Can you add the change to the 1.5.0.md in the path /change and /change/en-us

server/src/main/resources/file.conf
lvxianzheng: Now I put the configure into NettyServerConfig and change the name from server.xxx.xxx to transport.xxx.xxx

slievrly (approved):
LGTM

PR #3427, opened 2020-12-31
feature: add distributed lock interface

a364176773:


common/src/main/java/io/seata/common/Constants.java
```diff
@@ -131,4 +131,30 @@
...
+
+    /**
+     * The constant TX_TIMEOUT_CHECK
+     */
+    String TX_TIMEOUT_CHECK = "TxTimeoutCheck";
+
+    /**
+     * The constant UNDOLOG_DELETE
+     */
+    String UNDOLOG_DELETE = "UndoLogDelete";
```
xingfudeshi, line 158: UndologDelete

caohdgege (approved):
LGTM

common/src/main/java/io/seata/common/Constants.java
a364176773: ok

slievrly (approved):
LGTM,  need to further define how to load different lock implementations

l81893521 (approved):
LGTM

PR #3501, opened 2021-01-28
optimize: check whether undolog table exist before cleaning undolog

selfishlover:
when rm receive UndoLogDeleteRequest(send by TC), if user does not use AT mode at all(which mean their database does not have undo_log table), exception would occur(such as undo_log table doesn't existed~) and logger will print error info(which scares user~). Here, we check whether undo_log table exist before trying to delete.

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
+++ b/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
@@ -19,12 +19,15 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import io.seata.core.model.BranchType;
 import io.seata.core.model.ResourceManager;
 import io.seata.core.protocol.transaction.UndoLogDeleteRequest;
 import io.seata.rm.datasource.DataSourceManager;
 import io.seata.rm.datasource.DataSourceProxy;
+import io.seata.rm.datasource.undo.UndoLogManager;
 import io.seata.rm.datasource.undo.UndoLogManagerFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


rm-datasource/src/test/java/io/seata/rm/RMHandlerATTest.java
```diff
@@ -0,0 +1,82 @@
...
+import io.seata.rm.datasource.DataSourceManager;
+import io.seata.rm.datasource.DataSourceProxy;
+import io.seata.rm.datasource.undo.UndoLogManager;
+import org.junit.jupiter.api.Test;
+
+import java.sql.Connection;
+
+import static org.mockito.Mockito.*;
+
+import static org.junit.jupiter.api.Assertions.*;
```
xingfudeshi, line 28: Please do not use *.

rm-datasource/src/test/java/io/seata/rm/RMHandlerATTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+class RMHandlerATTest {
```
xingfudeshi, line 30: Please write @author.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/oracle/OracleUndoLogManager.java
```diff
@@ -99,4 +100,8 @@ private void insertUndoLog(String xid, long branchID, String rollbackCtx, byte[]
         }
     }
 
+    @Override
+    protected String getCheckUndoLogTableExistSql() {
+        return CHECK_UNDO_LOG_TABLE_EXIST_SQL;
+    }
```
xingfudeshi, line 106: How about using the following SQL statement instead of throwing an exception to determine whether a table exists?
```
select count(*) from user_tables where table_name =upper('undo_log_table_name')
```

rm-datasource/src/main/java/io/seata/rm/datasource/undo/AbstractUndoLogManager.java
```diff
@@ -431,4 +433,19 @@ protected abstract void insertUndoLogWithNormal(String xid, long branchId, Strin
...
+    @Override
+    public boolean hasUndoLogTable(Connection conn) {
+        String checkExistSql = getCheckUndoLogTableExistSql();
+        try (PreparedStatement checkPst = conn.prepareStatement(checkExistSql)) {
+            checkPst.executeQuery();
+            return true;
+        } catch (SQLException e) {
+            return false;
+        }
+    }
```
xingfudeshi, line 446: How about using the following SQL statement instead of throwing an exception to determine whether a table exists?
```
SELECT * 
FROM information_schema.tables
WHERE table_schema = 'db_name' 
    AND table_name = 'table_name'
LIMIT 1
```

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -40,43 +43,93 @@
...
+
+    int deleteUndoLog(UndoLogManager manager, Connection conn, Date division) {
+        try {
+            int deleteRows = manager.deleteUndoLogByLogCreated(division, LIMIT_ROWS, conn);
+            if (!conn.getAutoCommit()) {
+                conn.commit();
+            }
+            return deleteRows;
+        } catch (SQLException e) {
+            LOGGER.error("Failed to delete expired undo_log", e);
```
xingfudeshi, line 124: LOGGER.error("Failed to delete expired undo_log", e.getMessage(),e);

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -40,43 +43,93 @@
...
+            try {
+                if (!conn.getAutoCommit()) {
+                    conn.rollback();
                 }
+            } catch (SQLException re) {
+                LOGGER.error("Failed to rollback undolog", re);
```
xingfudeshi, line 130: LOGGER.error("Failed to rollback undolog {}", re.getMessage(),re);

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -40,43 +43,93 @@
...
         DataSourceManager dataSourceManager = (DataSourceManager)getResourceManager();
-        DataSourceProxy dataSourceProxy = dataSourceManager.get(request.getResourceId());
+        DataSourceProxy dataSourceProxy = dataSourceManager.get(resourceId);
         if (dataSourceProxy == null) {
-            LOGGER.warn("Failed to get dataSourceProxy for delete undolog on {}", request.getResourceId());
+            LOGGER.warn("Failed to get dataSourceProxy for delete undolog on {}", resourceId);
+            return;
+        }
+
+        boolean hasUndoLogTable = undoLogTableExistRecord.computeIfAbsent(resourceId, id -> checkUndoLogTableExist(dataSourceProxy));
```
xingfudeshi, line 58: At the beginning, I didn't create this table, but after a while, I created this table. Is there any problem?

xingfudeshi (commented):
I left some comments.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/oracle/OracleUndoLogManager.java
selfishlover: select 1 from {targetTable} is the fastest way to figure out whether target table exists or not. Since we don't care what it "looks like", no need to query user_tables.

rm-datasource/src/main/java/io/seata/rm/datasource/undo/AbstractUndoLogManager.java
selfishlover: 同上

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
selfishlover: message will be printed automatically.

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
selfishlover: 同上

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
selfishlover: if users don't have undo_log table at first, it means they must not using AT mode in their code, otherwise things would go wrong, right? Later if they decide to pick up AT mode and create undo_log table, they would need to modify their code and rebuild -> redeploy -> ...

xingfudeshi (approved):
LGTM.

a364176773:
@objcoding PTAL

PR #3783, opened 2021-05-30
bugfix: fix the problem that 'store mode' parameter does not take effect

objcoding:

After the original store mode is split into session store mode and lock store mode, after old users upgrade to the new version, the store mode in the upgraded system will become invalid.
Solution:
1. Remove the default values of session store mode and lock store mode;
2. The store mode is reserved as a global setting.

This will make seata better compatible with the old version and more user-friendly upgrades.


a364176773 (approved):
LGTM

seata-spring-autoconfigure/seata-spring-autoconfigure-server/src/main/java/io/seata/spring/boot/autoconfigure/properties/server/store/StoreProperties.java
```diff
@@ -57,7 +57,7 @@ public StoreProperties setPublicKey(String publicKey) {
     @Component
     @ConfigurationProperties(prefix = STORE_SESSION_PREFIX)
     public static class Session {
-        private String mode = "file";
+        private String mode;
 
         public String getMode() {
             return mode;
```
slievrly, line 63: Pls print log warning the item has expired when calling store.mode, but the function is compatible with previous and available.

wangliang181230 (approved):
LGTM

PR #3817, opened 2021-06-09
bugfix: TC SkyWalking topo calling node not gather

zhaoyuguang:
fixed https://github.com/seata/seata/issues/3810


caohdgege:
image

It seems that the newest topo is not expect. the node seata_tc & 127.0.0.1:8091 is the same node.

zhaoyuguang:
> image
> 
> It seems the newest topo is not expect. the node seata_tc & 127.0.0.1:8091 is the same node.

If use such as Chrome, go to developer mode, copy HTTP Response, let us check the env is wrong

caohdgege:
image

It seems ok after swtich to newest agent

caohdgege (approved):
LGTM

slievrly (approved):
LGTM

PR #3869, opened 2021-07-16
feature: support get configuration from environment

slievrly:

feature: support get configuration from environment


config/seata-config-core/src/main/java/io/seata/config/Configuration.java
```diff
@@ -16,15 +16,19 @@
...
+import io.seata.common.util.StringUtils;
+
 /**
  * The interface Configuration.
  *
  * @author slievrly
  */
 public interface Configuration {
 
+    Map<String, String> ENV_MAP = System.getenv();
```
a364176773, line 31: System.getProperties() 也支持一下吧？

slievrly: 1. Env is immutable, so there is no setEnv method and it can use static variable global reference.

2. System.getProperty() is runtime mutable. It can replace a property with System.setProperty, so you can't have a static global reference.

lj2018110133:
LGTM

a364176773 (approved):
LGTM

lj2018110133 (approved):
LGTM

PR #3881, opened 2021-07-19
bugfix: fix getConfig with different default value return the first

slievrly:

bugfix: fix getConfig with different value return the first


jsbxyyx (approved):
LGTM

PR #3932, opened 2021-08-05
optimize:use map instead of if else judge for more change in the future

coconut4j:
use two ImmutableMap instead of lots of if else judge, and the code is more readability,more convenient to modify.

 if enum class don't use  Strategy Pattern,you can use map to optimize . can also use  the same  to io.seata.server.storage.db.store.DataBaseTransactionStoreManager#writeSession for db.


xilou31:
I think it's a bit over-engineered, if you want to optimize lots of "if else", you can use switch.

server/src/main/java/io/seata/server/storage/redis/store/RedisTransactionStoreManager.java
```diff
@@ -92,20 +97,58 @@ public static RedisTransactionStoreManager getInstance() {
...
+     */
+    public static volatile ImmutableMap<LogOperation, Function<BranchTransactionDO, Boolean>> branchMap;
+
+    /**
+     * init globalMap
+     *
+     * @param
+     * @return void
+     */
+    @PostConstruct
```
a364176773, line 117: 注解应该是无效的，这个manager不是一个spring bean，应该放到getInstance 进行初始化

jsbxyyx:
git name and email relation github id.
the init xxx map code move to class constructor.

server/src/main/java/io/seata/server/storage/redis/store/RedisTransactionStoreManager.java
```diff
@@ -92,20 +97,58 @@ public static RedisTransactionStoreManager getInstance() {
         return instance;
     }
 
+
+    /*
+    Map for LogOperation Global Operation
+    */
+    public static volatile ImmutableMap<LogOperation, Function<GlobalTransactionDO, Boolean>> globalMap;
```
l81893521, line 104: Why not use one map instead of two?

coconut4j: in this class，two map Fuction<T,R> has diff T

caohdgege:
plz resolve the code conflicts, or maybe you can create a new branch from develop & resubmit the pr.

server/src/main/java/io/seata/server/storage/redis/store/RedisTransactionStoreManager.java
l81893521: ok, please fix another review.

PR #4005, opened 2021-09-07
bugfix:PK constraint name isn't the same as the unique index name which is belong to PK

### LINKED ISSUE #3987: all_cons_columns查出的主键名和all_indexes中的索引名字不一致导致事物异常
### Ⅰ. Issue Description
事物启动异常

### Ⅱ. Describe what happened
1630296053502_1630294964657_8FA84DE8-EA11-4fd3-979B-863ED99F4833
如图：构造tablemeta信息时，通过两个sql查询表的相关索引信息，在图中1的位置，通过查询all_indexes取得所有的索引，
[在图中2，3位置通过查询all_cons_columns设置主键，但是由于all_indexes查出来的信息不包含all_cons_columns查出来的信息，就导致没有执行到第3步。
查询数据库的索引信息如下图：
1630295961095_7FB4AC8F-8B4B-494e-A709-5AAABCC91D4D
1630295961211_9623A18E-26E8-4a98-818D-349D138C2BDA
查询的sql是根据debug整理出来的：
select null as table_cat,
       owner as table_schem,
       table_name,
       0 as NON_UNIQUE,
       null as index_qualifier,
       null as index_name, 0 as type,
       0 as ordinal_position, null as column_name,
       null as asc_or_desc,
       num_rows as cardinality,
       blocks as pages,
       null as filter_condition
from all_tables
where table_name = '表名'
  and owner = '数据库名'
union
select null as table_cat,
       i.owner as table_schem,
       i.table_name,
       decode (i.uniqueness, 'UNIQUE', 0, 1),
       null as index_qualifier,
       i.index_name,
       1 as type,
       c.column_position as ordinal_position,
       c.column_name,
       null as asc_or_desc,
       i.distinct_keys as cardinality,
       i.leaf_blocks as pages,
       null as filter_condition
from all_indexes i, all_ind_columns c
where i.table_name = '表名'
  and i.owner = '数据库名'
  and i.index_name = c.index_name
  and i.table_owner = c.table_owner
  and i.table_name = c.table_name
  and i.owner = c.index_owner
order by non_unique, type, index_name, ordinal_position
;
SELECT NULL AS table_cat,
       c.owner AS table_schem,
       c.table_name,
       c.column_name,
       c.position AS key_seq,
       c.constraint_name AS pk_name
FROM all_cons_columns c, all_constraints k
WHERE k.constraint_type = 'P'
  AND k.table_name = '表名'
  AND k.owner like '数据库名' escape '/'
  AND k.constraint_name = c.constraint_name 
  AND k.table_name = c.table_name 
  AND k.owner = c.owner 
ORDER BY column_name;

io.seata.common.exception.NotSupportYetException: ACT_RU_EXECUTION needs to contain the primary key.
	at io.seata.rm.datasource.sql.struct.TableMeta.getPrimaryKeyMap(TableMeta.java:126)
	at io.seata.rm.datasource.sql.struct.TableMeta.getPrimaryKeyOnlyName(TableMeta.java:140)

驱动的maven坐标：
<dependency>
            <groupId>com.oracle.database.jdbc</groupId>
            <artifactId>ojdbc8</artifactId>
            <version>12.2.0.1</version>
 </dependency>

基中一个表的语句（由于表是历史表，最原始的开发建表语句，中间有过什么操作的相关记录都遗失了，并且数据库有一批表有这种情况）：
-- Create table
create table ACT_RU_EXECUTION
(
  id_               NVARCHAR2(64) not null,
  rev_              NUMBER,
  proc_inst_id_     NVARCHAR2(64),
  business_key_     NVARCHAR2(255),
  parent_id_        NVARCHAR2(64),
  proc_def_id_      NVARCHAR2(64),
  super_exec_       NVARCHAR2(64),
  act_id_           NVARCHAR2(255),
  is_active_        NUMBER(1),
  is_concurrent_    NUMBER(1),
  is_scope_         NUMBER(1),
  is_event_scope_   NUMBER(1),
  suspension_state_ NUMBER,
  cached_ent_state_ NUMBER,
  tenant_id_        NVARCHAR2(255) default '',
  name_             NVARCHAR2(255),
  lock_time_        TIMESTAMP(6)
)
tablespace DB_DATA
  pctfree 10
  initrans 1
  maxtrans 255
  storage
  (
    initial 3
    next 8
    minextents 1
    maxextents unlimited
  );
-- Create/Recreate indexes 
create index ACT_IDX_EXEC_BUSKEY on ACT_RU_EXECUTION (BUSINESS_KEY_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 5M
    next 1M
    minextents 1
    maxextents unlimited
  );
create index ACT_IDX_EXE_PARENT on ACT_RU_EXECUTION (PARENT_ID_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 64K
    next 1M
    minextents 1
    maxextents unlimited
  );
create index ACT_IDX_EXE_PROCDEF on ACT_RU_EXECUTION (PROC_DEF_ID_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 6M
    next 1M
    minextents 1
    maxextents unlimited
  );
create index ACT_IDX_EXE_PROCINST on ACT_RU_EXECUTION (PROC_INST_ID_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 5M
    next 1M
    minextents 1
    maxextents unlimited
  );
create index ACT_IDX_EXE_SUPER on ACT_RU_EXECUTION (SUPER_EXEC_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 64K
    next 1M
    minextents 1
    maxextents unlimited
  );
create unique index SYS_C0012210 on ACT_RU_EXECUTION (ID_)
  tablespace DB_DATA
  pctfree 10
  initrans 2
  maxtrans 255
  storage
  (
    initial 5M
    next 1M
    minextents 1
    maxextents unlimited
  );
-- Create/Recreate primary, unique and foreign key constraints 
alter table ACT_RU_EXECUTION
  add primary key (ID_);
-- Create/Recreate check constraints 
alter table ACT_RU_EXECUTION
  add check (IS_EVENT_SCOPE_ IN (1,0));
alter table ACT_RU_EXECUTION
  add check (IS_SCOPE_ IN (1,0));
alter table ACT_RU_EXECUTION
  add check (IS_CONCURRENT_ IN (1,0));
alter table ACT_RU_EXECUTION
  add check (IS_ACTIVE_ IN (1,0));

### Ⅵ. Environment:

- JDK version :1.8
- seata1.4.2
- spring cloud Hoxton.SR12
- spring boot 2.3.12.RELEASE
- oracle 12.2.0.1.0
- jdbc driver ojdbc8 12.2.0.1

Musknine:
In some case,the table first created the unique index, and then created the PK but didn't give a explicit name the same as unique index name.

fixes #3987


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
```diff
@@ -147,19 +150,39 @@ private TableMeta resultSetMetaToSchema(DatabaseMetaData dbmd, String tableName)
...
                     IndexMeta index = tm.getAllIndexes().get(pkIndexName);
                     index.setIndextype(IndexType.PRIMARY);
+                } else {
+                    pkcol.add(rsPrimary.getString("COLUMN_NAME"));
                 }
             }
-            if (tm.getAllIndexes().isEmpty()) {
-                throw new ShouldNeverHappenException(String.format("Could not found any index in the table: %s", tableName));
+            if (!pkcol.isEmpty()) {
+                int times = 0;
```
l81893521, line 167: What is the times means for?

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
```diff
@@ -147,19 +150,39 @@ private TableMeta resultSetMetaToSchema(DatabaseMetaData dbmd, String tableName)
...
+                for (Map.Entry<String, IndexMeta> entry : tm.getAllIndexes().entrySet()) {
+                    IndexMeta index = entry.getValue();
+                    if (index.getIndextype().value() == IndexType.UNIQUE.value()) {
+                        for (ColumnMeta col : index.getValues()) {
+                            if (pkcol.contains(col.getColumnName())) {
+                                times++;
```
l81893521, line 173: How about change it to Primary directly if pkcol constains col.getColumnName?

Musknine: > How about change it to Primary directly if pkcol constains col.getColumnName?
I couldn't agree with you more.   but i thought we need to check the situation that PK is multiple column PK (i was not sure if the seata will surpport  composite primary key)

l81893521: I have a idea that is when rsPrimary is not empty and put a new IndexMeta instead of change it. We separate the Unique Key and Primary Key. How do you think about this?

l81893521: We change the Unique Key to Primary key before because when we create the primary key and Oracle will create a unique key automatically but now look like the solution of change the key type is now that good.

Musknine: for oracle will create a unique key automatically, it's not happend in  all the situation. only ,if there is no unique index,it would create one automatically.but i thought my PR logic is right.if the index is unique and the column is exacly same as PK column.  it means the unique index is the PK.     I couldn't agree with you more.  it's a good idea that use a individual IndexMeta to indicate PK,actually,i had thought this solution before. but i didn't check all the code of the seata,i have no idea how many places need to update for.    IN some other Oracle like database, example, 达梦,the defalut unique index name that was created by when we created the PK always  was not the same as PK name.

Musknine: for oracle will create a unique key automatically, it's not happend in  all the situation. only ,if there is no unique index,it would create one automatically.but i thought my PR logic is right.if the index is unique and the column is exacly same as PK column.  it means the unique index is the PK.     I couldn't agree with you more.  it's a good idea that use a individual IndexMeta to indicate PK,actually,i had thought this solution before. but i didn't check all the code of the seata,i have no idea how many places need to update for.    IN some other Oracle like database, example, 达梦,the defalut unique index name that was created by when we created the PK always  was not the same as PK name.

l81893521: You do not need to change all the code in seata, just need to keep the number of primary key and index type is right that s ok. You can check out the main code here, it will no problem if you keep the logic as same as before.
```
io.seata.rm.datasource.sql.struct.TableMeta#getPrimaryKeyOnlyName
io.seata.rm.datasource.sql.struct.TableMeta#getPrimaryKeyMap
```

Musknine: I got it.I'll push a new commit later.

Musknine: I got it what you meaning is.but it's not easy to determine how can we get the indexMeta property.actually,the PK is constraint,the actual unique index was created by PK is stored in the view of all_indexes.it means we need to determine which unique index was a part of the PK(PK include constraint name and index name).
image
indeed,at step 2,it's not a indexname,it's a constraint name. maybe the current PR is good.    i thought the key  logic is accoding the PK constriant name to determine which unique index name was a part of the PK.

l81893521: OK. You idea was right. You still have to add some comments and let other developer understand . For example:
```
// when we create a primary key constraint oracle will uses and existing unique index.
// if we create a unique index before create a primary constraint in the same column will cause the problem that primary key constraint name was different from the unique name.
// a table can have only one primary key
List<String> pkcol = new ArrayList<>();

...
//save the columns that constraint primary key name was different from unique index name
pkcol.add(rsPrimary.getString("COLUMN_NAME"));

...
//find the index that belong to the primary key constraint
for (Map.Entry<String, IndexMeta> entry : tm.getAllIndexes().entrySet())
```

Musknine: thanks for your advice. I'll add some necessary comments later.

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
slievrly: matchCols

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
```diff
@@ -30,6 +25,14 @@
 import io.seata.rm.datasource.sql.struct.TableMeta;
 import io.seata.sqlparser.util.JdbcConstants;
 
+import java.sql.Connection;
```
slievrly, line 28: don't move imports

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
slievrly: @Musknine

rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/OracleTableMetaCache.java
Musknine: @a364176773 @slievrly @l81893521
  1, I moved back some imports that i moved down before.
   2,I have added some comments.

l81893521 (approved):
LGTM

jsbxyyx (approved):
LGTM

PR #4013, opened 2021-09-12
optimize: optimize client channel alive check

### LINKED ISSUE #4002: Service address exploration of the registry center
## Why you need it?

In some cases, for example, the TC has registered 127.0.0.1:8091 to the registry center or the registered IP is unable to communicate with the client. In this case, the health detection result of the registry center is that the TC is alive, but the network cannot be connected to the client, so it is necessary to explore the service list of the registry center and remove the TC that the client cannot communicate;

## How it could be?
Ensure that the registry center list is network connectable to the client.

a364176773:

fixes #4002


slievrly (approved):
LGTM

PR #4085, opened 2021-10-16
optimize: client check whether undolog table exist before cleaning undolog

selfishlover:
when RM receive UndoLogDeleteRequest(send by TC), if user does not use AT mode at all(which mean their database does not have undo_log table), exception would occur(such as undo_log table doesn't existed~) and logger will print error info(which scares user~). Here, we check whether undo_log table exist before trying to delete undolog.


core/src/main/java/io/seata/core/rpc/netty/RmNettyRemotingClient.java
```diff
@@ -173,42 +167,49 @@ public void onRegisterMsgFail(String serverAddress, Channel channel, Object resp
...
+    /**
+     * build RegisterRMRequest
+     * note: BranchType store in message.extraData
+     */
+    private RegisterRMRequest buildRegisterRMRequest(String resourceId, BranchType type) {
         RegisterRMRequest message = new RegisterRMRequest(applicationId, transactionServiceGroup);
         message.setResourceIds(resourceId);
+        if (type != null) {
+            String typeName = type.name();
+            message.setExtraData(typeName);
```
slievrly, line 207: If you need to add additional fields to ExtraData, consider the extensibility design.

core/src/main/java/io/seata/core/rpc/netty/ChannelManager.java
```diff
@@ -121,81 +136,99 @@ private static RpcContext buildChannelHolder(NettyPoolKey.TransactionRole client
...
+        Integer clientPort = ChannelUtil.getClientPortFromChannel(channel);
+
+        idSet.forEach(resourceId -> {
+            addBranchTypeToResource(resourceId, type);
 
+            ConcurrentMap<Integer, RpcContext> portMap = getRmPortMap(resourceId, appId, clientIp);
+            portMap.put(clientPort, rpcContext);
             rpcContext.holdInResourceManagerChannels(resourceId, portMap);
-            updateChannelsResource(resourceId, clientIp, resourceManagerRequest.getApplicationId());
+        });
```
slievrly, line 191: Multiple RM channels and resourceId on the same client.
ip: ip1   channels: channel1, channel2    resourceId: resourceId1, resourceId2.

1.  resourceId1 init -> create channel1  send  RegisterRMRequest to TC.
2. resourceId2 init
3. create channel2  send resourceId1 & resourceId2 to TC
4. begin global transaction with resourceId2 through channel2
5.  channel2 inactive 
6. how use channel1 to continue the resourceId2's transaction?

PR #4096, opened 2021-10-25
feature：support db2 in AT mode

### LINKED ISSUE #3744: feature: Seata extension IBM db2 support
描述：

目前 Seata 支持的数据库有 MySQL、Oracle、PostgreSQL、TiDB等，期望 Seata 能够扩展更多的数据库支持。要求参照 Seata 已有的数据库支持实现 IBM db2 的扩展.

Seata currently supports MySQL, Oracle, PostgreSQL, TIDB, etc. Seata is expected to expand more database support. Requires reference to SEATA existing database support to implement IBM DB2.

#3741

GoodBoyCoder:

support DB2 DataBase

fixes #3744

- Identity column of type generated always as identity is not supported. This type of Identity value can only be generated by the system, which will cause Execption when performing Delete rollback compensation. If necessary, it is recommended to use BY DEFAULT
- "Insert into table_name (...) values (?, ?, ?, ?);" This type of SQL with ";" added after it, if you add a sign that needs to return the auto-incremented primary key value (such as Statement.RETURN_GENERATED_KEYS, primary key column), it will be abnormal when executed, it is normal not to add ";"


funky-eyes (dismissed):
LGTM

PR #4146, opened 2021-11-18
optimize: remote configuration reading

a364176773:

### DIFF
--- a/config/seata-config-core/src/main/java/io/seata/config/ConfigurationCache.java
+++ b/config/seata-config-core/src/main/java/io/seata/config/ConfigurationCache.java
@@ -38,6 +38,8 @@ public class ConfigurationCache implements ConfigurationChangeListener {
 
     private static final ConcurrentHashMap<String, ObjectWrapper> CONFIG_CACHE = new ConcurrentHashMap<>();
 
+    private static final Configuration FILE_CONFIG = ConfigurationFactory.CURRENT_FILE_INSTANCE;
+
     private Map<String, HashSet<ConfigurationChangeListener>> configListenersMap = new HashMap<>();
 
     public static void addConfigListener(String dataId, ConfigurationChangeListener... listeners) {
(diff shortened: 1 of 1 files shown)


config/seata-config-core/src/main/java/io/seata/config/ConfigurationCache.java
```diff
@@ -106,21 +109,32 @@ public Configuration proxy(Configuration originalConfiguration) {
...
                     if (args.length > 1 && method.getParameterTypes()[1].getSimpleName().equalsIgnoreCase(type.name())) {
                         defaultValue = args[1];
                     }
-                    if (null == wrapper || (null != defaultValue && !Objects.equals(defaultValue, wrapper.lastDefaultValue))) {
-                        Object result = method.invoke(originalConfiguration, args);
-                        // The wrapper.data only exists in the cache when it is not null.
-                        if (result != null) {
-                            wrapper = new ObjectWrapper(result, type, defaultValue);
-                            CONFIG_CACHE.put(rawDataId, wrapper);
+                    Object finalDefaultValue = defaultValue;
```
slievrly, line 117: why not contains `final` modifier.

a364176773: done

config/seata-config-core/src/main/java/io/seata/config/ConfigurationCache.java
```diff
@@ -106,21 +109,32 @@ public Configuration proxy(Configuration originalConfiguration) {
...
+                    ObjectWrapper wrapper = CollectionUtils.computeIfAbsent(CONFIG_CACHE, rawDataId,
+                        k -> new ObjectWrapper(null, type, finalDefaultValue));
+                    boolean notEquals = null != defaultValue && !Objects.equals(defaultValue, wrapper.lastDefaultValue);
+                    if (wrapper.getData() == null || notEquals) {
+                        synchronized (wrapper) {
+                            if (wrapper.getData() == null || notEquals) {
+                                Object result = method.invoke(originalConfiguration, args);
+                                Object localResult = null;
+                                // When the remote configuration item does not exist, try to read the local configuration item
+                                if (result == null || (defaultValue != null && Objects.equals(result, defaultValue))) {
```
slievrly, line 127: 1. not only the configuration center, but also the YML configuration file.
2. if yml as configuration file and set as default value, it will read twice.

slievrly:
I think the troubleshooting path will be increased. It is not clear whether the configuration center configuration or local configuration is used.

This situation is mainly to fix several dataid configuration problems. If the client has only one dataid, the probability of occurrence will be greatly reduced.

Therefore, I still suggest the fail-fast solution. When the key configuration is missing and there is no default value, an error will be throwed directly.

a364176773:
> I think the troubleshooting path will be increased. It is not clear whether the configuration center configuration or local configuration is used.
> 
> This situation is mainly to fix several dataid configuration problems. If the client has only one dataid, the probability of occurrence will be greatly reduced.
> 
> Therefore, I still suggest the fail-fast solution. When the key configuration is missing and there is no default value, an error will be throwed directly.

fast failure will bring some unnecessary troubles. For example, when I want to debug a configuration item locally and put it in the configuration center after the desired effect, this will not be possible. Unless you switch back and forth between file and other configuration centers, this will be cumbersome and redundant. We should not have file mode, but a way of combining file with other configuration centers.

PR #4216, opened 2021-12-18
optimize: client check whether undolog table exist before cleaning undolog

selfishlover:
when RM receive UndoLogDeleteRequest(send by TC), if user does not use AT mode at all(which mean their database does not have undo_log table), exception would occur(such as undo_log table doesn't existed~) and logger will print error info(which scares user~). Here, we check whether undo_log table exist before trying to delete undolog.

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
+++ b/rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
@@ -19,12 +19,15 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import io.seata.core.model.BranchType;
 import io.seata.core.model.ResourceManager;
 import io.seata.core.protocol.transaction.UndoLogDeleteRequest;
 import io.seata.rm.datasource.DataSourceManager;
 import io.seata.rm.datasource.DataSourceProxy;
+import io.seata.rm.datasource.undo.UndoLogManager;
 import io.seata.rm.datasource.undo.UndoLogManagerFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -40,43 +43,93 @@
...
+        DataSourceProxy dataSourceProxy = dataSourceManager.get(resourceId);
         if (dataSourceProxy == null) {
-            LOGGER.warn("Failed to get dataSourceProxy for delete undolog on {}", request.getResourceId());
+            LOGGER.warn("Failed to get dataSourceProxy for delete undolog on {}", resourceId);
+            return;
+        }
+
+        boolean hasUndoLogTable = undoLogTableExistRecord.computeIfAbsent(resourceId, id -> checkUndoLogTableExist(dataSourceProxy));
+        if (!hasUndoLogTable) {
+            LOGGER.debug("resource({}) has no undo_log table, UndoLogDeleteRequest will be ignored", resourceId);
```
a364176773, line 60: if (logger.isDebugEnabled()) {
}

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
```diff
@@ -40,43 +43,93 @@
...
             return;
         }
-        Date logCreatedSave = getLogCreated(request.getSaveDays());
-        Connection conn = null;
+
+        Connection conn = getConnection(dataSourceProxy);
```
a364176773, line 64: 用try-with-resources?

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
selfishlover: Almost all Logger implement will do the same level check inside deug(), info()......
So we don't need to check again.

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
selfishlover: good idea. I will use this.

rm-datasource/src/main/java/io/seata/rm/RMHandlerAT.java
a364176773: ok

a364176773 (approved):
LGTM

caohdgege (approved):
LGTM

caohdgege:
补充一下md

slievrly (approved):
LGTM

Issue #4267, opened 2022-01-05
develop 分支 Redis Pipeline 模式获取全局锁时返回结果数据顺序被打乱


objcoding:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
io.seata.server.storage.redis.lock.RedisLocker#acquireLockByPipeline
redis返回的数据打乱了，导致与原来的needLockKeys，needLockDOS 顺序对应不上

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- Seata version:  
- OS :
- Others:

a364176773:
该问题仅限develop分支，其余分支无任何影响

PR #4277, opened 2022-01-09
optimize: acquire lock return fail-fast code in redis-pipeline mode.

objcoding:

When the local transaction and the global transaction are enabled, the branch registration fails to acquire the global lock, the lock holder is in the second-stage rollback, and the branch registration fails to be retried quickly, because the retry with the local transaction does not release the database lock , resulting in a two-phase rollback wait. Therefore, if a global lock is found in the Rollbacking state, the fail-fast code is returned directly.

#4267


a364176773 (approved):
LGTM 补充下1.5.0.md

a364176773:
按标题的意思这个pr应该是个optimize

objcoding:
> 按标题的意思这个pr应该是个optimize

是的，原则上是pr3733的一个优化，因为pr3733并没有打包release版本，还处于develop分支。

slievrly (approved):
LGTM

PR #4331, opened 2022-01-24
bugfix:fix dirty write check exception that may occur when using ONLY_CARE_UPDATE_COLUMNS

### LINKED ISSUE #4330: AT模式脏写校验异常

### Ⅰ. Issue Description
Insert类型SQL，当开启ONLY_CARE_UPDATE_COLUMNS配置项，SQL语句中显式写明插入主键字段，且数据库表的字段与SQL语句中写明的字段大小写不一致时，回滚会出现脏写校验异常（会误认为存在两个主键）
Example:insert into table_name (id) values(?);
”id“为table_name中的主键字段，且表中存储的字段为大写“ID”
则BaseTransactionalExecutor#buildTableRecords会出现重复添加字段的情况
image

Insert type SQL, when the ONLY_CARE_UPDATE_COLUMNS configuration item is enabled, the primary key field to be inserted is explicitly stated in the SQL statement, and the case of the fields of the database table is inconsistent with the field case specified in the SQL statement, a dirty write check exception will occur during rollback (will mistakenly think that there are two primary keys)
Example: insert into table_name (id) values(?);
”id“ is the primary key field in table_name, and the field stored in the table is an uppercase ”ID“
Then BaseTransactionalExecutor#buildTableRecords will repeatedly add fields

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- Seata version:  
- OS :
- Others:

GoodBoyCoder:

使用TreeSet代替HashSet进行字段去重

Use TreeSet instead of HashSet to perform field deduplication

fixes #4330


a364176773 (approved):
LGTM

slievrly (approved):
LGTM

PR #4345, opened 2022-01-27
optimize: fix the path of the package

wangliang181230:
修正包目录名：
1. `io.seata.config` -> `io/seata/config`
2. `io.seata.plugin.jackson.parser.oracle` -> `io/seata/plugin/jackson/parser/oracle`


slievrly (approved):
LGTM

PR #4448, opened 2022-03-08
optimize:When the Seata client and server adopt Nacos as the registry, the client will subscribe to the Seata server and the application name will be unknown

hadoop835:
When the Seata client and server adopt Nacos as the registry, the client will subscribe to the Seata server and the application name will be unknown.
1646721213617_1DA5E72A-349A-45d3-8025-434339207EC0

client config
image
image


a364176773 (changes requested):
script文件夹中client中的相关配置补充一下

a364176773:
标题英文,optimize: 开头

seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/io/seata/spring/boot/autoconfigure/properties/registry/RegistryNacosProperties.java
```diff
@@ -34,9 +34,10 @@ public class RegistryNacosProperties {
     private String password;
     private String accessKey;
     private String secretKey;
-    private String application = "seata-server";
-    private String slbPattern;
-    private String contextPath;
+    private String application = "seata-server"; 
+    private String clientApplication = "";
```
a364176773, line 38: 未涉及的逻辑和代码不要去改动

a364176773 (commented):
yml中的使用方式需要补充进去

hadoop835:
> yml中的使用方式需要补充进去

加在哪里呢

a364176773:
> > yml中的使用方式需要补充进去
> 
> 加在哪里呢

server里的resources里的application.example.yml

PR #4625, opened 2022-05-21
refactor：Migrate the registerProcessor method

106umao:

I was thinking about implementing the grpc extension of the rpc module, and found that registerProcessor has little correlation with RemotingClient and RemotingServer. It is more like a feature implemented by Netty, so I migrated this method to AbstractNettyRemoting to purify the interface responsibilities.

### DIFF
--- a/core/src/main/java/io/seata/core/rpc/RemotingClient.java
+++ b/core/src/main/java/io/seata/core/rpc/RemotingClient.java
@@ -20,9 +20,6 @@ import io.seata.core.protocol.AbstractMessage;
 import io.seata.core.protocol.RpcMessage;
 import io.seata.core.rpc.netty.RmNettyRemotingClient;
 import io.seata.core.rpc.netty.TmNettyRemotingClient;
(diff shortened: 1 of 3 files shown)


a364176773:
Sorry. We'll handle the GRPC support ourselves

106umao:
> Sorry. We'll handle the gRPC support ourselves

I think the old code needs to be modified even aside from the grpc extension. Maybe my understanding of this piece is not as good as yours. Can you point out why the original code needs to be designed this way? Register the processor in the top-level calling interface

a364176773:
> Reference in ne

original extension is not abstracted correctly, which we will deal with later

106umao:
> > Reference in ne
> 
> original extension is not abstracted correctly, which we will deal with later

Ok, I signed up for this grpc project of Open Source Summer. If I am lucky enough to win the bid, I hope to communicate with you about the design plan later

a364176773:
If you don't mind my asking, is who invited you to join Seata github organization?

PR #4713, opened 2022-06-17
feature: support xid load balance

### LINKED ISSUE #4688: supports load balance by xid
## Why you need it?

Currently Seata interacts with TC load balance for each time, which causes some problems:

1. XA 2phase branch commit/rollback could not find the original XAConn

2. Atomic state issues between nodes, such as missing later registered branch transactions when rolling back

3. Log aggregation for a single transaction

4. Metrics single-node active transaction negative number problem

Therefore, load balance needs to be done according to xid to solve the above problems. In addition, the availability of TC should also be considered.

## How it could be?

a364176773:

fixes #4688

### DIFF
--- a/discovery/seata-discovery-core/src/main/java/io/seata/discovery/loadbalance/XIDLoadBalance.java
+++ b/discovery/seata-discovery-core/src/main/java/io/seata/discovery/loadbalance/XIDLoadBalance.java
@@ -25,6 +25,8 @@ import java.util.Objects;
 import io.seata.common.loader.EnhancedServiceLoader;
 import io.seata.common.loader.LoadLevel;
 import io.seata.common.util.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The type xid load balance.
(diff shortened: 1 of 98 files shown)


discovery/seata-discovery-core/src/main/java/io/seata/discovery/loadbalance/XIDLoadBalance.java
```diff
@@ -0,0 +1,58 @@
...
+        LoadBalanceFactory.RANDOM_LOAD_BALANCE);
+
+    private static final String SPLIT = ":";
+
+    @Override
+    public <T> T select(List<T> invokers, String xid) throws Exception {
+        if (StringUtils.isNotBlank(xid) && xid.contains(SPLIT)) {
+            String[] xidArray = xid.split(SPLIT);
+            int port = Integer.parseInt(xidArray[1]);
+            String ip = xidArray[0];
```
slievrly, line 46: consider ipv6 support

a364176773: done

a364176773:
> https://github.com/seata/seata/pull/4713/files#diff-67c903540585b858205a835c0483c7d97511061428c5aaf2de0f70a774579ee1R59 expect to downgrade to other load balancing modes to ensure availability.

done

slievrly (approved):
LGTM

PR #4761, opened 2022-07-09
optimize: use hget replace hmget because only one field

conghuhu:

In io.seata.server.storage.redis.lock.RedisLocker and 362 line, there only one field, so we should use hget not hmget 

run io.seata.server.lock.redis.RedisLockManagerTest#unLock()


caohdgege (approved):
LGTM

PR #4963, opened 2022-09-27
optimize: require ARM64 binary

odidev:
Modified file  **.github/workflows/test.yml**, **dependencies/pom.xml**, **discovery/seata-discovery-etcd3/pom.xml**, **discovery/seata-discovery-etcd3/src/test/java/io/seata/discovery/registry/etcd/EtcdRegistryServiceImplTest.java**, i**ntegration/grpc/pom.xml** and **serializer/seata-serializer-protobuf/pom.xml** for arm64 binary

**Signed-off-by:** odidev <odidev@puresoftware.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,24 @@ on:
     branches: [ develop, master ]
 
 jobs:
+  build_arm64-binary: 
+     runs-on: ubuntu-latest
+     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/develop' }}   
+     strategy:
+      fail-fast: false
+     steps:
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+      - name: Build arm-binary
+        run: |
+               docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+               arm64v8/ubuntu:20.04 \
+               bash -exc 'apt-get update -y && \
+                apt-get install maven -y && \
+                mvn -Prelease-seata -Dmaven.test.skip=true clean install -U'
+
   build:
     name: "build"
     runs-on: ubuntu-latest
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -53,7 +53,7 @@
         <eureka-clients.version>1.10.17</eureka-clients.version>
         <consul-clients.version>1.4.2</consul-clients.version>
         <nacos-client.version>1.4.2</nacos-client.version>
-        <etcd-client-v3.version>0.3.0</etcd-client-v3.version>
+        <etcd-client-v3.version>0.5.0</etcd-client-v3.version>
         <testcontainers.version>1.11.2</testcontainers.version>
         <guava.version>27.0.1-jre</guava.version>
         <javax-inject.version>1</javax-inject.version>
@@ -66,8 +66,8 @@
         <ant.version>1.10.12</ant.version>
         <lz4.version>1.7.1</lz4.version>
 
-        <protobuf.version>3.7.1</protobuf.version>
-        <grpc.version>1.17.1</grpc.version>
+        <protobuf.version>3.11.4</protobuf.version>
+        <grpc.version>1.27.1</grpc.version>
         <kryo.version>5.3.0</kryo.version>
         <kryo-serializers.version>0.45</kryo-serializers.version>
         <hessian.version>4.0.63</hessian.version>
--- a/build/pom.xml
+++ b/build/pom.xml
@@ -82,7 +82,7 @@
         <maven-clean-plugin.version>3.1.0</maven-clean-plugin.version>
         <!-- Compiler -->
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
-        <protobuf-maven-plugin.version>0.5.0</protobuf-maven-plugin.version>
+        <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
         <kotlin-maven-plugin.version>1.3.72</kotlin-maven-plugin.version>
         <!-- Check -->
         <maven-pmd-plugin.version>3.8</maven-pmd-plugin.version>
--- a/discovery/seata-discovery-etcd3/src/test/java/io/seata/discovery/registry/etcd/EtcdRegistryServiceImplTest.java
+++ b/discovery/seata-discovery-etcd3/src/test/java/io/seata/discovery/registry/etcd/EtcdRegistryServiceImplTest.java
@@ -18,7 +18,7 @@ package io.seata.discovery.registry.etcd;
 import io.etcd.jetcd.ByteSequence;
 import io.etcd.jetcd.Client;
 import io.etcd.jetcd.Watch;
-import io.etcd.jetcd.launcher.junit.EtcdClusterResource;
+import io.etcd.jetcd.launcher.junit4.EtcdClusterResource;
 import io.etcd.jetcd.options.DeleteOption;
 import io.etcd.jetcd.options.GetOption;
 import io.etcd.jetcd.watch.WatchResponse;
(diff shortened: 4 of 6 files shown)


slievrly:
@odidev 
I think that your PR solves:

1. Upgrade `protobuf` to solve the ARM64 compatibility problem.
2. Check whether there is any problem on the ARM64 platform CI through Github Action Workflow.

.github/workflows/build.yml
```diff
@@ -7,6 +7,28 @@ on:
...
+     steps:
+      - uses: actions/checkout@v2
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+      - name: Build arm-binary
+        run: |
+               docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+               arm64v8/ubuntu:20.04 \
+               bash -exc 'apt-get update -y && \
```
wangliang181230, line 25: `-exc` -> `-exec` ?

.github/workflows/build.yml
```diff
@@ -7,6 +7,28 @@ on:
     branches: [ develop, master ]
 
 jobs:
+  build_arm64-binary: 
+     runs-on: ubuntu-latest
+     strategy:
+      matrix:
+        java: [ 8, 11 ]
```
slievrly, line 14: `openjdk-8-jdk` is installed in Docker. Does the declaration here make sense?

.github/workflows/build.yml
slievrly, line 25: `apt-get update'  if necessary?  Is it time-consuming?

.github/workflows/build.yml
```diff
@@ -7,6 +7,28 @@ on:
...
+      - name: Set up QEMU
+        id: qemu
+        uses: docker/setup-qemu-action@v1
+      - name: Build arm-binary
+        run: |
+               docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+               arm64v8/ubuntu:20.04 \
+               bash -exc 'apt-get update -y && \
+                apt-get install openjdk-8-jdk -y && \
+                export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-arm64 && \
```
slievrly, line 27: `apt-get install` does the environment variable need to be set manually?

wangliang181230:
https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow
`runs-on` support `ARM64`, is it unnecessary to use scripts to build the `ARM64`?

slievrly:
workflow output
```
WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested
```
https://github.com/seata/seata/actions/runs/3135923871/jobs/5092189894

slievrly:
@odidev Any responses to the questions above?

odidev:
> Any responses to the questions above?

[@slievrly](https://github.com/slievrly) almost done, few things are left, it will take little more time. And sorry for the delay.

dependencies/pom.xml
```diff
@@ -65,9 +65,9 @@
         <commons-compress.version>1.19</commons-compress.version>
         <ant.version>1.10.6</ant.version>
         <lz4.version>1.7.1</lz4.version>
-
-        <protobuf.version>3.7.1</protobuf.version>
-        <grpc.version>1.17.1</grpc.version>
+        <protobuf-maven-plugin.version>0.6.1</protobuf-maven-plugin.version>
```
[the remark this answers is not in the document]
odidev: I have updated it in the build module as per your suggestion.

.github/workflows/build.yml
odidev: Not needed here so I have removed it.

.github/workflows/build.yml
odidev: It is necessary here because without apt-get update it doesn't locate the maven package and falling with below..
aaaaa

odidev:
>WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64) and no specific platform was requested

It's just a warning because, we are running an arm64 script over the amd machine using qemu.. I verified that hosted platform is also compatible for arm64 by using `uname-m`

bbbbbbbbbb

odidev:
> @odidev Any responses to the questions above?

[@slievrly](https://github.com/slievrly) I have updated all the changes as per review comments. Please have a look.

.github/workflows/build.yml
slievrly: make sense.

.github/workflows/build.yml
slievrly: resolved

slievrly:
> https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow `runs-on` support `ARM64`, is it unnecessary to use scripts to build the `ARM64`?

@odidev what do you think about of running on `ARM64 `?

odidev:
> > https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow
> > `runs-on` support `ARM64`, is it unnecessary to use scripts to build the `ARM64`?
> 
> @odidev what do you think about of running on `ARM64 `?

[@slievrly](https://github.com/slievrly)

I have built arm64-binary by using self-hosted arm64 runners which has to be hosted on arm64 machine. If it is fine, then we can get access to ARM H/W through [WorksOnArm](https://github.com/WorksOnArm)

**Self-Hosted arm64 Runner in GitHub-actions:**
**Commit Link:** https://github.com/odidev/seata/commit/85746ff8ac29c66bc2c23fda80f22d27d18fc456

**GHA Logs Link:** https://github.com/odidev/seata/actions/runs/3226451554/jobs/5279982536

If you agree, I will amend the PR.

slievrly:
> > > https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow
> > > `runs-on` support `ARM64`, is it unnecessary to use scripts to build the `ARM64`?
> > 
> > 
> > @odidev what do you think about of running on `ARM64 `?
> 
> [@slievrly](https://github.com/slievrly)
> 
> I have built arm64-binary by using self-hosted arm64 runners which has to be hosted on arm64 machine. If it is fine, then we can get access to ARM H/W through [WorksOnArm](https://github.com/WorksOnArm)
> 
> **Self-Hosted arm64 Runner in GitHub-actions:** **Commit Link:** [odidev@85746ff](https://github.com/odidev/seata/commit/85746ff8ac29c66bc2c23fda80f22d27d18fc456)
> 
> **GHA Logs Link:** https://github.com/odidev/seata/actions/runs/3226451554/jobs/5279982536
> 
> If you agree, I will amend the PR.

Why not just use `runs-on: ARM64 `?

odidev:
> Why not just use runs-on: ARM64 ?

@slievrly 

We can't add arm64 runner directly as arm64 runners are not supported directly for arm64. 

We can use two methods for running arm64 script over an x64 machine first one is using qemu and second is self-hosted runner for arm64. 

for self-hosted-runner, following are the steps.

**Adding a self-hosted runner to the repository:**

1. On GitHub.com, navigate to the main page of the repository.
2. Under odidev/seata, goto the **Settings**
    
3. In the left sidebar, click **Actions**, then click **Runners**
   
4. After that select the operating system as Linux and architecture as ARM64 for self-hosted runner machine. 


6. Then run these commands on your local machine:
    **For download**    
 ```
$ mkdir actions-runner && cd actions-runner
$ curl -o actions-runner-linux-arm64-2.298.2.tar.gz -L https://github.com/actions/runner/releases/download/v2.298.2/actions-runner-linux-arm64-2.298.2.tar.gz
$ echo "803e4aba36484ef4f126df184220946bc151ae1bbaf2b606b6e2f2280f5042e8  actions-runner-linux-arm64-2.298.2.tar.gz" | shasum -a 256 -c
$ tar xzf ./actions-runner-linux-arm64-2.298.2.tar.gz

```
**Configure**
```
$ ./config.sh --url https://github.com/odidev/seata --token AJXNBROLYUMW7SXRHZKTHQDDI62UY
$ ./run.sh
```
After that use “runs-on: self-hosted” yaml in the workflow file for each job.

slievrly:
@odidev I had some wrong understanding of this solution before, but now I have a clear understanding through github documents. I suggest using the QEMU method, but this workflow takes a long time to execute. I don't think it is suitable to put it on pull_ request。 I suggest putting it into the on push branch only for execution.

- https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners
- https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

odidev:
> @odidev I had some wrong understanding of this solution before, but now I have a clear understanding through github documents. I suggest using the QEMU method, but this workflow takes a long time to execute. I don't think it is suitable to put it on pull_ request. I suggest putting it into the on push branch only for execution.
> 
> * https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners
> * https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

[@slievrly](https://github.com/slievrly) 

It's ok to put it on a branch only for execution. Please let me know if I should commit the code to any specific branch or same branch works for you?

odidev:
@slievrly  
> @odidev I think it's enough to trigger the ARM64 compatibility check workflow when **push** the code into **develop** branch.

As per your suggestion, I updated the code into the on-push branch and skipped the pull_request.

slievrly (approved):
LGTM

slievrly:
@odidev thanks for your contribution.

PR #4992, opened 2022-10-14
fix on jdk17

onlyGuo:
bugfix: InaccessibleObjectException on jdk17


a364176773:
https://github.com/seata/seata/pull/4434

PR #5120, opened 2022-12-03
optimize:unify the format of configuration items in yml files

### LINKED ISSUE #5117: some parameters are not uniform

### Ⅰ. Issue Description
Some parameters have double quotes, but some parameters do not have double quotes, which should be unified.

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- Seata version:  
- OS :
- Others:

xingfudeshi:

fixes #5117


server/src/main/resources/application.example.yml
```diff
@@ -36,32 +36,32 @@ seata:
...
-      apollo-access-key-secret:
+      apollo-access-key-secret: ""
       cluster: seata
     zk:
       server-addr: 127.0.0.1:2181
       session-timeout: 6000
       connect-timeout: 2000
-      username:
-      password:
+      username: ""
```
slievrly, line 63: What is the yml specification for strings and why quotes are not required for String configurations that now have non-null values.

xingfudeshi: image

Because we have assigned the empty string value in the properties class and Assigning a default value is also a best practice to avoid null values.

xingfudeshi: I will review the properties class again and unify them.

seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/io/seata/spring/boot/autoconfigure/properties/config/ConfigApolloProperties.java
```diff
@@ -29,7 +29,7 @@ public class ConfigApolloProperties {
     private String appId = "seata-server";
     private String apolloMeta;
     private String namespace = "application";
-    private String apolloAccessKeySecret = "";
+    private String apolloAccessKeySecret;
```
a364176773, line 32: 如果默认值不为"",会不会有些地方拿了做处理导致npe?,默认值至为null需要保证下这些值没有被直接未判空使用,请做下检查

xingfudeshi: Yes.I've done the checking.

a364176773 (approved):
LGTM

slievrly (approved):
LGTM

PR #5186, opened 2022-12-21
test: fix `test-compile` fails when the version of the `spring-boot` is lower than `2.4.0`

wangliang181230:
test: fix the problem that the `test-compile` fails when the version of the `spring-boot` is lower than `2.4.0`.
测试：修复 当`spring-boot` 版本低于 `2.4.0` 时，`test-compile` 失败的问题。

The error log:
https://github.com/seata/seata/actions/runs/3749476555/jobs/6367999677


slievrly (approved):
LGTM

PR #5193, opened 2022-12-22
optimize: upgrade hikari druid mysql to stable version

JavaLionLi:
update hikari druid mysql8 depends on stable version


xingfudeshi (approved):
LGTM

a364176773 (approved):
LGTM

slievrly:
What problem was solved?

slievrly (commented):
What problem was solved?

JavaLionLi:
> What problem was solved?

hikaricp 3.X has stopped maintenance. if jdk17 is used, version 5.X is required, while jdk8 corresponds to version 4.X.

dependencies/pom.xml
```diff
@@ -97,14 +97,14 @@
...
+        <druid.version>1.2.12</druid.version>
         <commons-dbcp2.version>2.7.0</commons-dbcp2.version>
-        <hikari.version>3.4.3</hikari.version>
+        <hikari.version>4.0.3</hikari.version>
         <!-- sql parser -->
         <antlr4.version>4.8</antlr4.version>
         <!-- for jdbc driver when package -->
         <mysql5.version>${mysql.version}</mysql5.version>
-        <mysql8.version>8.0.27</mysql8.version>
+        <mysql8.version>8.0.31</mysql8.version>
```
slievrly, line 107: I recommend revert mysql dependencies and keep the others. The AT mode has depend on mysql-driver.

JavaLionLi: MySQL is already version 8.0.34, and the device versions are both compatible. Is this a major issue?

slievrly: mysql driver compatibility is at the user DML level, not at the JDBC metadata level. The AT strongly depends on the mysql driver metadata, and simply upgrading without testing the AT transaction capability is not acceptable. For example, there is a problem with mysql datetime type returning timestamp  and  localDateTime Type in high and low versions of resultSet, which will affect the serialization of AT.

---

mysql driver的兼容是对用户DML层面的兼容不是对JDBC元数据层面的兼容。AT强依赖mysql driver的元数据，如果没有测试AT事务能力，只是单纯升级是无法接受的。比如存在mysql datetime类型在resultSet高低版本返回timestamp和localDateTime问题，这会对AT的序列化造成影响。

PR #5196, opened 2022-12-24
optimize：adjust too short length(possible error) of table field 'lock_table.row_key'

xssdpgy:
optimize(maybe bugfix)：adjust too short length(possible error) of table field 'lock_table.row_key'

adjust length of table field 'lock_table.row_key',
according to the method 'io.seata.core.lock.AbstractLocker#convertToLockDO(io.seata.core.lock.RowLock)', the field 'lock_table.row_key' is union by 'lock_table.resource_id','lock_table.table_name' and 'lock_table.pk'，
image
but the length setting of field 'lock_table.row_key' is too short ,so when insert into ‘lock_table’  and the length sum of the other three fields beyond  'VARCHAR(128)', error will occur.
image

### DIFF
--- a/changes/en-us/develop.md
+++ b/changes/en-us/develop.md
@@ -11,6 +11,7 @@ Add changes here for all PR submitted to the develop branch.
 
 ### optimize:
 - [[#xxx](https://github.com/seata/seata/pull/xxx)] optimize xxx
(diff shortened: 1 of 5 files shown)


a364176773:
row key is inherently user-adjustable and does not need to be justified at the framework level, as it varies from person to person

a364176773 (commented):
row key is inherently user-adjustable and does not need to be justified at the framework level, as it varies from person to person

PR #5307, opened 2023-02-06
bugfix: that keywords don't add escaped characters

### LINKED ISSUE #5291: The datasource proxy's problem
- [x]  I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
Seata's datasource proxy's sql checker has a problem, it'll replace the character '`'.

### Ⅱ. Describe what happened
There's a field named 'all' in our table. But all is a sql keyword. We use MyBatisPlus, I used put a @TableFiled("`` `all` ``") annotation to avoid the sql syntax error, and it works.
Today, I imorted seata to our project. When I execute the sql include field 'all', a SQLSyntaxErrorException has been threw. It gave a message: `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'all,xxx FROM t_access_record ' at line 1`. Seems, seata's datasource proxy replaced the characted '`'.
Normally, mybatis-plus don't replace the character. But it has dirty-reading problem. Seata resolve these distrubut-system's problem, but its datasource proxy is not ideal.

I want to have an option to decide whether the data source agent should replace some characters.

1. Create an table, contains a field named 'all'
2. Include MyBatis-Plus, put an anotation '@TableField("`` `all` ``")' to the field
3. Enable GlobalTransaction, and use the field to execute **update** statement.

No.

### Ⅵ. Environment:

- JDK version : 1.8
- Seata version:  1.6.1
- OS : Windows
- Others: IDEA 2021.1, SpringBoot 2.3.12.RELEASE

a364176773:

fixes #5291

### DIFF
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
@@ -124,23 +124,23 @@ public class UpdateExecutor<T, S extends Statement> extends AbstractDMLBaseExecu
         return selectSQLJoiner.toString();
     }
 
-    protected List<String> getNeedUpdateColumns(String table, String tableAlias, List<String> originUpdateColumns) {
+    protected List<String> getNeedUpdateColumns(String table, String tableAlias, List<String> unescapeUpdateColumns) {
         List<String> needUpdateColumns = new ArrayList<>();
         TableMeta tableMeta = getTableMeta(table);
         if (ONLY_CARE_UPDATE_COLUMNS) {
-            if (!containsPK(table, originUpdateColumns)) {
+            if (!containsPK(table, unescapeUpdateColumns)) {
                 List<String> pkNameList = tableMeta.getEscapePkNameList(getDbType());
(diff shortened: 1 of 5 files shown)


jsbxyyx (approved):
LGTM

rm-datasource/src/main/java/io/seata/rm/datasource/exec/UpdateExecutor.java
```diff
@@ -137,18 +134,19 @@ protected List<String> getNeedUpdateColumns(String table, String tableAlias, Lis
                     needUpdateColumns.add(getColumnNamesWithTablePrefix(table,tableAlias,pkNameList));
                 }
             }
-            needUpdateColumns.addAll(originUpdateColumns);
+            needUpdateColumns.addAll(originUpdateColumns.parallelStream()
```
slievrly, line 137: We want the variable names to be more formal, such as the originUpdateColumns variable followed by Unescape.

a364176773: done

slievrly (approved):
LGTM

PR #5145, opened 2022-12-12
bugfix: global session is always begin in saga mode

tuwenlin:
fixes #5045 
#5050 can not fix this bug,because in saga mode,when a global transactional end , TM report the global session status to TC rather than do commit or do rollback

PR #5464, opened 2023-03-22
bugfix: global session is always begin in saga mode for 2.x

tuwenlin:
fixes https://github.com/seata/seata/issues/5045
https://github.com/seata/seata/pull/5050 can not fix this bug,because in saga mode,when a global transactional end , TM report the global session status to TC rather than do commit or do rollback
this pr same to pr #5145,this pr just for 2.x


tuwenlin:
> Instead of modifying the develop.md file for the 2.x branch, Pls modify the 2.0.0.md.

done

slievrly (approved):
LGTM

server/src/main/java/io/seata/server/session/SessionHelper.java
```diff
@@ -134,6 +134,10 @@ public static void endCommitted(GlobalSession globalSession, boolean retryGlobal
             MetricsPublisher.postSessionDoneEvent(globalSession, IdConstants.STATUS_VALUE_AFTER_COMMITTED_KEY, true,
                 beginTime, retryBranch);
         } else {
+            if (globalSession.isSaga()) {
+                globalSession.setStatus(GlobalStatus.Committed);
```
a364176773, line 138: 默认的defaultcore也会调这里，也是会进行一个setstatus，不如把这行提出来，把defaultcore里的setstatus删了，这个setstatus放到if判断的前面

PR #5485, opened 2023-04-03
optimize: optimize server log output

slievrly:

optimize server log output

after modified in file log:
image


core/src/main/java/io/seata/core/protocol/RegisterTMRequest.java
```diff
@@ -87,11 +87,4 @@ public short getTypeCode() {
         return MessageType.TYPE_REG_CLT;
     }
 
-    @Override
-    public String toString() {
-        return "RegisterTMRequest{" +
-            "applicationId='" + applicationId + '\'' +
-            ", transactionServiceGroup='" + transactionServiceGroup + '\'' +
-            '}';
-    }
```
wangliang181230, line 96: Why remove this override?

slievrly: `RegisterTMRequest` has no additional filed and implements `toString()` in its parent class `AbstractIdentifyRequest`.

wangliang181230 (approved):
LGTM

PR #5647, opened 2023-06-14
optimize: support case-sensitive attributes for table and column metadata

slievrly:

support case-sensitive attributes for table and column metadata

fix #5553


rm-datasource/src/main/java/io/seata/rm/datasource/sql/struct/cache/MysqlTableMetaCache.java
```diff
@@ -106,6 +106,9 @@ private TableMeta resultSetMetaToSchema(ResultSetMetaData rsmd, DatabaseMetaData
 
         TableMeta tm = new TableMeta();
         tm.setTableName(tableName);
+        //always true and nothing to do with escape characters for mysql.
+        // May be not consistent with lower_case_table_names
+        tm.setCaseSensitive(true);
```
a364176773, line 111: Why not use SHOW VARIABLES LIKE 'lower_case_table_names'.  其他几个db同理

PR #5850, opened 2023-09-15
optimize: decouple saga expression handling and remove evaluator package

### LINKED ISSUE #5436: seata saga 模式expression计算和spring接耦
## Why you need it?

## How it could be?

ptyin:

The current expression structure (as observed following) is actually quite fragmented, mainly due to the fact that the expression structure processing logic is currently scattered across invocations without a centralized intermediary layer to parse it. 

image

Thus, this PR consolidate the parsing of expression structures and adopts the expression structure in the green box in the figure on a backward-compatible basis.

Besides, the differences between the expression and evaluation packages are minimal, as both are used to handle expressions, mainly differing in return values. In fact, they can both be unified under the expression category. Separating them into two systems actually increases developers' understanding cost of the code.

fixes #5436

This optimization is purely a code optimization, without adding any new functionality. It can be verified through existing saga tests. The expected behavior is the same as usual.

By existing saga tests.


saga/seata-saga-engine/src/main/java/io/seata/saga/engine/expression/impl/DefaultExpressionResolver.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private ExpressionFactoryManager expressionFactoryManager;
+
+    private String rootObjectName;
+
+    @Override
+    public Expression getExpression(String expressionStr) {
+        ExpressionStruct struct = parseExpressionStruct(expressionStr);
+
+        ExpressionFactory expressionFactory = expressionFactoryManager.getExpressionFactory(struct.type);
```
wt-better, line 55: we need to compatible history expressions without type. so if expression type is null, default expression type?

ptyin: It works fine with passing `null`

saga/seata-saga-engine/src/main/java/io/seata/saga/engine/expression/impl/DefaultExpressionResolver.java
```diff
@@ -0,0 +1,138 @@
...
+        ExpressionStruct struct = new ExpressionStruct();
+        struct.typeStart = expressionStr.indexOf("$");
+        int dot = expressionStr.indexOf(".", struct.typeStart);
+        int leftBracket = expressionStr.indexOf("{", struct.typeStart);
+
+
+        boolean isOldEvaluatorStyle = false;
+        if (struct.typeStart == 0) {
+            if (leftBracket < 0 && dot < 0) {
+                throw new IllegalArgumentException(String.format("Expression [%s] type is not closed", expressionStr));
```
wt-better, line 66: if express is '$=1', this message is inappropriate. Whether to modify it to 'resolve expression %s fail ' ?

ptyin: I think the message "resolve expression %s fail" you mentioned is kind of too generic, and may confuse users. And the current message does make sense in case of "$=1". The message would be "Expression [$=1] type is not closed", which can be tranlated as "表达式[$=1]类型未闭合". Since the the standard format of expression is `$expressionType.expressionContent`, this message indicates the `expressionType` is not closed, i.e., the type end indicator '.' is missing.

ptyin: Or, an optimized solution can be "Failed to resolve the expression [%s]: the type end indicator is missing." What do you think?

wt-better:
LGTM

saga/seata-saga-engine/src/main/java/io/seata/saga/engine/expression/impl/DefaultExpressionResolver.java
wt-better: I have priority the second proposal.

PR #5876, opened 2023-09-24
feature: add admin api used by seata-ctl

An-DJ:

Add admin API used by `seata-ctl`, which will be held under the `Seata` organization.

([temporary location](https://github.com/an-dj/seata-ctl))

#### Admin API list
* StatusController, for querying the TC server status (ok/ not ok) in a cluster
* RegistryController, for set/get operation to `seata.registry.*` configuration (in memory only)
* ConfigCenterController, for set/get operation to `seata.config.*` configuration (in memory only)
* ConfigController, for set/get operation to normal configuration
* TxnTryController, for trying an example transaction to check if the server is ok

#### Configurations which can be dynamically modified
* seata.server.max-commit-retry-timeout
* seata.server.max-rollback-retry-timeout
* seata.server.rollback-retry-timeout-unlock-enable
* seata.server.enable-check-auth
* seata.server.retry-dead-threshold
* seata.server.xaer-nota-retry-timeout
* seata.server.undo.log-save-days

#### In addition
1. Registry configuration (seata.registry.x) and config-center configuration (seata.config.x) can only be modified dynamically in memory. That is, if the server restarts, all changes of them will be lost.
2. Normal configuration (except for registry & config-center conf) can be obtained by `get`, but only several configuration item can be modified dynamically. This is because that the others are used by the server initialization (such as service.port), they cannot be easily modified if the server does not shutdown (or reboot).

#### Validate the configuration key-value pair

In this PR, ConfigurationValidator is introduced to validate if a configuration key-value pair is valid.

Configuraiton schema is defined in `config/seata-config-core/src/main/resources/config-schema.yml`.


An-DJ:
@slievrly Ready for your review: 

1. Support `get/set` multiple configurations at the same time, like below:

	```shell
	get config --key '[ \
	    "server.servicePort", \
	    "server.recovery.timeoutRetryPeriod", \
	    "server.undo.logDeletePeriod" \
	]'
	```

	```shell
	set config --data '{ \
	    "server.maxCommitRetryTimeout": "3000", \
	    "server.maxRollbackRetryTimeout": "3000", \
	    "server.undo.logSaveDays": "14" \
	}'
	```

2. Enhance the `try` command to simulating transction steps `begin`, `commit`, and `rollback`:

	* Try to begin an example transaction:

	```shell
	127.0.0.1:7091 > try begin --timeout 300000
	Try an example txn successfully, xid=192.168.163.1:8091:8755443813836259333
	```

	* Commit a transaction by xid:

	```shell
	127.0.0.1:7091 > try commit --xid 192.168.163.1:8091:8755443813836259333
	Commit txn successfully, xid=192.168.163.1:8091:8755443813836259333
	```

	* Rollback a transaction by xid:

	```shell
	127.0.0.1:7091 > try rollback --xid 192.168.163.1:8091:8755443813836259333
	Rollback txn successfully, xid=192.168.163.1:8091:8755443813836259333
	```

3. Support atomically modifying the configuration of `registry` and `conf center`

4. Use unified command entry to configuring `registry`, `conf center`, and `general configuration`.

	```shell
	127.0.0.1:7091 > set config -h
	Set the configuration

	Usage:
	   set config [flags]

	Flags:
	      --config-center   If set configuration center conf
	      --data string     Configuration map (default "{}")
	  -h, --help            help for config
	      --registry        If set registry conf
	```

	The three configurations differ in processing logic, so additional parameters are used to distinguish the purpose.

More examples can be found there: https://github.com/An-DJ/seata-ctl

BTW, is it suitable now to place that repo under the 'Seata' organization?

discovery/seata-discovery-core/src/main/java/io/seata/discovery/registry/RegistryService.java
```diff
@@ -95,6 +95,19 @@ public interface RegistryService<T> {
      */
     List<InetSocketAddress> lookup(String key) throws Exception;
 
+    /**
+     * Get all nodes' ip in that server's cluster
+     * @return ips
+     * @throws Exception the exception
+     */
+    List<InetSocketAddress> getClusterNodes() throws Exception;
```
slievrly, line 103: Looking at the various implementations, this interface I don't think is necessary.

server/src/main/java/io/seata/server/console/controller/ConfigurationController.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Junduo Dong
+ */
+@RestController
+@RequestMapping("/api/v1/admin/configuration")
+public class ConfigurationController {
```
slievrly, line 38: The version of the API should also be reflected in the file structure.

server/src/main/java/io/seata/server/console/controller/ConfigurationController.java
```diff
@@ -0,0 +1,104 @@
...
+                return SingleResult.success(configurationService.setConf(conf));
+            } catch (Exception e) {
+                return SingleResult.failure(Code.ERROR.code, e.getMessage());
+            }
+        }
+    }
+
+    @RequestMapping(method = RequestMethod.POST, value = "/registry")
+    SingleResult<Map<String, List<String>>> setRegistryConf(@RequestBody Map<String, String> conf) {
+        synchronized (setConfLock) {
```
slievrly, line 63: we need a global lock, not a process-level lock.

mark as TODO.

server/src/main/java/io/seata/server/console/controller/TxnTryController.java
```diff
@@ -0,0 +1,87 @@
...
+import static io.seata.server.console.Constant.DEMO_APP_NAME;
+import static io.seata.server.console.Constant.DEMO_GROUP_NAME;
+import static io.seata.server.console.Constant.DEMO_TXN_NAME;
+
+/**
+ * @author Junduo Dong
+ */
+@RestController
+@RequestMapping("/api/v1/admin/txn")
+public class TxnTryController {
```
slievrly, line 42: txn  -> trx
TxnTryController  ->  TxnMockController

An-DJ: Renamed.

server/src/main/java/io/seata/server/console/controller/ConfigurationController.java
An-DJ: Done.

discovery/seata-discovery-core/src/main/java/io/seata/discovery/registry/RegistryService.java
An-DJ: Although the core logic in `getClusterNodes()` is calling the private method `lookupByCluster`, which the same as an  existing function `lookup`, but `lookup` is exposed to be used in CLIENT side.

So IMHO, we need expose a method to be used in SERVER side. That's why I introduce the `getClusterNodes`.

slievrly (approved):
LGTM

discovery/seata-discovery-core/src/main/java/io/seata/discovery/registry/RegistryService.java
slievrly: ok.

PR #5892, opened 2023-09-28
feature: support polardb-x 2.0

hsien999:

Extending AT mode support for PolarDB-X 2.0.
Note: Transparent PolarDB-X support is currently available only in the form of MySQL inheritance, and DRDS partitioning may need to be supported in the future.


rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -101,6 +102,20 @@ private void init(DataSource dataSource, String resourceGroupId) {
         RootContext.setDefaultBranchType(this.getBranchType());
     }
 
+    /**
+     * test mysql adaptive type for PolarDB-X
+     *
+     * @param connection db connection
+     */
+    private void testMySQLAdaptiveType(Connection connection) {
```
slievrly, line 110: Method names need to be more standardized, generally only UT will use the test name.

sqlparser/seata-sqlparser-druid/src/main/java/io/seata/sqlparser/druid/DruidSQLRecognizerFactoryImpl.java
```diff
@@ -68,4 +69,17 @@ public List<SQLRecognizer> create(String sql, String dbType) {
...
         return recognizers;
     }
+
+    /**
+     * Get adaptive db type for druid parser.
+     *
+     * @param dbType origin db type
+     * @return adaptive db type
+     */
+    private static String getAdaptiveDbType(String dbType) {
```
slievrly, line 79: expect a common class to determine if it's a MySQL family of databases.

slievrly (commented):
Pls. add change in:
https://github.com/seata/seata/blob/develop/changes/en-us/develop.md
https://github.com/seata/seata/blob/develop/changes/zh-cn/develop.md

slievrly (approved):
LGTM

PR #5900, opened 2023-10-05
docs: add docs about Seata configuration files

### LINKED ISSUE #5608: confused about the documentation [important]
I think there do not have a clear describe regard to seata component and their respective configration.
Yea, you guys wrote a documentation but it just describe every option's mean.
Yea, you guys impletment various of config and registy center but it's not clearly about their's usage.
But I do not recomment wrote a documentaion all in one. 

Let's have a look about what happen
Fristly I download a seata server package and found there have 2 files about the configruation.
One is conf/application.yaml which one I think is a configuration for seata sever seems like our springcloud's application.yml.
One is script/config-center/config.txt which one is refer to [seata offical documentation](https://seata.io/zh-cn/docs/user/configurations.html)

question 1:
参数同步到配置中心使用demo
I did it.
CleanShot 2023-05-25 at 10 32 08
And then It looks strange. My first impression of this is something maybe really wrong. but  ,I don't know.
question 2:
There do not have a documentation about application.yaml Just have a example file .
`seata.config.nacos.data-id` for what? 
Does this prove someting about question one ,I don't know.
```
seata:
  config:
    # support: nacos 、 consul 、 apollo 、 zk  、 etcd3
    type: file
    nacos:
      server-addr: 127.0.0.1:8848
      namespace:
      group: SEATA_GROUP
      username:
      password:
      context-path:
      ##if use MSE Nacos with auth, mutex with username/password attribute
      #access-key: ""
      #secret-key: ""
      data-id: seataServer.properties
```
I'm still confused about where this configuration should be placed,and when they will be loaded and use.

meteor-01:

Add comments on logs and parameters related to nacos as the configuration center and registry

fixes #5608


slievrly (commented):
There is a lack of configuration item description except Nacos. Is there any plan to supplement it?Is it better to merge with application.example.yml here?

slievrly:
CLA check failed.

PR #5918, opened 2023-10-09
optimize: standardized the properties of codecov.yml

l81893521:

standardized the properties of codecov.yml

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,24 +2,25 @@ codecov:
   require_ci_to_pass: yes
 coverage:
   status:
-    patch: no
+    patch: off
+    changes: off
     project:
       default:
         threshold: 1%
         if_not_found: success
-    changes: no
+  round: down
   precision: 2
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -2,24 +2,25 @@ codecov:
...
   precision: 2
-  range: "50...100"
+  range: "50..75"
 ignore:
- - "test/.*"
- - ".github/.*"
- - ".mvn/.*"
- - ".style/.*"
- - "*.md"
+ - "**/test"
```
slievrly, line 15: more upper folder?

Does this include the following cases?

server/test/xxx

l81893521: yes, this rule will ignore any folders and files that call `test`. The example: https://docs.codecov.com/docs/ignoring-paths#ignoring-specific-file-names

slievrly (approved):
LGTM

Issue #5923, opened 2023-10-11
zk unit test error


l81893521:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
<img width="1471" alt="image" src="https://github.com/seata/seata/assets/4516752/772431a8-565c-409a-b14e-f738f8bcb0ba">

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

Minimal yet complete reproducer code (or URL to code):

### Ⅵ. Environment:

- JDK version(e.g. `java -version`):
- Seata client/server version:  
- Database version:
- OS(e.g. `uname -a`):
- Others:

PR #5924, opened 2023-10-11
bugfix: fix zookeeper unit test error

l81893521:
remove the method `buildZkTest` because the logic already cover in `testAll`

#5923

### DIFF
--- a/discovery/seata-discovery-zk/src/test/java/io/seata/discovery/registry/zk/ZookeeperRegisterServiceImplTest.java
+++ b/discovery/seata-discovery-zk/src/test/java/io/seata/discovery/registry/zk/ZookeeperRegisterServiceImplTest.java
@@ -60,12 +60,6 @@ public class ZookeeperRegisterServiceImplTest {
     }
 
     @Test
-    public void buildZkTest() {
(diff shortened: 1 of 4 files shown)


discovery/seata-discovery-zk/src/test/java/io/seata/discovery/registry/zk/ZookeeperRegisterServiceImplTest.java
```diff
@@ -59,12 +59,6 @@ public void getInstance() {
         Assertions.assertEquals(service1, service);
     }
 
-    @Test
```
funky-eyes, line 62: Why should this test case be deleted? If it must be deleted, why did the previous test case pass?

l81893521: because the test case in `ZookeeperRegisterServiceImplTest` would cause some problem when parallel execution. And the have cover the same code so if delete one of them it would not cause the coverage decrease.

funky-eyes (approved):
LGTM，please sync to 2.x branch

slievrly (approved):
LGTM

Issue #5904, opened 2023-10-06
SAGA state machine definition lacks syntax and semantic parsing


ptyin:
## Why you need it?
> Is your feature request related to a problem? Please describe in details

1. JSON syntax is not strictly defined currently, which leads to unclarified usage. 
2. No semantic verification is performed in parsing phase. Some errors that can be checked during parsing are not throwed until the state machine is run.

## How it could be?
> A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

1. Strict JSON shcema definitions. 
    Beyond that, JSON schema offers syntax highlight, code completion, and other functionalities achieved by popular editors like vscode .
2. Semantic verification.

## Other related information
> Add any other context or screenshots about the feature request here.

I can implement.

wt-better:
Yes, it`s good idea.

PR #5928, opened 2023-10-12
optimize: add saga statelang semantic validation

ptyin:

To avoid throwing exception in Saga runtime, I added Saga statelang semantic validation phase. Validation rules include:
- `FiniteTerminationRule`: Rule to check if the state machine can terminate in finite time, i.e. if there is an infinite loop.
- `NoRecursiveSubStateMachineRule`: Rule to check if all SubStateMachines has no recursive call.
- `StateNameExistsRule`: Rule to check if all the state names exist.

Users can add more validation rules using SPI.

Related to #5904 

Notes for `FiniteTerminationRule`: the main idea is to find all cycles in the graph. And if there is a cycle without outgoing flow, it means the state machine can definitely not stop.

---

`FiniteTerminationRule`的备注：核心思想是找到图中的所有环。如果有任何一个环没有连向非环的出边，则意味着状态机肯定不能停止。


slievrly (approved):
good job.

1.Rule is implemented by SPI
2. Abstract verification rules
3. Implement common rules

slievrly:
@ptyin add change in 2.0.0.md.

ptyin:
> @ptyin add change in 2.0.0.md.

done

PR #5952, opened 2023-10-24
bugfix: fix zk unit test

l81893521:

remove the method `buildZkTest` because the test case in `ZookeeperRegisterServiceImplTest` would cause some problem when parallel execution. And they cover the same code so if delete one of them it would not cause the coverage decrease.

#5923


slievrly:
It seems that the fix for #5924 still does not solve the problem.

slievrly (commented):
maybe fixed in #5960 , close this.

PR #5966, opened 2023-10-27
optimize: decouple saga expression handling and remove evaluator package

### LINKED ISSUE #5436: seata saga 模式expression计算和spring接耦

ptyin:

The current expression structure (as observed following) is actually quite fragmented, mainly due to the fact that the expression structure processing logic is currently scattered across invocations without a centralized intermediary layer to parse it. 

image

Thus, this PR consolidate the parsing of expression structures and adopts the expression structure in the green box in the figure on a backward-compatible basis.

Besides, the differences between the expression and evaluation packages are minimal, as both are used to handle expressions, mainly differing in return values. In fact, they can both be unified under the expression category. Separating them into two systems actually increases developers' understanding cost of the code.

fixes #5436

This optimization is purely a code optimization, without adding any new functionality. It can be verified through existing saga tests. The expected behavior is the same as usual.

By existing saga tests.

Same as pr #5850


slievrly (approved):
LGTM

PR #5977, opened 2023-10-30
bugfix: fix that rpcServer is not closed when raftServer is closed

### LINKED ISSUE #5976: raftserver in SessionHolder  can not be destroyed correctly
https://github.com/seata/seata/blob/64c5b8a915845355ed347063a4be0a2c59b28c6b/server/src/main/java/io/seata/server/session/SessionHolder.java#L386

```
        this.raftGroupService = new RaftGroupService(groupId, serverId, nodeOptions, rpcServer, true);

```
because of sharedRpcServer, when com.alipay.sofa.jraft.RaftGroupService#shutdown , the netty port will be used and not closed. 

```
public synchronized void shutdown() {
        if (!this.started) {
            return;
        }
        if (this.rpcServer != null) {
            try {
                if (!this.sharedRpcServer) {
                    this.rpcServer.shutdown();
                }
            } catch (final Exception ignored) {
                // ignore
            }
            this.rpcServer = null;
        }
        this.node.shutdown();
        NodeManager.getInstance().removeAddress(this.serverId.getEndpoint());
        this.started = false;
        LOG.info("Stop the RaftGroupService successfully.");
    }

```

funky-eyes:

fixes #5976


server/src/main/java/io/seata/server/session/SessionHolder.java
```diff
@@ -380,14 +379,7 @@ public static boolean distributedLockAndExecute(String key, NoArgsFunc func) {
...
     public static void destroy() {
-        Collection<RaftServer> raftServers = RaftServerFactory.getInstance().getRaftServers();
-        if (raftServers != null) {
-            try {
-                raftServers.forEach(RaftServer::close);
-            } catch (Exception e) {
-                LOGGER.error(e.getMessage());
-            }
-        }
+        RaftServerFactory.getInstance().destroy();
```
slievrly, line 382: In the future, it is necessary to refactor the code to decouple SessionHolder with RaftXXX.

server/src/main/java/io/seata/server/cluster/raft/RaftServerFactory.java
```diff
@@ -155,6 +157,20 @@ public void start() {
         }
     }
 
+    @Override
+    public void destroy() {
+        this.close();
+    }
+
+    @Override
+    public void close() {
```
slievrly, line 166: The general purpose of the Java Factory class is to create and return objects of other classes, not to control behavior. If rpcServer is part of RaftServer, then it is more appropriate to control behavior in RaftServer.

server/src/main/java/io/seata/server/session/SessionHolder.java
funky-eyes: > In the future, it is necessary to refactor the code to decouple SessionHolder with RaftXXX.

Since the raft cluster must match the raft mode exactly, it must be written in the sessionHolder, otherwise the timing of the initialization is not very controllable, and the raft cluster must be started after the sessionmanager has finished initializing, otherwise there may be npe issues

server/src/main/java/io/seata/server/cluster/raft/RaftServerFactory.java
funky-eyes: > The general purpose of the Java Factory class is to create and return objects of other classes, not to control behavior. If rpcServer is part of RaftServer, then it is more appropriate to control behavior in RaftServer.

A single rpc server can be used by multiple raft servers, so all raft servers should be controlled in the raft server factory, as well as their behavior when destroyed.

leizhiyuan (approved):
LGTM

Issue #5936, opened 2023-10-17
redis主从复制，lua脚本script load后主节点并没有同步至从节点导致的异常问题


PeppaO:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
redis主从复制，lua脚本script load后主节点并没有同步至从节点导致的异常问题

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```
redis.clients.jedis.exceptions.JedisNoScriptException: NOSCRIPT No matching script. Please use EVAL.
	at redis.clients.jedis.Protocol.processError(Protocol.java:130)
	at redis.clients.jedis.Protocol.process(Protocol.java:166)
	at redis.clients.jedis.Protocol.read(Protocol.java:220)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:318)
	at redis.clients.jedis.Connection.getOne(Connection.java:296)
	at redis.clients.jedis.Jedis.getEvalResult(Jedis.java:2879)
	at redis.clients.jedis.Jedis.evalsha(Jedis.java:2907)
	at redis.clients.jedis.Jedis.evalsha(Jedis.java:2900)

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

Minimal yet complete reproducer code (or URL to code):

### Ⅵ. Environment:

- JDK version(e.g. `java -version`):
- Seata client/server version:   1.6
- Database version:
- OS(e.g. `uname -a`):
- Others:

PR #5990, opened 2023-11-02
bugfix: fix the issue that the Lua script is not synchronized when the redis sentinel master node is down

PeppaO:

redis主从复制，当主节点宕机后，从节点选举为主，当遇到NOSCRIPT异常时，重新script load一遍

#5936


server/src/main/java/io/seata/server/storage/redis/LuaParser.java
```diff
@@ -31,20 +32,26 @@
...
  * @author conghuhu
  */
 public class LuaParser {
+    private static final Logger LOGGER = LoggerFactory.getLogger(LuaParser.class);
 
     private static final String WHITE_SPACE = " ";
 
     private static final String ANNOTATION_LUA = "--";
 
+    private static  Map<String, String> LUA_FILE_MAP = new HashMap<>();
```
funky-eyes: static final

funky-eyes (commented):
todo：之所以把lua脚本外置，而不是在源码中，是为了发现lua脚本执行有bug的时候可以人工介入修正脚本，重新进行scriptload，但是目前看来这是个伪需求，等待真实需求的用户出现后，再进行重新记录sha1编码的行为也不迟.目前来看，这个pr的处理已经足够了，即便存在线程安全的问题，如scriptload可能会执行多次，但是对于执行结果而言没有影响，且master节点宕机概率并不会太高，没有必要做到分布式下的线程安全，可以考虑下单节点是否避免多次scriptload，以lua文件名为粒度做锁
The reason why the lua script external, rather than in the source code, is to find the lua script execution has a bug when you can manually intervene to correct the script, re-scriptload, but at present it seems that this is a pseudo-need to follow up on the real needs of users can wait for the emergence of users, and then re-record the optimization of sha1 encoding is not too late. At present, this pr processing is already enough, even if there are thread-safety issues, such as scriptload may be executed more than once, but for the implementation of the results have no impact, and the probability of downtime of the master node will not be too high, there is no need to do the distributed thread-safety, you can consider whether the next single node to avoid multiple scriptload, to lua file name as the granularity of the lock

server/src/main/java/io/seata/server/storage/redis/LuaParser.java
```diff
@@ -149,4 +157,14 @@ public static <T> List<T> getListFromJson(String json, Class<T> classz) {
             throw new StoreException(e.getMessage());
         }
     }
+
+    public static Object jedisEvalSha(Jedis jedis, String luaSHA, String luaFileName, List<String> keys, List<String> args){
+        try {
+            return jedis.evalsha(luaSHA, keys, args);
+        }catch (JedisNoScriptException e) {
+            LOGGER.warn("jedis ex: " + e.getMessage());
```
funky-eyes: 这里日志要提示一下，否则用户会认为异常了，但实际上我们继续了reload script和重新执行了lua脚本
Here the logs should be prompted, otherwise the user will think it's an exception, but in fact we continue the reload script and re-execute the lua script.

PeppaO: done

server/src/main/java/io/seata/server/storage/redis/LuaParser.java
PeppaO: done

server/src/main/java/io/seata/server/storage/redis/LuaParser.java
```diff
@@ -149,4 +157,14 @@ public static <T> List<T> getListFromJson(String json, Class<T> classz) {
             throw new StoreException(e.getMessage());
         }
     }
+
+    public static Object jedisEvalSha(Jedis jedis, String luaSHA, String luaFileName, List<String> keys, List<String> args) {
+        try {
+            return jedis.evalsha(luaSHA, keys, args);
+        } catch (JedisNoScriptException e) {
```
leizhiyuan, line 164: Can this handling method be applied to other Redis exceptions as well, not just limited to JedisNoScriptException?

PeppaO: other redis exceptions do not require script load

leizhiyuan (approved):
LGTM

funky-eyes (approved):
LGTM

PR #6018, opened 2023-11-11
bugfix: fix incorrect metric report

### LINKED ISSUE #6016: Duplicate commit event publishing causes incorrect transaction metric

### Ⅰ. Issue Description

There is a **committed event publishing** in ll.169 of `DefaultCore#commit`.

https://github.com/seata/seata/blob/8192138895ada3cc7cd09977f3219ed8371c17ae/server/src/main/java/io/seata/server/coordinator/DefaultCore.java#L169

However, there is another **committed event publishing** in ll.150 of `SessionHelper#endCommitted`,

https://github.com/seata/seata/blob/8192138895ada3cc7cd09977f3219ed8371c17ae/server/src/main/java/io/seata/server/session/SessionHelper.java#L140-L161

which leads to calling `MetricsSubscriber#processGlobalStatusCommitted` twice, i.e., a double deduction on the `active` metric and a double increasement on the `committed` metric.

image

### Ⅵ. Environment:

- Database version:
- OS(e.g. `uname -a`):
- Others:

ptyin:

Fix incorrect metric report. 

It's actually a accidental code error. The conditions in

https://github.com/seata/seata/blob/8192138895ada3cc7cd09977f3219ed8371c17ae/server/src/main/java/io/seata/server/session/SessionHelper.java#L149

and in

https://github.com/seata/seata/blob/8192138895ada3cc7cd09977f3219ed8371c17ae/server/src/main/java/io/seata/server/session/SessionHelper.java#L221

intend to distinguish whether `retryGlobal` is true or false. Therefore, `retryGlobal` should be evaluated instead of `DELAY_HANDLE_SESSION`.

fixes #6016

### DIFF
--- a/server/src/main/java/io/seata/server/session/SessionHelper.java
+++ b/server/src/main/java/io/seata/server/session/SessionHelper.java
@@ -147,7 +147,7 @@ public class SessionHelper {
             }
             globalSession.end();
             if (!DELAY_HANDLE_SESSION) {
-                MetricsPublisher.postSessionDoneEvent(globalSession, false, false);
+                MetricsPublisher.postSessionDoneEvent(globalSession, retryGlobal, false);
             }
             MetricsPublisher.postSessionDoneEvent(globalSession, IdConstants.STATUS_VALUE_AFTER_COMMITTED_KEY, true,
                 beginTime, retryBranch);
@@ -219,7 +219,7 @@ public class SessionHelper {
             }
             globalSession.end();
             if (!DELAY_HANDLE_SESSION && !timeoutDone) {
-                MetricsPublisher.postSessionDoneEvent(globalSession, false, false);
+                MetricsPublisher.postSessionDoneEvent(globalSession, retryGlobal, false);
             }
             MetricsPublisher.postSessionDoneEvent(globalSession, IdConstants.STATUS_VALUE_AFTER_ROLLBACKED_KEY, true,
                     beginTime, retryBranch);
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/seata/server/session/SessionHelper.java
```diff
@@ -146,7 +146,7 @@ public static void endCommitted(GlobalSession globalSession, boolean retryGlobal
                 globalSession.changeGlobalStatus(GlobalStatus.Committed);
             }
             globalSession.end();
-            if (!DELAY_HANDLE_SESSION) {
```
funky-eyes, line 149: file和raft模式时,外层的if是不成立的,所以需要在这里打end的点和after的点,而其它模式在二阶段决议时是走的else逻辑,committed已经在二阶段决议时完成了,所以在这不需要再打点
file and raft mode, the outer if is not established, so you need to hit the end of the point here and after the point, while the other modes in the two-stage resolution is to go to the else logic, committed has been completed in the two-stage resolution, so no need to hit the point here!

ptyin: Got it. Leave this line unchanged, and modify the `MetricsPublisher.postSessionDoneEvent(globalSession, false, false);` to `MetricsPublisher.postSessionDoneEvent(globalSession, retryGlobal, false);` in the inner block. I think that makes more sense.

funky-eyes (approved):
LGTM

PR #6031, opened 2023-11-14
optimize: undo_log table check optimize

### LINKED ISSUE #6022: [Optimize] add a check for the existence of the undolog table
## Why you need it?
When proxying the datasourceProxy, a check is performed to see if the undolog table exists. If it doesn't exist, an exception is thrown as a prompt. Remember to add a check switch, as it's possible that a user may only use the TCC mode but inadvertently enable the check switch.

## How it could be?

laywin:

https://github.com/seata/seata/issues/6022

fixes #6022

### DIFF
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -21,5 +21,6 @@
 
 <!-- 请确保您的 GitHub ID 在以下列表中 -->
 - [slievrly](https://github.com/slievrly)
+- [laywin](https://github.com/laywin)
 
 同时，我们收到了社区反馈的很多有价值的issue和建议，非常感谢大家。
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -9,7 +9,7 @@
 - [[#PR_NO](https://github.com/seata/seata/pull/PR_NO)] 准确简要的PR描述
 
 ### optimize:
-- [[#PR_NO](https://github.com/seata/seata/pull/PR_NO)] 准确简要的PR描述
+- [[#6031](https://github.com/seata/seata/pull/6031)] 添加undo_log表的存在性校验
 
 ### security:
 - [[#PR_NO](https://github.com/seata/seata/pull/PR_NO)] 准确简要的PR描述
--- a/rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
+++ b/rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
@@ -24,6 +24,7 @@ import javax.sql.DataSource;
 
 import io.seata.common.ConfigurationKeys;
 import io.seata.common.Constants;
+import io.seata.common.loader.EnhancedServiceNotFoundException;
 import io.seata.config.ConfigurationFactory;
 import io.seata.core.constants.DBType;
 import io.seata.core.context.RootContext;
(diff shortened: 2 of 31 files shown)


rm-datasource/src/test/java/io/seata/rm/RMHandlerATTest.java
funky-eyes, line 1: Write a test case for the newly modified code.

laywin: done

changes/zh-cn/2.x.md
```diff
@@ -9,7 +9,7 @@
 - [[#PR_NO](https://github.com/seata/seata/pull/PR_NO)] 准确简要的PR描述
 
 ### optimize:
-- [[#PR_NO](https://github.com/seata/seata/pull/PR_NO)] 准确简要的PR描述
+- [[#6031](https://github.com/seata/seata/pull/6031)] 添加undo_log表的存在性校验
```
funky-eyes, line 12: 该版本下首次提交pr，作者信息需要进行登记一下
For the first PR submission in this version, the author's information needs to be registered.

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -143,6 +153,32 @@ private boolean isPolardbXProduct() {
...
+     *
+     * @param conn db connection
+     */
+    private void checkUndoLogTableExist(Connection conn) {
+        if (DBType.optionalof(dbType).isPresent()) {
+            UndoLogManager undoLogManager;
+            try {
+                undoLogManager = UndoLogManagerFactory.getUndoLogManager(dbType);
+            } catch (EnhancedServiceNotFoundException e) {
+                LOGGER.error("can't find undoLogManager service provider for dbtype: {}", dbType);
```
funky-eyes: 我建议这里直接抛异常出去，说明用户再使用一个我们不支持的数据库，要提示这个数据库不支持
I suggest throwing an exception directly here, indicating that if the user uses a database that we do not support, we need to prompt that the database is not supported.

changes/zh-cn/2.x.md
laywin: 是这个吗，有登记额
Is this it? It's registered
image

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
laywin: 没有抛出去，是因为很多单测使用了数据库代理，会导致单测失败。
It is not thrown out because many unit tests use database proxy, which can be lead it fail.

funky-eyes: > 没有抛出去，是因为很多单测使用了数据库代理，会导致单测失败。 It is not thrown out because many unit tests use database proxy, which can be lead it fail.

有没有什么办法在单测的时候不进行校验undolog table？或者增加一个开关，这个开关允许用户配置check，如果用户关闭，就不进行check，然后我们默认开启，在单测的时候关闭check
Is there any way to not check undolog table during single test or add a switch that allows user to configure check and if user turns it off, no check is performed, then we turn it on by default and turn off check during single test

changes/zh-cn/2.x.md
funky-eyes: 你看下中文的那份文件是否没登记
Check if the Chinese file is not registered.

laywin: done

rm-datasource/src/main/java/io/seata/rm/datasource/DataSourceProxy.java
laywin: update mockito version to 4.5.1 same as seata server's mockito dependency to support static mock. when unit test. mock undolog table check.

funky-eyes (approved):
LGTM

PR #6111, opened 2023-12-08
optimize: it is better to handle tc server offline message from registry center directly instead of 10s polling

laywin:

https://github.com/seata/seata/issues/6109

https://github.com/seata/seata/issues/6109


funky-eyes (commented):
没有必要写的如此复杂,只需要在每个注册中心的监听实现中,将下线节点的变更同步至`CURRENT_ADDRESS_MAP`中`group`对应的`list`即可
There's no need to write it so complicated. It's only necessary to synchronize the changes of the offline nodes to the corresponding 'list' in the CURRENT_ADDRESS_MAP within the listener implementation of each registration center.

PR #6166, opened 2023-12-17
optimize: nettyPoolKey creation and update logic

### LINKED ISSUE #6142: TMClient's nettyClientKeyPool doesn't work as it should

### Ⅰ. Issue Description
在NettyClientChannelManager的doConnect方法中，每次都会创建一个新的NettyPoolKey。如果是TMClient的话，会直接调用poolKeyMap.put而不是像RMClient一样的poolKeyMap.putIfAbsent，这会导致每次poolKeyMap中serverAddress对应的NettyPoolKey都会被更新。而下面的nettyClientKeyPool.borrowObject每次都会使用到这一更新后的NettyPoolKey作为key去池子中获取对象，这相当于每次都使用一个新的池子来新建对象，丧失了缓存池的意义

---

In the doConnect method of the NettyClientChannelManager, a new NettyPoolKey is created each time. if it is a TMClient, it will directly call poolKeyMap.put instead of poolKeyMap.putIfAbsent like RMClient. putIfAbsent, which will result in the NettyPoolKey corresponding to the serverAddress in the poolKeyMap being updated every time. The following nettyClientKeyPool.borrowObject will use this updated NettyPoolKey as the key to get the object in the pool every time, which is equivalent to using a new pool to create new objects every time, which loses the meaning of the cache pool.

image

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- Database version:
- OS(e.g. `uname -a`):
- Others:

PleaseGiveMeTheCoke:

1. 修复了 tmclient 的 nettyPoolKey 会在每次doConnect时新建导致对象池重复扩建的问题
2. 独立了tm和rm的 nettyPoolKey 更新逻辑，作为构造参数传给 NettyClientChannelManager

----

1. Fixed the problem that nettyPoolKey of tmclient will be newly created in every doConnect, resulting in duplicate expansion of the object pool.
2. Separate the nettyPoolKey update logic for tm and rm, and pass it to NettyClientChannelManager as a constructor parameter.

fixes #6142


slievrly (commented):
This fix solution is complex, and not all problems need to be solved by adding parameters. I believe there are simpler solutions. I suggest discussing the approach in the issue before submitting a PR.

PleaseGiveMeTheCoke:
> This fix solution is complex, and not all problems need to be solved by adding parameters. I believe there are simpler solutions. I suggest discussing the approach in the issue before submitting a PR.

Ok, I'll think about a simpler solution and discuss it in issue

slievrly (approved):
LGTM

core/src/main/java/io/seata/core/rpc/netty/NettyClientChannelManager.java
```diff
@@ -253,16 +251,8 @@ private Channel doConnect(String serverAddress) {
...
-            } else {
-                NettyPoolKey previousPoolKey = poolKeyMap.putIfAbsent(serverAddress, currentPoolKey);
-                if (previousPoolKey != null && previousPoolKey.getMessage() instanceof RegisterRMRequest) {
-                    RegisterRMRequest registerRMRequest = (RegisterRMRequest) currentPoolKey.getMessage();
-                    ((RegisterRMRequest) previousPoolKey.getMessage()).setResourceIds(registerRMRequest.getResourceIds());
-                }
-            }
-            channelFromPool = nettyClientKeyPool.borrowObject(poolKeyMap.get(serverAddress));
+            poolKeyMap.put(serverAddress, currentPoolKey);
+            channelFromPool = nettyClientKeyPool.borrowObject(currentPoolKey);
```
funky-eyes, line 255:
```
                    RegisterRMRequest registerRMRequest = (RegisterRMRequest) currentPoolKey.getMessage();
                    ((RegisterRMRequest) previousPoolKey.getMessage()).setResourceIds(registerRMRequest.getResourceIds());

```
这段代码去除后,用户如果使用到动态数据源,或者tcc上使用api动态的在运行中增删resource该如何解决这个问题?
After this code is removed, if the user uses a dynamic data source, or uses the API on the tcc to dynamically add or delete resources during operation, how can this problem be solved?

funky-eyes: 是我看错了,以为currentPoolKey是previousPoolKey 
I read it wrong and thought currentPoolKey was previousPoolKey.

funky-eyes (approved):
LGTM

PR #6209, opened 2023-12-25
optimize : Eliminate RpcMessage and Encoder/Decoder dependencies

### LINKED ISSUE #6210: Protocol layer strongly relies on rpcmessage, which has implications for multi-version protocol adaptation.
## Why you need it?
encoder/decoder强依赖rpcmessage，在多版本协议改造场景里相当于要求每个版本的rpcmessage保持高度一致，这是不太可能的，为此需要处理他们的依赖关系，让协议解析层（encoder/decoder）的处理和rpcmessage无关
encoder/decoder strongly depends on rpcmessage, which in a multi-version protocol transformation scenario is equivalent to requiring a high degree of consistency between each version of rpcmessage, which is unlikely, and for this reason their dependency needs to be handled in a way that makes the protocol parsing layer's (encoder/decoder's) processing independent of rpcmessage

## How it could be?

Bughue:

新引入ProtocolRpcMessage（接口），用以解开rpcmessage和encoder/decoder的依赖关系。encoder/decoder只依赖ProtocolRpcMessage（支持不同版本协议的区分），而原有的RpcMessage继续服务于协议解析后的业务逻辑
A new ProtocolRpcMessage interface is introduced to decouple rpcmessage from encoder/decoder, which only relies on ProtocolRpcMessage (to support differentiation between different versions of the protocol), while the original RpcMessage continues to serve the The original RpcMessage continues to serve the business logic after the protocol is parsed.

fixes https://github.com/apache/incubator-seata/issues/6210


core/src/main/java/io/seata/core/protocol/RpcMessage.java
```diff
@@ -169,6 +171,14 @@ public void setMessageType(byte messageType) {
         this.messageType = messageType;
     }
 
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
```
funky-eyes: client也会使用到rpcmessage，为什么不让新版的客户端直接传递当前的协议版本号？
The client will also use rpcmessage, why not let the new version of the client side directly pass the current protocol version number?

funky-eyes (approved):
LGTM

PR #6302, opened 2024-01-25
optimize: add io.seata package shade

slievrly:

add io.seata package shade


compatible/pom.xml
```diff
@@ -30,7 +30,7 @@
 
     <dependencies>
         <dependency>
-            <groupId>${project.groupId}</groupId>
+            <groupId>org.apache.seata</groupId>
```
funky-eyes: 如果他不依赖org.apache.seata:seata-all,会影响应用使用吗?假设我使用一个io.seata:seata-all 2.1.0-SNAPSHOT
If he doesn't depend on org.apache.seata:seata-all, will it affect the application usage? Let's say I use an io.seata:seata-all 2.1.0-SNAPSHOT

funky-eyes: In my judgment, if the compatible module doesn't depend on org.apache.seata:seata-all, then it's not complete and users will have problems using it because of the missing classes. If he's missing classes, what's the difference between that and just using io.seata:seata-all and typing a version specifically to be used as a usedby tag?
根据我的判断，如果兼容模块不依赖于org. apache.seata：seata-all，那么它就不完整，用户会因为缺少类而在使用它时遇到问题。如果他缺少类，这与仅使用io.seata：seata-all并输入专门用作usedby标记的版本有什么区别？

funky-eyes:
image

compatible/pom.xml
```diff
@@ -30,7 +30,7 @@
```
funky-eyes:
```suggestion
    <properties>
        <maven.compiler.source>8</maven.compiler.source>
        <maven.compiler.target>8</maven.compiler.target>
    </properties>
```

compatible/pom.xml
slievrly: io.seata:seata-all won't release module package.

funky-eyes (approved):
LGTM

PR #6315, opened 2024-01-30
optimize: compatible with lower versions of SPI

### LINKED ISSUE #6307: Compatible with lower versions of SPI
## Why you need it?

When the user upgrades from io.seata to org.apache.seata and the user implements a lower version of the io.seata.xxx interface, the user's SPI implementation needs to be compatible without the user having to change any code.

## How it could be?

xingfudeshi:

fixes #6307


compatible/pom.xml
```diff
@@ -20,33 +20,52 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <groupId>io.seata</groupId>
-    <version>2.1.0-SNAPSHOT</version>
+    <parent>
+        <groupId>org.apache.seata</groupId>
+        <artifactId>seata-parent</artifactId>
+        <version>${revision}</version>
```
slievrly: This place can't be changed to any other package name.The original used by data is required to be retained.

common/src/main/java/org/apache/seata/common/loader/EnhancedServiceLoader.java
```diff
@@ -43,13 +43,25 @@
...
 
     /**
      * Class->InnerEnhancedServiceLoader map
      */
     private static final ConcurrentMap<Class<?>, InnerEnhancedServiceLoader<?>> SERVICE_LOADERS =
             new ConcurrentHashMap<>();
 
+    private static <S> Class<S> getCompatibleService(Class<S> originService) {
+        String apacheSeataName = originService.getName();
+        String ioSeataName = apacheSeataName.replace(APACHE_SEATA_PACKAGE_NAME, IO_SEATA_PACKAGE_NAME);
```
slievrly, line 58: Explicitly identify the Apache Seata SPI using sartwith or other methods.

slievrly: If both exist, what is the priority?

compatible/src/main/java/io/seata/core/compressor/Compressor.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.seata.core.compressor;
+
+/**
+ * The type Compressor
+ *
+ */
+@Deprecated
```
slievrly: Compatible modules do not need @Deprecated, all are deprecated.

xingfudeshi: Ok

compatible/pom.xml
xingfudeshi: Ok

common/src/main/java/org/apache/seata/common/loader/EnhancedServiceLoader.java
xingfudeshi: If both package names exist, the current one is based on io. seata, and the package for org. apache. seata will only be loaded if it cannot be loaded. In addition, I believe that there should not be two types of package names at the same time, as compatibility itself is to provide users with a transitional solution, not a persistent compatibility solution. If users need to customize a new SPI implementation, they should switch to the package name of Apache seata as soon as possible.

slievrly (commented):
I suggest changing the SPI loading priority here.
1. Load the Apache Seata SPI first. Because Apache Seata is the final state of the future.
2. If the Apache seata SPI cannot be found, output a warn log and try to load the io.seata SPI.
3. If the io.seata SPI also fails to be loaded, handle the exception according to the logic in the original SPI.

xingfudeshi:
> I suggest changing the SPI loading priority here.
> 
> 1. Load the Apache Seata SPI first. Because Apache Seata is the final state of the future.
> 2. If the Apache seata SPI cannot be found, output a warn log and try to load the io.seata SPI.
> 3. If the io.seata SPI also fails to be loaded, handle the exception according to the logic in the original SPI.

Ok

xingfudeshi:
Done. @slievrly

slievrly (approved):
LGTM

PR #6317, opened 2024-01-31
optimize: add jwt authentication for RegisterXXRequests

ggbocoder:

 implements JwtCheckAuthHandler for RegisterTMRequest and RegisterRMRequest


core/src/main/java/org/apache/seata/core/auth/JwtAuthManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+    public static JwtAuthManager getInstance() {
+        if (instance == null) {
+            instance = new JwtAuthManager();
+        }
+        return instance;
+    }
+
+    public void init() {
+        username = ConfigurationFactory.CURRENT_FILE_INSTANCE.getConfig("security." + PRO_USERNAME);
```
funky-eyes: `security.` Why not put it in configurationKeys?
`security.` 为什么不将它放到configurationKeys中

core/src/main/java/org/apache/seata/core/auth/JwtAuthManager.java
```diff
@@ -0,0 +1,133 @@
...
+
+
+public class JwtAuthManager {
+    private String accessToken;
+
+    private String username;
+
+    private String password;
+
+    public final static String PRO_USERNAME = "username";
```
funky-eyes, line 34: 需要在springbootstarter，autoconfig module中添加对应的properties，和config文件夹里的spring下的示例配置
You need to add the corresponding properties in the springbootstarter, autoconfig module, and the example configuration under spring in the config folder.

server/src/main/java/org/apache/seata/server/auth/JwtCheckAuthHandler.java
funky-eyes, line 1: AbstractCheckAuthHandler should only work in one place, and there needs to be configuration to specify it.
AbstractCheckAuthHandler应该只能有一处生效，需要有配置去做指定。

funky-eyes (commented):
当一个陌生的连接，未认证通过，应该不进行任何处理，直接返回一个未认证的response。
When an unfamiliar connection, unauthenticated, passes, it should be left unprocessed and an unauthenticated response should be returned directly.

server/src/main/java/org/apache/seata/server/auth/JwtCheckAuthHandler.java
ggbocoder: Done.Currently, the default loading is DefaultCheckAuthHandler through SPI extension.

core/src/main/java/org/apache/seata/core/auth/JwtAuthManager.java
ggbocoder: done.

core/src/main/java/org/apache/seata/core/auth/JwtAuthManager.java
ggbocoder: done.

ggbocoder:
> 当一个陌生的连接，未认证通过，应该不进行任何处理，直接返回一个未认证的response。 When an unfamiliar connection, unauthenticated, passes, it should be left unprocessed and an unauthenticated response should be returned directly.

Currently, unauthenticated or failed authentication requests or unfamiliar requests will be immediately closed, without further execution of the channel.

PR #6326, opened 2024-02-04
feature: support node metadata sync

funky-eyes:

增加集群间元数据同步能力
1.leader选举成功后，生成只包含它自身的元数据，该元数据为一个单例，并同步至follower
2.follower收到这个元数据后，发送自身的元数据至server侧，进行补充follower的元数据
3.leader收到后，在元数据的follower中增加该节点，并推送元数据变更至client和其它tc节点
并且元数据中增加version字段，用于后续比如存在新旧多版本server，leader可以在同步的时候使用所有节点都能支持的协议，以便向下兼容。

Increase the ability of inter-cluster metadata synchronization 
1. After the leader election is successful, generate metadata containing only itself, which is a singleton and synchronized to followers
2. After the follower receives this metadata, it sends its own metadata to the server side to supplement the follower's metadata.
3. After the leader receives it, add the node to the follower of the metadata, and push the metadata changes to the client and other tc nodes
And the version field is added to the metadata for subsequent use. For example, if there are old and new multi-version servers, the leader can use a protocol that all nodes can support during synchronization for backward compatibility.

image
image


server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
```diff
@@ -287,39 +327,98 @@ public void setRaftLeaderMetadata(RaftClusterMetadata raftClusterMetadata) {
...
 
+    private void syncCurrentNodeInfo(PeerId leaderPeerId) {
+        if (init.compareAndSet(false, true)) {
+            try {
+                RaftServer raftServer = RaftServerManager.getRaftServer(group);
+                PeerId cureentPeerId = raftServer.getServerId();
+                Node node = raftClusterMetadata.createNode(XID.getIpAddress(), XID.getPort(), cureentPeerId.getPort(),
+                    Integer.parseInt(
+                        ((Environment)ObjectHolder.INSTANCE.getObject(OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT))
+                            .getProperty("server.port", String.valueOf(8088))),
```
ptyin: Why 8088 instead of 7091 (i.e., default console port).

server/src/main/java/org/apache/seata/server/cluster/raft/sync/msg/dto/RaftClusterMetadata.java
```diff
@@ -45,11 +47,13 @@ public RaftClusterMetadata(long term) {
         this.term = term;
     }
 
-    public Node createNode(String host, int txPort, int controlPort, String group, Map<String, Object> metadata) {
+    public Node createNode(String host, int txPort,int internalPort, int controlPort, String group, Map<String, Object> metadata) {
```
ptyin: missing space around

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
```diff
@@ -231,21 +244,48 @@ public void onStopFollowing(final LeaderChangeContext ctx) {
...
+                Set<PeerId> newLearners = conf.getLearners();
+                List<Node> currentFollowers = raftClusterMetadata.getFollowers();
+                if (CollectionUtils.isNotEmpty(newFollowers)) {
+                    raftClusterMetadata.setFollowers(currentFollowers.stream()
+                        .filter(node -> contains(node, newFollowers)).collect(Collectors.toList()));
+                }
+                if (CollectionUtils.isNotEmpty(newLearners)) {
+                    raftClusterMetadata.setLearner(raftClusterMetadata.getLearner().stream()
+                        .filter(node -> contains(node, newLearners)).collect(Collectors.toList()));
+                }
```
ptyin, line 267: Why not just override followers and learners, e.g., `raftClusterMetadata.setFollowers(newFollowers)`

funky-eyes: > Why not just override followers and learners, e.g., `raftClusterMetadata.setFollowers(newFollowers)`

Because the peerid in the raft is the port on which only the raft is elected and synchronized, and no other port information for the server exists, it cannot be used directly and needs to be identified

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
funky-eyes: > Why 8088 instead of 7091 (i.e., default console port).

You're right. I made a mistake on this one.

server/src/main/java/org/apache/seata/server/cluster/raft/sync/msg/dto/RaftClusterMetadata.java
funky-eyes: > missing space around

done

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
funky-eyes: > Why 8088 instead of 7091 (i.e., default console port).

done

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
ptyin: OK, I got it.

ptyin (approved):
LGTM

PR #6372, opened 2024-02-22
bugfix: fix postgresql sql  script index name conflicts

gao1143632516:
In the same mode, in order to avoid the conflict of different table index names, it is best to spell the table name in the index name

fix the postgresql.sql index name conflicts


script/server/db/postgresql.sql
funky-eyes, line 1: Why do they conflict?

gao1143632516: CREATE INDEX idx_xid ON public.lock_table (xid);
CREATE INDEX idx_xid ON public.branch_table (xid);
Under the same pattern, there cannot be the same index name,sql inside now there is the same index name idx_xid.

funky-eyes: Is your actual business really experiencing indexing conflicts?

gao1143632516: Yes, we resolved the conflict by concatenating the table names in the index names to ensure that the index names are unique

funky-eyes: > Yes, we resolved the conflict by concatenating the table names in the index names to ensure that the index names are unique

Can you tell me exactly which index is conflicting? If it's all by table name combined with column name as you said, why are they conflicting?

gao1143632516: The postgresql.sql file contains two identical index names idx_xid, which cause conflicts 
`
CREATE INDEX idx_xid ON public.lock_table (xid);
CREATE INDEX idx_xid ON public.branch_table (xid);
`

funky-eyes, line 1: Please add the pr and author information to the 2.x.md files for both languages in the changes folder.

slievrly (approved):
LGTM

Pls add change:

leizhiyuan:
@gao1143632516 need to solve conflict

slievrly:
@gao1143632516 thanks for you contributions.

PR #6393, opened 2024-03-05
optimize: determine the version before sync metadata and add  retry mechanism

funky-eyes:

### DIFF
--- a/server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
+++ b/server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -125,6 +126,8 @@ public class RaftStateMachine extends StateMachineAdapter {
 
     private final AtomicBoolean initSync = new AtomicBoolean(false);
 
+    private ScheduledFuture<?> scheduledFuture;
+    
     public boolean isLeader() {
         return this.leaderTerm.get() > 0;
     }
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
```diff
@@ -141,6 +147,7 @@ public RaftStateMachine(String group) {
             EXECUTES.put(REMOVE_GLOBAL_SESSION, new RemoveGlobalSessionExecute());
             EXECUTES.put(UPDATE_BRANCH_SESSION_STATUS, new UpdateBranchSessionExecute());
             EXECUTES.put(RELEASE_BRANCH_SESSION_LOCK, new BranchReleaseLockExecute());
+            RESYNC_METADATA_POOL.scheduleAtFixedRate(() -> syncCurrentNodeInfo(group), 10, 10, TimeUnit.SECONDS);
```
ptyin: Is it necessary to try to synchronize at startup? Since the node will always synchronize at `onStartFollowing`

liuqiufeng: Also, is it better to have a fixed rate period greater than the rpc timeout?

liuqiufeng (commented):
Is it considered to close the scheduling thread pool after multiple failures, or after a successful sync?

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
funky-eyes: > Is it necessary to try to synchronize at startup? Since the node will always synchronize at `onStartFollowing`

Because of the onStartFollowing synchronization may have accidents, resulting in the lack of successful synchronization, there is a need to have a timed task to compensate for this, in the future, this timed task may also be used as a synchronization of some dynamic metadata, because there is no need to synchronize the dynamic information, so it seems to be some redundancy of this timed task.
因为在onStartFollowing同步时可能出现意外，导致没有同步成功，需要有一个定时任务进行补偿，未来这个定时任务也可能用作一些动态metadata的同步，因为现在不需要动态信息同步，所以显得这个定时任务有一些多余。

funky-eyes:
> Is it considered to close the scheduling thread pool after multiple failures, or after a successful sync?

It can be designed to shut down the thread after the synchronization is complete

server/src/main/java/org/apache/seata/server/cluster/raft/RaftStateMachine.java
ptyin: > > Is it necessary to try to synchronize at startup? Since the node will always synchronize at `onStartFollowing`
> 
> Because of the onStartFollowing synchronization may have accidents, resulting in the lack of successful synchronization, there is a need to have a timed task to compensate for this, in the future, this timed task may also be used as a synchronization of some dynamic metadata, because there is no need to synchronize the dynamic information, so it seems to be some redundancy of this timed task. 因为在onStartFollowing同步时可能出现意外，导致没有同步成功，需要有一个定时任务进行补偿，未来这个定时任务也可能用作一些动态metadata的同步，因为现在不需要动态信息同步，所以显得这个定时任务有一些多余。

OK, it makes sense for me.

funky-eyes:
> Is it considered to close the scheduling thread pool after multiple failures, or after a successful sync?

done

xingfudeshi (approved):
LGTM

liuqiufeng (approved):
LGTM

ptyin (approved):
LGTM

PR #6411, opened 2024-03-11
optimize: optimize discovery compatible module

slievrly:

optimize discovery compatible module

### DIFF
--- a/compatible/src/main/java/io/seata/discovery/registry/RegistryProvider.java
+++ b/compatible/src/main/java/io/seata/discovery/registry/RegistryProvider.java
@@ -20,11 +20,12 @@ package io.seata.discovery.registry;
  * the interface registry provider
  * Notes: used for io.seata SPI interface
  */
-public interface RegistryProvider {
+public interface RegistryProvider extends org.apache.seata.discovery.registry.RegistryProvider{
     /**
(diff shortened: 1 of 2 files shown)


compatible/src/main/java/io/seata/discovery/loadbalance/LoadBalance.java
funky-eyes, line 1: Why not add markup that they are deprecated?

slievrly: Based on my assessment, the io.seata API is likely to remain in existence for the long term and will not be removed. Without any further action, I believe that marking it as deprecated would be meaningless. If io.seata acts as an adapter for org.seata, it would mean that both sets of APIs are available for use.

funky-eyes: > Based on my assessment, the io.seata API is likely to remain in existence for the long term and will not be removed. Without any further action, I believe that marking it as deprecated would be meaningless. If io.seata acts as an adapter for org.seata, it would mean that both sets of APIs are available for use.

Without marking for deprecation, users may never switch to the new api

wt-better (approved):
LGTM

PR #6420, opened 2024-03-14
refactor: optimize Configuration Cache

slievrly:

optimize Configuration Cache


config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
```diff
@@ -0,0 +1,28 @@
...
+package org.apache.seata.config;
+
+public interface CachedConfigurationChangeListener extends ConfigurationChangeListener {
+
+    ConfigurationCache CONFIGURATION_CACHE = ConfigurationCache.getInstance();
+
+    @Override
+    default void afterEvent(ConfigurationChangeEvent event) {
+        ConfigurationChangeListener listener = (ConfigurationChangeListener)CONFIGURATION_CACHE;
+        listener.onProcessEvent(event);
```
funky-eyes, line 26:
```suggestion
   ConfigurationChangeListener afterListener = ConfigurationCache.getInstance();

    @Override
    default void afterEvent(ConfigurationChangeEvent event) {
        afterListener.onProcessEvent(event);
```
这样写是不是更加合适?
Is it more appropriate to write this way?

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationCache.java
```diff
@@ -90,12 +53,10 @@ public void onChangeEvent(ConfigurationChangeEvent event) {
                 CONFIG_CACHE.put(event.getDataId(), new ObjectWrapper(event.getNewValue(), null));
             } else {
                 Object newValue = new ObjectWrapper(event.getNewValue(), null).convertData(oldWrapper.getType());
-                if (!Objects.equals(oldWrapper.getData(), newValue)) {
-                    CONFIG_CACHE.put(event.getDataId(), new ObjectWrapper(newValue, oldWrapper.getType(),oldWrapper.getLastDefaultValue()));
-                }
+                CONFIG_CACHE.replace(event.getDataId(), oldWrapper, new ObjectWrapper(newValue, oldWrapper.getType(), oldWrapper.getLastDefaultValue()));
```
funky-eyes: If the value remains unchanged, it will result in a meaningless replace.

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationCache.java
```diff
@@ -115,6 +76,9 @@ public Configuration proxy(Configuration originalConfiguration) throws Exception
                     }
                     if (null == wrapper
                             || (null != defaultValue && !Objects.equals(defaultValue, wrapper.lastDefaultValue))) {
+                        if (DATA_ID_CACHED.add(rawDataId)) {
+                            originalConfiguration.addConfigListener(rawDataId, this);
```
funky-eyes, line 83: Why do need to listen on every key?

config/seata-config-core/src/test/java/org/apache/seata/config/FileConfigurationTest.java
```diff
@@ -41,13 +41,17 @@ void addConfigListener() throws InterruptedException {
...
+            @Override
+            public void onChangeEvent(ConfigurationChangeEvent event) {
+                Assertions.assertEquals(Boolean.parseBoolean(event.getNewValue()), !Boolean.parseBoolean(event.getOldValue()));
+                countDownLatch.countDown();
+            }
         });
         System.setProperty("service.disableGlobalTransaction", String.valueOf(!value));
         countDownLatch.await(5, TimeUnit.SECONDS);
         System.setProperty("file.listener.enabled", "false");
+        Thread.sleep(1000);
```
funky-eyes, line 61: Why sleep for 1 second?

config/seata-config-etcd3/src/main/java/org/apache/seata/config/etcd3/EtcdConfiguration.java
```diff
@@ -394,7 +394,7 @@ public void onNext(WatchResponse watchResponse) {
                         List<KeyValue> keyValues = getResponse.getKvs();
                         if (CollectionUtils.isNotEmpty(keyValues)) {
                             event.setDataId(dataId).setNewValue(keyValues.get(0).getValue().toString(UTF_8));
-                            listener.onChangeEvent(event);
+                            listener.onProcessEvent(event);
```
funky-eyes, line 397: I'm worried that the user's extended implementation doesn't call onProcessEvent this way.
我担心用户的扩展实现里并不会这样去调用onProcessEvent

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationCache.java
slievrly: The hierarchical structure of the cache is below that of the config; not getting the configuration is meaningless for listening purposes.

config/seata-config-core/src/test/java/org/apache/seata/config/FileConfigurationTest.java
slievrly: For thread safety, the following sleep of 2 seconds is meaningless.

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
slievrly: CachedConfigurationChangeListener is an instance that holds state.

config/seata-config-core/src/test/java/org/apache/seata/config/FileConfigurationTest.java
funky-eyes: ok

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
funky-eyes: > CachedConfigurationChangeListener is an instance that holds state.

Since we must notify ConfigurationCache, and ConfigurationCache also implements ConfigurationChangeListener, why do we need to use it to make it strong, instead of defining it in a member variable?
既然必定要通知到ConfigurationCache,而且ConfigurationCache 也是实现了ConfigurationChangeListener,为什么要在使用使强壮,而不再成员变量中定义好?

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationCache.java
funky-eyes: ok

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
funky-eyes: And then I think since it's an afterlistener, shouldn't it just call onChangeEvent instead of onProcessEvent?
然后我认为既然是afterlistener,是否应该直接调用onChangeEvent,而不是onProcessEvent?

slievrly: Design should not be function-oriented but object-oriented; consider design from the perspective of interactive behavior, not at the level of assembling functions. If ConfigurationCache's before and after are defined to be triggered identically, a logical call stack would be: NacosConfiguration.addConfigListener -> event -> listener.beforeEvent -> listener.onChangeEvent -> listener.afterEvent (ConfigurationCache.beforeEvent -> ConfigurationCache.onChangeEvent -> ConfigurationCache.afterEvent)

slievrly: > > CachedConfigurationChangeListener is an instance that holds state.
> 
> Since we must notify ConfigurationCache, and ConfigurationCache also implements ConfigurationChangeListener, why do we need to use it to make it strong, instead of defining it in a member variable? 既然必定要通知到ConfigurationCache,而且ConfigurationCache 也是实现了ConfigurationChangeListener,为什么要在使用使强壮,而不再成员变量中定义好?

Sorry, I'm unable to understand the logic expressed here.

config/seata-config-etcd3/src/main/java/org/apache/seata/config/etcd3/EtcdConfiguration.java
slievrly: What I am defining here is the interface standard, and what users rely on is also the interface, which is the basic polymorphic pattern of object-oriented programming. Is there a good method that would allow us to inversely constrain user implementations?

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
funky-eyes: > > > CachedConfigurationChangeListener is an instance that holds state.
> > 
> > 
> > Since we must notify ConfigurationCache, and ConfigurationCache also implements ConfigurationChangeListener, why do we need to use it to make it strong, instead of defining it in a member variable? 既然必定要通知到ConfigurationCache,而且ConfigurationCache 也是实现了ConfigurationChangeListener,为什么要在使用使强壮,而不再成员变量中定义好?
> 
> Sorry, I'm unable to understand the logic expressed here.

Why not define it as a ConfigurationChangeListener afterListener = ConfigurationCache.getInstance(); when the member variable is defined, instead of doing the type conversion in afterEvent?
为什么不在成员变量定义的时候就定义为一个ConfigurationChangeListener afterListener = ConfigurationCache.getInstance(); 而是在afterEvent中再进行类型转换?

funky-eyes: > Design should not be function-oriented but object-oriented; consider design from the perspective of interactive behavior, not at the level of assembling functions. If ConfigurationCache's before and after are defined to be triggered identically, a logical call stack would be: NacosConfiguration.addConfigListener -> event -> listener.beforeEvent -> listener.onChangeEvent -> listener.afterEvent (ConfigurationCache.beforeEvent -> ConfigurationCache.onChangeEvent -> ConfigurationCache.afterEvent)

我是认为ConfigurationCache不需要再有beforeEvent的相关逻辑了,如果有的话我认为是否未来可能会出现套娃现象?
I was thinking that ConfigurationCache doesn't need to have beforeEvent related logic anymore, and if it does do I think there might be nesting in the future?

config/seata-config-etcd3/src/main/java/org/apache/seata/config/etcd3/EtcdConfiguration.java
funky-eyes: > What I am defining here is the interface standard, and what users rely on is also the interface, which is the basic polymorphic pattern of object-oriented programming. Is there a good method that would allow us to inversely constrain user implementations?

Perhaps we can keep the behavior of onChangeEvent consistent with that of onProcessEvent and add a new method to replace the behavior of the original onChangeEvent?
或许我们可以将onChangeEvent的行为保持和onProcessEvent一致,新增一个方法来代替原来的onChangeEvent的行为?

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
slievrly: > > Design should not be function-oriented but object-oriented; consider design from the perspective of interactive behavior, not at the level of assembling functions. If ConfigurationCache's before and after are defined to be triggered identically, a logical call stack would be: NacosConfiguration.addConfigListener -> event -> listener.beforeEvent -> listener.onChangeEvent -> listener.afterEvent (ConfigurationCache.beforeEvent -> ConfigurationCache.onChangeEvent -> ConfigurationCache.afterEvent)
> 
> 我是认为ConfigurationCache不需要再有beforeEvent的相关逻辑了,如果有的话我认为是否未来可能会出现套娃现象? I was thinking that ConfigurationCache doesn't need to have beforeEvent related logic anymore, and if it does do I think there might be nesting in the future?

This is about extendable design; it's not about whether a feature exists or not. If it doesn't exist, you can maintain an empty implementation, and if there is predefined logic, you can add it as needed. I suggest that you should not judge the design solely based on current functionalities. For example, with cache, aspects like secret computing and storage are points that can be extended both before and after.

slievrly: > > > > CachedConfigurationChangeListener is an instance that holds state.
> > > 
> > > 
> > > Since we must notify ConfigurationCache, and ConfigurationCache also implements ConfigurationChangeListener, why do we need to use it to make it strong, instead of defining it in a member variable? 既然必定要通知到ConfigurationCache,而且ConfigurationCache 也是实现了ConfigurationChangeListener,为什么要在使用使强壮,而不再成员变量中定义好?
> > 
> > 
> > Sorry, I'm unable to understand the logic expressed here.
> 
> Why not define it as a ConfigurationChangeListener afterListener = ConfigurationCache.getInstance(); when the member variable is defined, instead of doing the type conversion in afterEvent? 为什么不在成员变量定义的时候就定义为一个ConfigurationChangeListener afterListener = ConfigurationCache.getInstance(); 而是在afterEvent中再进行类型转换?

I understand what you mean. However, this might be related to my personal habits. It can be replaced with one line syntactically, but in reality, it involves two steps. 
But from a design perspective, I would like to point out a few things. Member variables are intended for internal use within a class to maintain consistency in the architecture. Once converted into an interface here, it means that other methods can only use methods defined by the interface.

config/seata-config-etcd3/src/main/java/org/apache/seata/config/etcd3/EtcdConfiguration.java
slievrly: The original configuration center had two methods of invocation: onChangeEvent and onProcessEvent. Which do you think should be the standard? The current changes have unified all previous onChangeEvent calls to onProcessEvent. What kind of expectations do you think this has for user behavior? For users to extend on their own, it means they need to understand structurally the presence or absence of cache, as well as the difference between synchronous and asynchronous operations.

funky-eyes:
image
image

config/seata-config-core/src/main/java/org/apache/seata/config/CachedConfigurationChangeListener.java
funky-eyes: ok

config/seata-config-etcd3/src/main/java/org/apache/seata/config/etcd3/EtcdConfiguration.java
funky-eyes: Because the configuration is mainly used by seata, not the user, if this change can't make the user-side implementation behave the same as our implementation, it will affect the question of whether the configuration is effective in seata or not. So this should be a strong constraint setting, all notifications should go onProcessEvent behavior, otherwise the user's extension will lose a lot of functionality without perception
因为配置主要是由seata使用，并不是用户使用，如果这个改动无法让用户侧的实现跟我们的实现行为一致，这将影响到配置在seata中是否生效的问题。所以这块应该是强约束的设定，所有的通知应该都走onProcessEvent这个行为，否则用户的扩展将在无感知的情况下丧失很多功能

funky-eyes (approved):
LGTM

PR #6456, opened 2024-04-01
test: adjust the test cases related to dynamic configuration

funky-eyes:

### DIFF
--- a/config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationChangeListener.java
+++ b/config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationChangeListener.java
@@ -17,7 +17,7 @@
 package org.apache.seata.config;
 
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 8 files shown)


xingfudeshi (approved):
LGTM

config/seata-config-core/src/test/java/org/apache/seata/config/FileConfigurationTest.java
```diff
@@ -52,13 +52,20 @@ public void onChangeEvent(ConfigurationChangeEvent event) {
             }
         });
         System.setProperty(dataId, String.valueOf(!value));
-        countDownLatch.await(2, TimeUnit.SECONDS);
-        System.setProperty("file.listener.enabled", "false");
+        countDownLatch.await(10, TimeUnit.SECONDS);
```
slievrly, line 55: I think you should add a synchronous assertion of the return value or the first value change.
In the case of past failures, one possible reason is that the configCache update was not triggered after await(2s), causing fileConfig.getBoolean(dataId) to return the cached old value. In the CachedConfigurationChangeListener assertion is not effective, because in the asynchronous link, every change in the value of the trigger, but the timing of the trigger is not controllable. Even if the test process exits, the asynchronous thread may not execute.

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationChangeListener.java
```diff
@@ -40,9 +40,10 @@ public interface ConfigurationChangeListener {
     /**
      * The constant EXECUTOR_SERVICE.
      */
-    ExecutorService EXECUTOR_SERVICE = new ThreadPoolExecutor(CORE_LISTENER_THREAD, MAX_LISTENER_THREAD,
-        Integer.MAX_VALUE, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>(),
-        new NamedThreadFactory("configListenerOperate", MAX_LISTENER_THREAD));
+    ExecutorService EXECUTOR_SERVICE =
+        new ThreadPoolExecutor(CORE_LISTENER_THREAD, MAX_LISTENER_THREAD, MAX_LISTENER_THREAD, TimeUnit.MILLISECONDS,
```
slievrly: The event notification model is not suited for this threading model, the backpressure mode is better suited for dealing with direct interactions with users. Inputs external to the event model should not introduce unpredictable behavior to the core's operation.

config/seata-config-core/src/test/java/org/apache/seata/config/FileConfigurationTest.java
funky-eyes: done

config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationChangeListener.java
funky-eyes: done

PR #6121, opened 2023-12-09
bugfix: the branch transaction order error when rolling back

funky-eyes:
- [ ] I have registered the PR [changes](../changes).

### Ⅱ. Does this pull request fix one issue?

fixes #6120

PR #6547, opened 2024-05-14
bugfix: Corrected an issue in the Seata AT mode where, upon multiple …

mobuxian:
…modifications to the same record and encountering an exception during business execution, the rollback process would fail.

When a business operation fails, Seata initiates a data rollback and concurrently checks if the data has been modified. If modifications are found, the rollback process may fail. In scenarios where the same record is updated multiple times, Seata's default sequential rollback approach can lead to rollback failures. By altering Seata's rollback strategy to perform rollbacks in reverse order—starting with the most recent modification and proceeding backwards—the issue can be resolved. This adjustment ensures that changes are undone from the latest to the earliest, effectively addressing the problem. (The fix involves storages such as DB, Redis, and file systems correctly returning branchIDs.)


funky-eyes:
https://github.com/apache/incubator-seata/pull/6121

PR #6370, opened 2024-02-22
feature: saga-remove spring

wt-better:
- [x] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did

Seata saga current is strong dependency on spring. such as: 

1. code was dependency spring beans (InitializingBean ..etc)
2. expression evaluation dependency spring el 
3. spring bean service Task dependency spring ioc 

This is not conducive to saga's multilanguage building, so I has decouple saga with spring.

1. To ensure code compatibility, make new module named 'seata-saga-spring' and move related classes(not modified path) to this module.

2. 'seata-saga-tm' has spring dependency, this module is not necessary. First I removed this module and move class to 'seata-saga-spring' , I will sink TM's core capabilities ‘seata-saga-engine’ module later pr.

3. 'seata-saga-engine-store' remove the implementation of serialization and make it more pure

4. 'seata-saga-rm' has class named `StateMachineEngineHolder`, i moved to ‘seata-saga-engine’ module, and package name change, it`s not compatible. why ? I think it is inner use, users rarely use it.

PR #6553, opened 2024-05-16
bugfix: fix saga "cannot matching status"

ptyin:

`ELExpression` is introduced in the following PR,
- #6370

And the modification exposed a bug, causing no evaluator can be applied after executing the 'ServiceTask'.
```text
State [X]  execute finished, but cannot matching status, pls check its status manually.
```
image-20240509200546136


wt-better (approved):
LGTM

PR #6580, opened 2024-05-25
feature：seata benchmark

wt-better:
Building a benchmark for seata. benchmark is divided into three parts from micro to macro:
- The first part, using jmh for micro-benchmark testing as a performance baseline, applying atomic methods
- The second part is basic benchmark testing of methods or services, mainly profiler and tracing.
- The third part, AT, tcc and other transaction modes, business scenario testing


seata-benchmark/seata-benchmark-baseline/src/main/java/org/apache/seata/benchmark/baseline/BenchmarkChartGenerator.java
```diff
@@ -0,0 +1,122 @@
...
+            File chartFile = new File(modulePath + "chart/" + className + ".png");
+            ChartUtils.saveChartAsPNG(chartFile, barChart, 640, 480);
+
+            LOGGER.info("Generated chart: " + chartFile.getAbsolutePath());
+        }
+    }
+
+    public static String readFileContent(File file) throws IOException {
+        StringBuilder contentBuilder = new StringBuilder();
+        // 使用当前线程的类加载器获取资源作为流
```
xjlgod, line 100: Use english comment here.

seata-benchmark/seata-benchmark-scenario/src/main/resources/benchmark.sh
xjlgod, line 1: What is the use of this file.

seata-benchmark/seata-benchmark-baseline/src/main/java/org/apache/seata/benchmark/baseline/BenchmarkChartGenerator.java
wt-better: ok, i will fix

seata-benchmark/seata-benchmark-scenario/src/main/resources/benchmark.sh
wt-better: Strictly speaking, the scenario is not implemented yet, this is holder, can delete.

Issue #6696, opened 2024-07-24
v0 ByteBuf should not decode by super class


Bughue:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
v0 ByteBuf should not be decoded by super class, it will lead to a IllegalReferenceCountException

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:83)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:148)
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

Minimal yet complete reproducer code (or URL to code):

### Ⅵ. Environment:

- JDK version(e.g. `java -version`):
- Seata client/server version:  
- Database version:
- OS(e.g. `uname -a`):
- Others:

PR #6697, opened 2024-07-24
bugfix: v0 ByteBuf should not decode by super class

Bughue:

https://github.com/apache/incubator-seata/issues/6696

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/v0/ProtocolDecoderV0.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/v0/ProtocolDecoderV0.java
@@ -151,5 +151,4 @@ public class ProtocolDecoderV0 extends LengthFieldBasedFrameDecoder implements P
             throw new DecodeException(exx);
         }
     }
-
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/seata/core/rpc/netty/v0/ProtocolDecoderV0.java
```diff
@@ -144,22 +144,15 @@ public RpcMessage decodeFrame(ByteBuf in) {
...
-                }
+            if (in instanceof ByteBuf) {
+                return decodeFrame(in);
             }
         } catch (Exception exx) {
             LOGGER.error("Decode frame error, cause: {}", exx.getMessage());
             throw new DecodeException(exx);
         }
-        return decoded;
+        return in;
```
funky-eyes: why return in？

Bughue: Running here means that an exception or incorrect type was encountered. 
Follow the original logic, skip this processing, return **in**

funky-eyes (approved):
LGTM

PR #6770, opened 2024-08-19
optimize: automatic deletion of namingserver vgroup through Caffeine map

ggbocoder:


namingserver/src/main/java/org/apache/seata/namingserver/manager/NamingManager.java
```diff
@@ -83,12 +89,22 @@ public class NamingManager {
...
     public NamingManager() {
         this.instanceLiveTable = new ConcurrentHashMap<>();
-        this.vGroupMap = new ConcurrentHashMap<>();
         this.namespaceClusterDataMap = new ConcurrentHashMap<>();
     }
 
     @PostConstruct
     public void init() {
+        this.vGroupMap = Caffeine.newBuilder()
+                .expireAfterAccess(10000, TimeUnit.MILLISECONDS) // expired time
```
funky-eyes: The time should be the heartbeat time

namingserver/src/main/java/org/apache/seata/namingserver/manager/NamingManager.java
```diff
@@ -83,12 +89,22 @@ public class NamingManager {
...
+                .maximumSize(Integer.MAX_VALUE)
+                .removalListener(new RemovalListener<Object, Object>() {
+
+                    @Override
+                    public void onRemoval(@Nullable Object key, @Nullable Object value, @NonNull RemovalCause cause) {
+                        System.err.println("out " + key + " " + value);
```
funky-eyes: 判断是否启用了debug级别,如果是就输出debug日志,提示用户该vgroup已被删除
Determine whether the debug level is enabled, and if so, print the debug log, prompting the user that the vgroup has been deleted

funky-eyes (approved):
LGTM

PR #6781, opened 2024-08-25
bugfix: the issue where the TC occasionally fails to go offline from the NamingServer

funky-eyes:

1.修复tc下线时,由于定时任务没有先关闭,导致下线后还会被注册上,需要靠namingserver的健康检查来下线的bug
2.修复change group时,由于先添加后删除,导致先添加的节点已经注册到了namingserver侧,去删除时使用错了cluster的节点去删除,导致change失败的bug

Fix the bug where TC remains registered after going offline because the scheduled tasks were not stopped first; the issue is dependent on the NamingServer's health check to eventually deregister.

Fix the bug that occurs during a change group operation where nodes are added before being deleted. This causes nodes added first to be registered on the NamingServer side, and the deletion process mistakenly uses the wrong cluster nodes, leading to a change failure.

### DIFF
--- a/seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/registry/RegistryNamingServerProperties.java
+++ b/seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/registry/RegistryNamingServerProperties.java
@@ -25,7 +25,7 @@ import static org.apache.seata.spring.boot.autoconfigure.StarterConstants.REGIST
 @ConfigurationProperties(prefix = REGISTRY_NAMINGSERVER_PREFIX)
 public class RegistryNamingServerProperties {
     private String cluster = "default";
-    private String serverAddr = "127.0.0.1:8500";
+    private String serverAddr = "127.0.0.1:8081";
     private String namespace = "public";
 
     private int heartbeatPeriod = 5000;
--- a/seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/registry/RegistryNamingServerProperties.java
+++ b/seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/registry/RegistryNamingServerProperties.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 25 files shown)


xingfudeshi (approved):
LGTM

seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/registry/RegistryNamingServerProperties.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+import static org.apache.seata.spring.boot.autoconfigure.StarterConstants.REGISTRY_NAMINGSERVER_PREFIX;
+
+@Component
+@ConfigurationProperties(prefix = REGISTRY_NAMINGSERVER_PREFIX)
+public class RegistryNamingServerProperties {
+    private String cluster = "default";
+    private String serverAddr = "127.0.0.1:8500";
```
funky-eyes:
```suggestion
    private String serverAddr = "127.0.0.1:8081";
```

server/pom.xml
```diff
@@ -194,11 +194,10 @@
...
         </dependency>
-<!--        <dependency>-->
-<!--            <groupId>org.mariadb.jdbc</groupId>-->
-<!--            <artifactId>mariadb-java-client</artifactId>-->
-<!--            <scope>provided</scope>-->
-<!--        </dependency>-->
+        <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+        </dependency>
```
funky-eyes:
```suggestion
<!--        <dependency>-->
<!--            <groupId>org.mariadb.jdbc</groupId>-->
<!--            <artifactId>mariadb-java-client</artifactId>-->
<!--            <scope>provided</scope>-->
<!--        </dependency>-->
```

server/pom.xml
```diff
@@ -165,11 +165,11 @@
             <version>${project.version}</version>
         </dependency>
 
-        <dependency>
+<!--        <dependency>
```
funky-eyes:
```suggestion
        <dependency>
```

server/pom.xml
```diff
@@ -165,11 +165,11 @@
...
             <groupId>${project.groupId}</groupId>
             <artifactId>seata-console</artifactId>
             <version>${project.version}</version>
-        </dependency>
+        </dependency>-->
```
funky-eyes:
```suggestion
        </dependency>
```

server/src/main/java/org/apache/seata/server/ServerApplication.java
```diff
@@ -25,7 +25,7 @@
  */
 @SpringBootApplication(scanBasePackages = {"org.apache.seata"})
 public class ServerApplication {
-    public static void main(String[] args) throws IOException {
+    public static void main(String[] args) throws IOException, ClassNotFoundException {
```
funky-eyes:
```suggestion
    public static void main(String[] args) throws IOException {
```

PR #6786, opened 2024-08-27
bugfix: fix prometheus fail to return seata metrics data when using Nacos

slievrly:

fix prometheus fail to return seata metrics data when using Nacos


common/src/main/java/org/apache/seata/common/DefaultValues.java
```diff
@@ -259,7 +259,7 @@ public interface DefaultValues {
     /**
      * the const DEFAULT_METRICS_ENABLED
      */
-    boolean DEFAULT_METRICS_ENABLED = false;
+    boolean DEFAULT_METRICS_ENABLED = true;
```
funky-eyes, line 262: Why is the default value changed to true?

script/config-center/config.txt
```diff
@@ -164,7 +164,7 @@ server.raft.sync=true
 
 
 #Metrics configuration, only for the server
-metrics.enabled=false
+metrics.enabled=true
```
funky-eyes, line 167: Why is the default value changed to true?

common/src/main/java/org/apache/seata/common/DefaultValues.java
slievrly: Monitoring data should be exposed from the kernel layer by default, and whether to display it is controlled by whether the service layer is connected to the corresponding monitoring system.

xjlgod:
LGTM.I have tested by some examples.

funky-eyes (approved):
LGTM

PR #6806, opened 2024-09-02
optimize: avoid dead loop logging during cache plan refresh exceptions

wuwen5:
Avoid dead loop logging during cache plan refresh exceptions

### DIFF
--- a/rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
+++ b/rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
@@ -135,6 +135,7 @@ public class TableMetaCacheFactory {
                         }
                     } catch (Exception exx) {
                         LOGGER.error("table refresh error:{}", exx.getMessage(), exx);
+                        lastRefreshFinishTime = System.nanoTime();
                     }
                 }
             });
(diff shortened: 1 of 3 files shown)


rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
```diff
@@ -135,6 +135,7 @@ static class TableMetaRefreshHolder {
                         }
                     } catch (Exception exx) {
                         LOGGER.error("table refresh error:{}", exx.getMessage(), exx);
+                        lastRefreshFinishTime = System.nanoTime();
```
funky-eyes, line 139: 增加一行注释说明下为什么这么改动，以便于用户看到这段代码时，能知道为什么这么做的场景。然后再登记下pr，属于优化类型。
Add a comment explaining why this change was made, so users understand the context when they see this code. Then, log this PR, noting that it is for optimization.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
```diff
@@ -135,6 +135,8 @@ static class TableMetaRefreshHolder {
                         }
                     } catch (Exception exx) {
                         LOGGER.error("table refresh error:{}", exx.getMessage(), exx);
+                        //Avoid dead loop logging and high CPU during exceptions
```
funky-eyes:
```suggestion
                        // Avoid high CPU usage due to infinite loops caused by database exceptions
```

funky-eyes (approved):
LGTM

PR #6809, opened 2024-09-03
test CodeQL

wangliang181230:


rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/cache/AbstractTableMetaCache.java
slievrly, line 1: what's happen, empty file?

test/src/test/resources/saga/spring/statemachine_engine_db_mockserver_test.xml
```diff
@@ -1,70 +1,70 @@
...
-	</bean>
-
-	<bean name="callerRunsPolicy" class="java.util.concurrent.ThreadPoolExecutor.CallerRunsPolicy">
-	</bean>
-
-	<bean id="demoService" class="org.apache.seata.saga.engine.mock.DemoService"/>
-
-	<bean id="mockSagaTransactionTemplate" class="org.apache.seata.saga.engine.mock.MockSagaTransactionTemplate"/>
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
slievrly, line 2: It looks like git.autocrlf has changed, but the content actually hasn't changed.

PR #6826, opened 2024-09-07
optimize: remove the branch registration operation of the XA read-only transaction

GoodBoyCoder:
…n and other related operations

XA模式下。如果是只读事务就不注册分支及xa的任何操作
In XA mode, if it is a read-only transaction, the branch and any operation of xa will not be registered.

Manually set the connection's readOnly to true or set the annotation @Transactional(readOnly = true) in the Spring framework


funky-eyes (approved):
LGTM

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
```diff
@@ -389,4 +394,8 @@ private void reportStatusToTC(BranchStatus status) {
         }
     }
 
+    private boolean checkReadOnly() throws SQLException {
```
slievrly: It might be more appropriate to change the method name to isReadOnly. checkXXX indicates an action and isXXX indicates a state.

GoodBoyCoder: Done

xingfudeshi (approved):
LGTM

PR #6893, opened 2024-09-29
Fix issue like PR #6702 in snapshot module.

CFionaBF:
Fix an issue similar to pull request #6702 in the snapshot module.

### DIFF
--- a/server/src/main/java/org/apache/seata/server/cluster/raft/snapshot/RaftSnapshotSerializer.java
+++ b/server/src/main/java/org/apache/seata/server/cluster/raft/snapshot/RaftSnapshotSerializer.java
@@ -16,13 +16,13 @@
  */
 package org.apache.seata.server.cluster.raft.snapshot;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/seata/server/cluster/raft/snapshot/RaftSnapshotSerializer.java
```diff
@@ -16,13 +16,13 @@
  */
 package org.apache.seata.server.cluster.raft.snapshot;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
+import java.io.*;
```
funky-eyes, line 19:
```suggestion
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.ObjectInputStream;
import java.io.ObjectOutputStream;
```

funky-eyes (commented):
You should add another test case to ensure that the whitelist mechanism for this change is effective
你应该再增加一个测试用例，以保证这个改动的白名单机制生效

PR #6973, opened 2024-11-02
feature: support saga annotation

wt-better:
Support the two-phase transaction of saga annotation.

1. Saga branch transactions are divided into two phases. Unlike tcc, the first phase is directly submitted without the second phase. It is a degenerate tcc.
2. Add the annotation `CompensationBusinessAction`, which needs to be added to the service interface or implementation of the saga mode branch transaction
3. Compared with the state machine, the saga annotation mode currently does not support the forward retry function.
4. The saga annotation mode currently does not support the anti-hanging function.

In order to support the above functions, we have also made the following changes/optimizations:
1. The historical implementation of `DefaultResouecePasser`  will cause a resource to be registered multiple times; in order to avoid excessive abstraction and reduce the complexity of parsing cost, i was removed it.


wt-better:
link: https://github.com/apache/incubator-seata/pull/5300

Changes:
1. Remove the dependency of saga annotation on tcc. This method is heavy, and tcc has more customized logic, such as tccHook. Independent development is a better solution

2. Fix the duplication of saga annotation interfacerparser type

3. Remove DefaultResouecePasser and corresponding implementation to avoid over-abstraction and time-consuming parsing

4. Resolve conflicts with the latest branch

test/pom.xml
```diff
@@ -183,6 +182,12 @@
             <artifactId>rocketmq-client</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.seata</groupId>
+            <artifactId>seata-saga-annotation</artifactId>
+            <version>2.3.0-SNAPSHOT</version>
```
funky-eyes:
```suggestion
            <version>${project.version}</version>
```

test/src/test/java/org/apache/seata/saga/annotation/NormalSagaAnnotationAction.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.seata.rm.tcc.api.BusinessActionContextParameter;
+import org.apache.seata.rm.tcc.api.LocalTCC;
+import org.apache.seata.saga.rm.api.CompensationBusinessAction;
+
+import java.util.List;
+
+/**
+ * The interface saga action.
+ */
+@LocalTCC
```
funky-eyes: 为什么还要用@LocalTCC，注解化saga应该跟TCC有明显的类名和注解的区分
Why use @LocalTCC? Annotated saga should be clearly distinguished from TCC by class names and annotations

wt-better: Saga annotation is generally used for rpc, localTcc will be parsed into remote bean, for the convenience of testing

test/pom.xml
wt-better: ok

saga/seata-saga-annotation/src/main/resources/META-INF/services/org.apache.seata.core.model.ResourceManager
```diff
@@ -14,4 +14,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-io.seata.rm.tcc.resource.parser.TccRegisterResourceParser
\ No newline at end of file
+org.apache.seata.saga.rm.SagaAnnotationResourceManager
```
slievrly, line 17: whether the SPI interface in the compatibility package are no longer supported?

wt-better: Yes, we hope to upgrade to the Apache package to use the new features. The compatibility package is not a long solution.

test/src/test/java/org/apache/seata/saga/annotation/NormalSagaAnnotationAction.java
wt-better: Fixed, use class annotation instead.

saga/seata-saga-annotation/src/main/resources/META-INF/services/org.apache.seata.core.model.ResourceManager
wt-better: RegisterResourceParser  design is too complex, currently only used in tcc proxy registry, and there are few scenarios for extended use, so it is decide not compatible

funky-eyes (approved):
LGTM

saga/seata-saga-annotation/src/main/resources/META-INF/services/org.apache.seata.core.model.ResourceManager
slievrly: ok.

slievrly (approved):
LGTM
I have checked the seata-samples and it works as expected.

PR #6988, opened 2024-11-07
optimize: delete unnecessary cached data

MaoMaoandSnail:

delete unnecessary cached data


saga/seata-saga-engine/src/main/java/org/apache/seata/saga/engine/repo/impl/StateMachineRepositoryImpl.java
```diff
@@ -71,8 +71,6 @@ public StateMachine getStateMachineById(String stateMachineId) {
                         stateMachine.setStartState(parsedStatMachine.getStartState());
                         stateMachine.getStates().putAll(parsedStatMachine.getStates());
                         item.setValue(stateMachine);
-                        stateMachineMapById.put(stateMachine.getName() + "_" + stateMachine.getTenantId(),
-                            item);
```
slievrly, line 75: It is not currently in use, but you are not advised to delete the tenant information.

MaoMaoandSnail: Since other places do not store data with a key in the format "Name+tenantId" when populating the stateMachineMapById cache, even though such data could theoretically be stored, I recommend deleting it to maintain consistency with the caching style. Otherwise, it may cause confusion.

MaoMaoandSnail: > It is not currently in use, but you are not advised to delete the tenant information.
Name+tenantId is an implicit primary key and could theoretically be stored in stateMachineMapById. However, after adding it here, other parts of the system do not call this cache, nor do they use Name+tenantId as an implicit primary key when storing data in stateMachineMapById. Instead, they store the data separately. This is the only place where it is handled individually. I believe the approach should be unified: either handle it everywhere or not handle it at all; otherwise, it may cause confusion.

MaoMaoandSnail: > It is not currently in use, but you are not advised to delete the tenant information.


Name+tenantId 是一个隐式的主键,按道理说是可以放在stateMachineMapById中,但是我们这里放进去之后,首先其他地方并没有调用该缓存,并且其他地方缓存数据时,也并没有将Name+tenantId作为隐式主键放到stateMachineMapById中,而是分开存储,只有这个地方单独处理,我觉的应该统一方式,要处理就都处理,要不处理就都不处理,不然容易让人产生困惑
Name+tenantId is an implicit primary key and could theoretically be stored in stateMachineMapById. However, after adding it here, other parts of the system do not call this cache, nor do they use Name+tenantId as an implicit primary key when storing data in stateMachineMapById. Instead, they store the data separately. This is the only place where it is handled individually. I believe the approach should be unified: either handle it everywhere or not handle it at all; otherwise, it may cause confusion.

PR #6998, opened 2024-11-14
optimize : skip sending some request if client-version is v0

### LINKED ISSUE #6999: v0-client(<0.7.1) can not decode UndoLogDeleteRequest
## Why you need it?

## How it could be?

Bughue:

1. 新增了客户端版本+msgType来判断是否跳过的逻辑
2. 这个功能在`sendAsync`和`sendSync`都加上了，保证代码逻辑的一致性，但`sendSync`的“if”是不会进的。
3. sendSync的时候会返回一个`VersionNotSupportMessage`。目前来说TC使用sendSync的逻辑不会做这种判断，后续如果有skip的情况就需要在接收的时候先判断类型，而不是直接强转。
4. 在NettyClientChannelManager中为channel注册版本号

---

1. Added client version + msgType to determine whether to skip the logic.
2. this feature is added to both `sendAsync` and `sendSync` to ensure the consistency of code logic, but the “if” of `sendSync` will not be entered.
3. sendSync will return a `VersionNotSupportMessage`. at present, TC use sendSync logic will not do this kind of judgment, and then if there is a skip case you need to judge the type of the first time you receive, rather than directly force the transfer.
4. Register the version number for the channel in the NettyClientChannelManager

Translated with DeepL.com (free version)

fixes https://github.com/apache/incubator-seata/issues/6999


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemoting.java
```diff
@@ -173,6 +175,12 @@ protected Object sendSync(Channel channel, RpcMessage rpcMessage, long timeoutMi
             LOGGER.warn("sendSync nothing, caused by null channel.");
             return null;
         }
+        if (MsgVersionHelper.versionNotSupport(channel, rpcMessage)) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Message sending will be skipped as the client version does not support it,{}", rpcMessage);
+            }
+            return new VersionNotSupportMessage();
```
funky-eyes, line 182: 看起来没有使用到它的地方，后续将有何作用？
It seems that it is not being used anywhere. What will its role be in the future?

Bughue: > 看起来没有使用到它的地方，后续将有何作用？ It seems that it is not being used anywhere. What will its role be in the future?

比如以后的版本里如果要sync发送某个全新的request类，旧版本无法识别，在remote通信这层会返回VersionNotSupportMessage，发送者判断到返回类型是这个后做忽略处理或者抛出异常

For example, if you want to send a sync to a new request class in a future version that is not recognized by the old version, the versionNotSupportMessage will be returned at the remote communication level, and the sender will either ignore the return type or throw an exception.

core/src/main/java/org/apache/seata/core/rpc/MsgVersionHelper.java
```diff
@@ -0,0 +1,52 @@
...
+        Object msg = rpcMessage.getBody();
+        String version = Version.getChannelVersion(channel);
+        if (StringUtils.isBlank(version) || msg == null) {
+            return false;
+        }
+        boolean isV0 = Version.isV0(version);
+        if (!isV0 || !(msg instanceof MessageTypeAware)) {
+            return false;
+        }
+        short typeCode = ((MessageTypeAware) msg).getTypeCode();
```
slievrly, line 49: Do not force cast.

core/src/main/java/org/apache/seata/core/rpc/MsgVersionHelper.java
```diff
@@ -0,0 +1,52 @@
...
+        return SKIP_MSG_CODE_V0.contains(typeCode);
```
slievrly, line 50: Pls confirm that the zero value is a message TypeCode with no business significance.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemoting.java
slievrly, line 182: Why not throw an exception?

core/src/main/java/org/apache/seata/core/rpc/MsgVersionHelper.java
Bughue: rpcMessage.getBody() is of type Object, how do I getTypeCode without forced cast?

core/src/main/java/org/apache/seata/core/rpc/MsgVersionHelper.java
Bughue: SKIP_MSG_CODE_V0 is not a zero value, it is a list of types to skip. 

`List<Short> SKIP_MSG_CODE_V0 = Arrays.asList(MessageType.TYPE_RM_DELETE_UNDOLOG);`

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemoting.java
Bughue: If an exception is thrown, the server may abort the current execution. In that case, the message-sending step should be skipped, yet still considered successfully completed.

slievrly (approved):
LGTM

PR #7027, opened 2024-11-26
optimize: raft mode maintains the reload logic consistent with the file

funky-eyes:

存在于file一样的缺点，reload后会删除failed状态的事务及涉及的分支事务和锁
以及reload时机于file有区别，file是在重启时，而raft是在leader选举后进行reload（一般leader重选举就是leader挂了）

It has the same drawbacks as in the file: after reload, failed transactions, associated branch transactions, and locks will be deleted. Additionally, the timing of the reload differs from that of the file; the file reload occurs during a restart, while the RAFT reload happens after a leader election (usually, the leader re-election occurs when the leader has failed).


slievrly (approved):
LGTM

PR #7030, opened 2024-11-28
test: add some test case.

ujjboy:

add some test case.


slievrly (commented):
don't suggest to add an author or personal information.

common/src/test/java/org/apache/seata/common/exception/ExceptionUtilTest.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.seata.common.exception;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.UndeclaredThrowableException;
+
+/**
+ * @author <a href=mailto:ujjboy@qq.com>GengZhang</a>
```
xjlgod: @funky-eyes Do we need author info in class header ?

slievrly (approved):
LGTM

PR #7033, opened 2024-11-29
optimize: add the decompression size limit

slievrly:

add the decompression size limit

### DIFF
--- a/compressor/seata-compressor-zstd/src/main/java/org/apache/seata/compressor/zstd/ZstdUtil.java
+++ b/compressor/seata-compressor-zstd/src/main/java/org/apache/seata/compressor/zstd/ZstdUtil.java
@@ -24,6 +24,8 @@ import com.github.luben.zstd.Zstd;
  */
 public class ZstdUtil {
 
+    public static final int MAX_COMPRESSED_SIZE = 4 * 1024 * 1024;
+
     public static byte[] compress(byte[] bytes) {
         if (bytes == null) {
             throw new NullPointerException("bytes is null");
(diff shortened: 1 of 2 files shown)


compressor/seata-compressor-zstd/src/main/java/org/apache/seata/compressor/zstd/ZstdUtil.java
```diff
@@ -24,6 +24,8 @@
  */
 public class ZstdUtil {
 
+    public static final int MAX_COMPRESSED_SIZE = 4 * 1024 * 1024;
```
funky-eyes, line 27: 直接定死这个配置值了吗？不需要开放一个配置项吗？
Is this configuration value set directly, or is there an option to configure it?

funky-eyes:
Can the maximum value for other compression type also be configured?

compressor/seata-compressor-zstd/src/main/java/org/apache/seata/compressor/zstd/ZstdUtil.java
slievrly: Not open for now; 4MB is quite a large size. The variable data in Seata mainly comes from the primary key values in the branch registration. If we estimate based on a primary key size of 128, it can hold around 32,768 primary key values.

funky-eyes: ok

slievrly:
> Can the maximum value for other compression type also be configured?

I checked that the other compression methods all use ByteArrayOutputStream as the stream output and use the reusable buffer size, there is no similar issue.

funky-eyes (approved):
LGTM

PR #7038, opened 2024-11-30
feature: add fury serializer support

GoodBoyCoder:

Add support for serializing rpc messages using Fury

set transport.serialization=fury


serializer/seata-serializer-fury/src/main/java/org/apache/seata/serializer/fury/FurySerializer.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public <T> T deserialize(byte[] bytes) {
+        if (bytes == null || bytes.length == 0) {
+            throw new IllegalArgumentException("bytes is null");
+        }
+        ThreadSafeFury threadSafeFury = FurySerializerFactory.getInstance().get();
+        try {
+            return  (T) threadSafeFury.deserialize(bytes);
+        } catch (Exception e) {
+            e.printStackTrace();
```
funky-eyes: 应该将其输出为error日志，或者抛出异常
It should be output as an error log, or an exception should be thrown

dependencies/pom.xml
```diff
@@ -879,6 +882,12 @@
                 <version>${rocketmq-version}</version>
             </dependency>
 
+            <!-- Fury Serialize -->
```
slievrly: Does it need to be declared in the seata-all pom?

GoodBoyCoder: This is not the pom.xml of the seata-all module, but the seata-dependencies module.

serializer/seata-serializer-fury/src/main/java/org/apache/seata/serializer/fury/FurySerializer.java
GoodBoyCoder: The code for the test exception was not completely cleaned up. Done

funky-eyes (approved):
LGTM

Issue #7042, opened 2024-12-02
The default value of the property `client.rm.tableMetaCheckEnable` should be changed to `false` to align with the behavior of Seata Client 2.1.0 and earlier


linghengqian:
## Why you need it?
 Is your feature request related to a problem? Please describe in details

- The default value of the property `client.rm.tableMetaCheckEnable` is always `true`, but this property never took effect in Seata Client 2.1.0 and before until https://github.com/apache/incubator-seata/pull/6661 was merged on the Seata 2.2.0 milestone.
- `client.rm.tableMetaCheckEnable` remains `true`, which is like entering the big world of Otome game for unit testing. It will continue to try to send requests after the TM and RM of Seata Client are destroyed, **which looks like a connection leak.** 
- I created a unit test at https://github.com/linghengqian/seata-table-meta-check-enable-test . We introduce a premise that the default value of `client.rm.tableMetaCheckerInterval` is `60000L`, corresponding to 60 seconds. **The unit test deliberately takes 2 minutes to complete through `Awaitility.await().timeout(Duration.ofMinutes(5L)).pollDelay(Duration.ofMinutes(2L)).until(() -> true);`.** Verified unit test under Ubuntu 22.04.4 LTS with `SDKMAN!` and `Docker CE`.

```shell
sdk install java 23-open

git clone git@github.com:linghengqian/seata-table-meta-check-enable-test.git
cd ./seata-table-meta-check-enable-test/
sdk use java 23-open
./mvnw -T 1C clean test
```
- The core logic is as follows:
```java
@SuppressWarnings("resource")
public class SimpleTest {
    @Test
    void test() {
        assertThat(System.getProperty("service.default.grouplist"), is(nullValue()));
        try (GenericContainer<?> seataContainer = new GenericContainer<>("apache/seata-server:2.2.0")
                .withExposedPorts(7091, 8091)
                .waitingFor(Wait.forHttp("/health").forPort(7091).forStatusCode(200).forResponsePredicate("ok"::equals))
        ) {
            seataContainer.start();
            System.setProperty("service.default.grouplist", "127.0.0.1:" + seataContainer.getMappedPort(8091));
            TMClient.init("test-first", "default_tx_group");
            RMClient.init("test-first", "default_tx_group");
            HikariConfig config = new HikariConfig();
            config.setDriverClassName("org.testcontainers.jdbc.ContainerDatabaseDriver");
[...]

PR #7044, opened 2024-12-03
bugfix: fix tablemeta refresh after closed

LegGasai:
#7042


rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
```diff
@@ -142,7 +161,15 @@ static class TableMetaRefreshHolder {
...
 
-
-
+        /**
+         * Helper method to determine if the exception is caused by the data source being closed.
+         *
+         * @param ex the SQLException to check
+         * @return true if the exception indicates the data source is closed; false otherwise
+         */
+        private boolean isDataSourceClosedException(SQLException ex) {
```
funky-eyes, line 171: If it's just the connection being closed, can the two be distinguished?

LegGasai: use sqlState 08006 and exception message keywords to distinguish

funky-eyes (approved):
LGTM

PR #7054, opened 2024-12-10
optimize: In file mode when the lock cannot be acquired output the holder's xid.

funky-eyes:


lightClouds917 (approved):
LGTM

PR #7064, opened 2024-12-19
optimize: remove unnecessary null checks

MaoMaoandSnail:

The return value of the ExceptionUtils.getNetExceptionType method will never be null. Remove the null check for the return value of this method.


slievrly (approved):
LGTM

PR #7071, opened 2024-12-23
如果是读写分离，在用aop设置主库的时候，会走这里，需要重新设置resourceId

liuzhebaba:


rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -210,6 +210,11 @@ public String getDbType() {
     @Override
     public ConnectionProxy getConnection() throws SQLException {
         Connection targetConnection = targetDataSource.getConnection();
+        //如果是读写分离，在用aop设置主库的时候，会走这里，需要重新设置resourceId
+        if (targetConnection!=null) {
+            this.jdbcUrl = targetConnection.getMetaData().getURL();
```
funky-eyes, line 215: 1.为什么要代理只读的从库？
2.这个方式能让兼容大部分的场景吗？只要是从库直接通过`targetConnection.getMetaData().getURL();` 得到的一定是主库的url？

Why proxy the read-only replica?
Does this approach work for most scenarios? If it's a replica, will `targetConnection.getMetaData().getURL();` always return the URL of the primary database?

funky-eyes (changes requested):

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
liuzhebaba: 我使用了两种方式进行验证，

- 一种是使用的sharding
- 一种是使用aop，进行设置主库

两种方式验证，都没有问题，我确定可以兼容几乎所有场景

liuzhebaba: 在真正执行的时候，targetConnection才是它真正使用到的connection，外面的ConnectionProxy并不一定是其真正使用的connection，所以使用targetConnection来确定resourceId是最为准确的。

funky-eyes: If the user is not using read-write separation, but instead has multiple dynamic data sources?

liuzhebaba: i am not sure this is useful for dynamic data source,
but because targetConnection is the final real connection it use,
i think if they use dynamic data source, it works too。but it need test cases to verify it.

funky-eyes: > i am not sure this is useful for dynamic data source, but because targetConnection is the final real connection it use, i think if they use dynamic data source, it works too。but it need test cases to verify it.

Why is the last connection always considered the correct address? If my dynamic data source has 3 master databases and 3 replica databases, how do I get the correct master database address? In this case, wouldn't it solve the problem to only proxy the master database? There's no need to recognize the mapping relationship between the master and replica databases.

liuzhebaba: 你是对的，这样并不行，我测试了下，100次请求中有99次都，没有问题，但是会有1次，有问题的

liuzhebaba: 谢谢您的解答，我把pr关闭了 @funky-eyes

PR #7072, opened 2024-12-23
optimize: caching the ConsistentHashSelector to avoid remapping when select every time.

navms:
Caching the ConsistentHashSelector to avoid remapping when select every time.

Caching the ConsistentHashSelector to avoid remapping when select every time.


discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/loadbalance/ConsistentHashLoadBalance.java
```diff
@@ -19,9 +19,7 @@
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.List;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.*;
```
slievrly: don't import *

slievrly (approved):
LGTM

PR #7073, opened 2024-12-23
feature:[loom] replace the usages of synchronized with ReentrantLock

### LINKED ISSUE #6971: feature: [loom] replace the usages of synchronized with ReentrantLock
## Why you need it?
To use virtual threads in JDK 21, we need to remove the usage of synchronized to prevent the issue where "a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier.

<details><summary>Details</summary>
<p>

There are two scenarios in which a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier:

When it executes code inside a synchronized block or method, or
When it executes a native method or a [foreign function](https://openjdk.java.net/jeps/424).
Pinning does not make an application incorrect, but it might hinder its scalability. If a virtual thread performs a blocking operation such as I/O or BlockingQueue.take() while it is pinned, then its carrier and the underlying OS thread are blocked for the duration of the operation. Frequent pinning for long durations can harm the scalability of an application by capturing carriers.

The scheduler does not compensate for pinning by expanding its parallelism. Instead, avoid frequent and long-lived pinning by revising synchronized blocks or methods that run frequently and guard potentially long I/O operations to use [java.util.concurrent.locks.ReentrantLock](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html) instead. There is no need to replace synchronized blocks and methods that are used infrequently (e.g., only performed at startup) or that guard in-memory operations. As always, strive to keep locking policies simple and clear.

</p>
</details> 

https://openjdk.org/jeps/444
https://openjdk.org/projects/loom/

## How it could be?

lightClouds917:

To use virtual threads in JDK 21, we need to remove the usage of synchronized to prevent the issue where "a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier."

There are two scenarios in which a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier:

When it executes code inside a synchronized block or method, or When it executes a native method or a [foreign function](https://openjdk.java.net/jeps/424).
Pinning does not make an application incorrect, but it might hinder its scalability. If a virtual thread performs a blocking operation such as I/O or BlockingQueue.take() while it is pinned, then its carrier and the underlying OS thread are blocked for the duration of the operation. Frequent pinning for long durations can harm the scalability of an application by capturing carriers.

The scheduler does not compensate for pinning by expanding its parallelism. Instead, avoid frequent and long-lived pinning by revising synchronized blocks or methods that run frequently and guard potentially long I/O operations to use [java.util.concurrent.locks.ReentrantLock](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html) instead. There is no need to replace synchronized blocks and methods that are used infrequently (e.g., only performed at startup) or that guard in-memory operations. As always, strive to keep locking policies simple and clear.

You can refer to the following documents:

https://openjdk.org/jeps/444
https://openjdk.org/projects/loom/

fixes #6971 

This pr does not include the saga module.


YvCeung:
Maybe not including * when importing the package would be more in line with programming standards.
image

lightClouds917:
> Maybe not including * when importing the package would be more in line with programming standards. image

done

lightClouds917:
The files below saga are handled separately by another pr.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
lightClouds917, line 1: The changes to this class are the synchronized removal of methods and the addition of a try in the code (resourceLock ignored = Resourcelock.obtain ()).
There are more changes here because after adding try, the internal code format is indented

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
```diff
@@ -127,10 +130,12 @@ private void xaEnd(XAXid xaXid, int flags) throws XAException {
      * @param applicationData application data
      * @throws SQLException SQLException
      */
-    public synchronized void xaCommit(String xid, long branchId, String applicationData) throws XAException {
-        XAXid xaXid = XAXidBuilder.build(xid, branchId);
-        xaResource.commit(xaXid, false);
-        releaseIfNecessary();
+    public void xaCommit(String xid, long branchId, String applicationData) throws XAException {
```
slievrly, line 133: Is there a more elegant way to write code? resourceLock.obtain() is more complex than the synchronized keyword.

funky-eyes: 我认为这点还是可以接受的，因为
I think this is still acceptable because
```
synchronized(this){
}
```
与
with
```
try(ResourceLock ignored = resourceLock.obtain()){
}
```
书写方式区别不大。
The writing style is not much different.

lightClouds917: > Is there a more elegant way to write code? resourceLock.obtain() is more complex than the synchronized keyword.

The reason i write this way is to avoid having to add unlock to finally;
`class X {
    private final ReentrantLock lock = new ReentrantLock();
    public void m() {
        lock.lock();  // block until condition holds
        try {
            // ... method body
        } finally {
            lock.unlock();//unlock
        }
    }
}`

funky-eyes (approved):
LGTM

PR #7089, opened 2025-01-02
optimize: support instance registration to the registry center

funky-eyes:

support instance registration to the registry center
支持通过传递实例注册至注册中心，由于未来要扩展元数据，而先前注册时使用的InetSocketAddress入参已经不满足需求了，故进行修改为全局实例来满足该需求，后续可通过该实例中的metadata进行注册元数据至注册中心

### DIFF
--- a/common/src/main/java/org/apache/seata/common/metadata/Instance.java
+++ b/common/src/main/java/org/apache/seata/common/metadata/Instance.java
@@ -28,8 +28,8 @@ public class Instance {
     private String namespace;
     private String clusterName;
     private String unit;
-    private Node.Endpoint control = new Node.Endpoint();
-    private Node.Endpoint transaction = new Node.Endpoint();
+    private Node.Endpoint control;
+    private Node.Endpoint transaction;
     private double weight = 1.0;
     private boolean healthy = true;
     private long term;
--- a/common/src/main/java/org/apache/seata/common/metadata/Instance.java
+++ b/common/src/main/java/org/apache/seata/common/metadata/Instance.java
@@ -14,20 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.seata.common.metadata.namingserver;
+package org.apache.seata.common.metadata;
 
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.seata.common.metadata.ClusterRole;
-import org.apache.seata.common.metadata.Node;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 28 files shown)


xingfudeshi (approved):
LGTM

discovery/seata-discovery-namingserver/src/main/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImpl.java
```diff
@@ -148,9 +147,11 @@ static NamingserverRegistryServiceImpl getInstance() {
 
     @Override
     public void register(InetSocketAddress address) throws Exception {
-        NetUtil.validAddress(address);
-        Instance instance = Instance.getInstance();
-        instance.setTransaction(new Node.Endpoint(address.getAddress().getHostAddress(), address.getPort(), "netty"));
+        register(Instance.getInstance());
```
slievrly, line 150: The IP address should be checked to ensure that it is not a loopback address.

slievrly (commented):
https://github.com/apache/incubator-seata/pull/7089/files#diff-769bc40736ea7e87dd6b0dc2e3c5e04e5f1794d58dbf5e69aa21bb44082367baR161
Note that there may be vulnerabilities when deserializing, and the simple structure recommends serializing itself.

core/src/main/java/org/apache/seata/core/rpc/netty/NettyServerBootstrap.java
```diff
@@ -170,9 +172,13 @@ public void initChannel(SocketChannel ch) {
         try {
             this.serverBootstrap.bind(port).sync();
             LOGGER.info("Server started, service listen port: {}", getListenPort());
-            InetSocketAddress address = new InetSocketAddress(XID.getIpAddress(), XID.getPort());
+            Instance instance = Instance.getInstance();
```
slievrly, line 175: I don't think it makes sense for `Instance` to be designed as a singleton.

common/src/main/java/org/apache/seata/common/metadata/Instance.java
```diff
@@ -14,27 +14,22 @@
...
-import static org.apache.seata.common.util.CollectionUtils.mapToJsonString;
-
-
 public class Instance {
     private String namespace;
     private String clusterName;
     private String unit;
-    private Node.Endpoint control = new Node.Endpoint();
-    private Node.Endpoint transaction = new Node.Endpoint();
+    private Node.Endpoint control;
```
slievrly, line 31: What is the specific meaning of `Node.Endpoint control`?

Issue #7058, opened 2024-12-16
RemotingClient does not support setting a listener


funky-eyes:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description

RemotingClient不支持设置listener，当消息发出后channel断开时，无法通过listener进行快速响应失败。
RemotingClient does not support setting a listener. When the channel is disconnected after the message is sent, it cannot quickly respond to the failure through the listener

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

Minimal yet complete reproducer code (or URL to code):

### Ⅵ. Environment:

- JDK version(e.g. `java -version`):
- Seata client/server version:  
- Database version:
- OS(e.g. `uname -a`):
- Others:

lyl2008dsg:
I would like to claim this task

YongGoose:
@lyl2008dsg 

Any updates?
If you're too busy to handle this issue, I'll take it.

YongGoose:
@funky-eyes 

I worked on this in my personal repository https://github.com/YongGoose/incubator-seata/commit/9ae0709ddab229a058bcc31885f84519033cb1e0.

This task adds a `channel event listener` feature to Seata's `RemotingClient`.
With this enhancement, the system can now immediately detect events such as `network disconnections` or `message delivery failures` through `callbacks`, allowing custom logic—like retries or alerts—to be triggered as needed.
In short, it improves the handling of network issues and connection problems in a distributed transaction environment, making the system more responsive and flexible.

What do you think?

YongGoose:
@funky-eyes 

I've created a PR!
- https://github.com/apache/incubator-seata/pull/7337

Please take a look and share your feedback when you have time. 🙂

PR #7095, opened 2025-01-04
optimize:  support fast fail when channel disconnect after message sent  for RemotingClient

lyl2008dsg:

This PR enhances the RemotingClient by implementing a mechanism to quickly fail pending requests when a channel disconnects after a message has been sent. The key changes include:

#7058

wait a moment


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -193,6 +198,7 @@ public Object sendSyncRequest(Channel channel, Object msg) throws TimeoutExcepti
             return null;
         }
         RpcMessage rpcMessage = buildRequestMessage(msg, ProtocolConstants.MSGTYPE_RESQUEST_SYNC);
+        channelToRequestIds.computeIfAbsent(channel, ch -> ConcurrentHashMap.newKeySet()).add(rpcMessage.getId());
```
Bughue, line 201: Should the computeIfAbsent operation be extracted method? and integrated with super.sendSync? Avoid code duplication and missing calls

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -222,6 +229,7 @@ public void sendAsyncRequest(Channel channel, Object msg) {
     public void sendAsyncResponse(String serverAddress, RpcMessage rpcMessage, Object msg) {
         RpcMessage rpcMsg = buildResponseMessage(rpcMessage, msg, ProtocolConstants.MSGTYPE_RESPONSE);
         Channel channel = clientChannelManager.acquireChannel(serverAddress);
+        channelToRequestIds.computeIfAbsent(channel, ch -> ConcurrentHashMap.newKeySet()).add(rpcMessage.getId());
```
Bughue: Is fast-failure necessary for AsyncSend? Is there Future waiting in this case?

funky-eyes, line 232: If the request is successful, how will this element be deleted?

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -501,4 +513,59 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+            }
+            return;
+        }
+
+        if (requestId == null) {
+            LOGGER.warn("Attempted to remove a null requestId from channel {}.", channel);
+            return;
+        }
+
+        channelToRequestIds.computeIfPresent(channel, (ch, requestIds) -> {
```
funky-eyes, line 556: Assuming an inactive event occurs on channelread, is it possible for this channel to remain permanently in channelToRequestIds?

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -501,4 +513,59 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+            boolean removed = requestIds.remove(requestId);
+            if (removed) {
+                LOGGER.debug("Removed requestId {} from channel {}.", requestId, ch);
```
funky-eyes: You need to determine whether the current log level is debug level or not.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -501,4 +513,59 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+            } else {
+                LOGGER.warn("Attempted to remove non-existing requestId {} from channel {}.", requestId, ch);
+            }
+
+            if (requestIds.isEmpty()) {
+                LOGGER.debug("No more requestIds associated with channel {}. Channel removed from mapping.", ch);
```
funky-eyes: You need to determine whether the current log level is debug level or not.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lyl2008dsg: You're right that extracting computeIfAbsent into super.sendSync would reduce duplication and missed calls. 
However, since this is a client-side issue, I won't move it to AbstractNettyRemoting for now.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lyl2008dsg: After reviewing the code, it's clear that AsyncSend doesn't wait on a Future—it handles outcomes via callbacks. Thus, fast-failure isn't needed in this context.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lyl2008dsg: Yeah, you're right.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lyl2008dsg: Yeah, you're right.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lyl2008dsg: After reviewing the code, it's clear that AsyncSend doesn't wait on a Future—it handles outcomes via callbacks. Thus, fast-failure isn't needed in this context.

funky-eyes:
You need to merge the latest 2.x branch code and submit it.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -501,4 +512,71 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+     * Iterates over futures and marks all MessageFutures associated with the specified Channel as failed.
+     *
+     * @param channel The Channel that has been disconnected or encountered an exception.
+     * @param cause   The reason for the failure.
+     */
+    private void failFuturesForChannel(Channel channel, Throwable cause) {
+        Set<Integer> requestIds = channelToRequestIds.remove(channel);
+        if (requestIds != null) {
+            for (Integer requestId : requestIds) {
+                MessageFuture future = futures.remove(requestId);
```
funky-eyes, line 525: The childToParentMap has not been handled together.

PR #7114, opened 2025-01-15
feature: support raft mode registry to namingserver

funky-eyes:


ggbocoder:
please add signature

xingfudeshi (approved):
LGTM

discovery/seata-discovery-namingserver/src/main/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImpl.java
```diff
@@ -156,6 +156,9 @@ public void register(Instance instance) throws Exception {
         doRegister(instance, getNamingAddrs());
     }
 
+    public void doRegister(List<Instance> instance, List<String> urlList) {
+
```
xjlgod, line 160: why this method is not implemented?

namingserver/src/main/java/org/apache/seata/namingserver/manager/NamingManager.java
```diff
@@ -152,7 +154,9 @@ public Result<String> createGroup(String namespace, String vGroup, String cluste
             LOGGER.error("no instance in cluster {}", clusterName);
             return new Result<>("301", "no instance in cluster" + clusterName);
         } else {
-            Node node = nodeList.get(0);
+            Node node =
+                nodeList.stream().filter(n -> n.getRole() == ClusterRole.LEADER || n.getRole() == ClusterRole.MEMBER)
```
xjlgod, line 158: why leader and memeber can be selected?

xjlgod:
Maybe we can add some example to show how to config raft mode of namingserver?

namingserver/src/main/java/org/apache/seata/namingserver/manager/NamingManager.java
funky-eyes: For a Raft cluster, you need to find the leader to create the group, whereas for a regular cluster, you need to find a member to create the group.

funky-eyes:
> Maybe we can add some example to show how to config raft mode of namingserver?

Integration testing is required, but the Seata repository currently does not support it. However, you can compile the code locally and test it using a combination of client, server, and naming server.

discovery/seata-discovery-namingserver/src/main/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImpl.java
funky-eyes: This is a reserved interface because, when multi-raft is implemented, a node will have multiple instances.

PR #7117, opened 2025-01-16
bugfix: prefix: [seata.server.raft.ssl] should not be null

funky-eyes:


seata-spring-autoconfigure/seata-spring-autoconfigure-server/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/server/raft/ServerRaftSSLClientProperties.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties(prefix = SERVER_RAFT_SSL_CLIENT_KEYSTORE_PREFIX)
+public class ServerRaftSSLClientProperties {
+
+    private String path = "ssl/cbolt.pfx";
+
+    private String password = "password";
```
slievrly: There should be no default values for sensitive information.

slievrly (approved):
LGTM

Issue #6774, opened 2024-08-23
optimize seata-client TM&RM netty eventloop


funky-eyes:
## Why you need it?
构建多个eventloopgroup时，会构建对应的平台线程，会增加线程切换的开销，而TM和RM都属于client，如果一个应用恰好是TM和RM正好可以降低额外的线程数。
When building multiple EventLoopGroups, corresponding platform threads are created, which increases the overhead of context switching. Since both TM and RM are clients, if an application happens to be both TM and RM, it can reduce the number of additional threads.

## How it could be?
A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

zhengkezhou1:
Hi @funky-eyes, can i pick up this ticket?

funky-eyes:
> Hi @funky-eyes, can i pick up this ticket?

Of course, looking forward to your PR.

YongGoose:
@funky-eyes 

If there are no updates, i'd like to take this issue.

funky-eyes:
> [@funky-eyes](https://github.com/funky-eyes)
> 
> If there are no updates, i'd like to take this issue.

Of course，I have assigned it to you and look forward to your PR

YongGoose:
> > [@funky-eyes](https://github.com/funky-eyes)
> > If there are no updates, i'd like to take this issue.
> 
> Of course，I have assigned it to you and look forward to your PR

Thank you :)
Since we are in a similar time zone, it makes communication easier. 😂

YongGoose:
@funky-eyes 
I’m leaving this comment as I had a question while designing this issue.

I understand that the core idea of this issue is to ensure that both `TM` and `RM` clients share the same `EventLoopGroup`.

With that in mind, I thought it would be best to create an `EventLoopGroup` in `GlobalTransactionScanner` and pass it to both `TMClient` and `RMClient` during their initialization via the `initClient` method.
- https://github.com/apache/incubator-seata/blob/6b25b21b6de91c9a9c47d7373943cdf2549c55af/spring/src/main/java/org/apache/seata/spring/annotation/GlobalTransactionScanner.java#L233

However, I couldn’t find any part of the `TMClient` and `RMClient` `init` or the `getInstance` method that directly handles `EventLoopGroup`.
- https://github.com/apache/incubator-seata/blob/6b25b21b6de91c9a9c47d7373943cdf2549c55af/tm/src/main/java/org/apache/seata/tm/TMClient.java#L45
[...]

PR #7133, opened 2025-01-31
feature: Implement scheduled handling for end status transaction

### LINKED ISSUE #6875: TimeoutRollbacked and Rollbacked residues

### Ⅰ. Issue Description
TimeoutRollbacked和Rollbacked在db，redis等网络出现问题时，会删除失败，除非重启server，否则这个数据将一直残留，file下如盘满了等情况change status成功后，删除可能就失败了，也会残留
目前DefaultCoordinator中针对committed这种状态是有补偿的，对TimeoutRollbacked和Rollbacked没有任何补偿处理。
In cases where there are network issues with databases, Redis, etc., data may be deleted unsuccessfully and will remain unless the server is restarted. If the disk is full or there are other issues, even after a successful change of status, deletion may fail, leaving the data residual. Currently, in DefaultCoordinator, there is compensation handling for the committed state, but there is no compensation handling for the TimeoutRollbacked and Rollbacked states.
```
    private final GlobalStatus[] retryRollbackingStatuses = new GlobalStatus[] {
        GlobalStatus.TimeoutRollbacking,
        GlobalStatus.TimeoutRollbackRetrying, GlobalStatus.RollbackRetrying};

    private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
```

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- Database version:
- OS(e.g. `uname -a`):
- Others:

YongGoose:

- Add compensation handling for TimeoutRollbacked and Rollbacked states

fixes #6875


server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,7 +189,11 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
 
     private final GlobalStatus[] retryRollbackingStatuses = new GlobalStatus[] {
         GlobalStatus.TimeoutRollbacking,
-        GlobalStatus.TimeoutRollbackRetrying, GlobalStatus.RollbackRetrying};
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying,
+        GlobalStatus.TimeoutRollbacked,
+        GlobalStatus.Rollbacked
```
funky-eyes: 由于rollback的概率是非常小的，并且rollbacked和TimeoutRollbacked的状态残留的概率更低，故这两种状态是特意放置在server重启时进行处理。
Since the probability of rollback is very low, and the chances of the rollbacked and TimeoutRollbacked states persisting are even lower, these two states are specifically handled during server restart.

YongGoose: > 由于rollback的概率是非常小的，并且rollbacked和TimeoutRollbacked的状态残留的概率更低，故这两种状态是特意放置在server重启时进行处理。 Since the probability of rollback is very low, and the chances of the rollbacked and TimeoutRollbacked states persisting are even lower, these two states are specifically handled during server restart.

Thank you for your comment :)

I didn’t fully understand.

```
Currently, in DefaultCoordinator, there is compensation handling for the committed state, but there is no compensation handling for the TimeoutRollbacked and Rollbacked states.
```

Based on the comment above, I added these two cases to the array for management.
If any changes are needed, could you please provide additional explanations?

YongGoose:
@funky-eyes 

In the issue description, it was mentioned that the state remains unchanged unless the server is `restarted`.
This means that the data will persist unless the server is restarted.

So, aside from handling it specifically during a server restart, I believe additional measures should be taken.

WDYT?

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
funky-eyes: Since the probability of a transaction rollback occurring is very low, the likelihood of the rollbacked intermediate state persisting is even lower. After the rollbacking state, it transitions to rollbacked and is then deleted. These two actions are separated by just one I/O operation, meaning that once the status is successfully updated to rollbacked, the transaction will be immediately deleted, unless a network or storage-side exception occurs during deletion, preventing the transaction from being deleted. Otherwise, there will be no rollbacked transactions left.
The commit of a transaction is a high-probability event, so the likelihood of the committed state persisting is much higher than that of rollbacked. To prevent storage-side pressure and meet performance requirements, committed transactions need to be processed during the server's runtime, while the low-probability rollbacked transactions are handled during the server startup.
Since both of these probabilities are very low, the final result is that they are both deleted. If you think it is necessary to handle rollbacked transactions during runtime, I suggest that you create a separate scheduled task to batch process transactions in the rollbacked and committed states, rather than blocking transactions that urgently need to be rolled back and retried.

YongGoose: > If you think it is necessary to handle rollbacked transactions during runtime, I suggest that you create a separate scheduled task to batch process transactions in the rollbacked and committed states, rather than blocking transactions that urgently need to be rolled back and retried.

Thank you for your suggestion!

The idea you proposed is very interesting, and I’m planning to work on it. :)

Additionally, I’ve posted a message on the Seata project’s issue list and mailing list regarding this topic.
I’ll leave the links below—could you kindly take a look when you have a chance?
- https://lists.apache.org/thread/xlkshkqs5v2hflp10jgz418kos8vwrsc
- https://github.com/apache/incubator-seata/issues/6774

Thanks again! 🚀

funky-eyes:
Do you use the DingTalk app?

YongGoose:
> Do you use the DingTalk app?

Unfortunately, no.. 

However, if using DingTalk allows for better communication, I’m willing to install it!

funky-eyes:
> > Do you use the DingTalk app?
> 
> Unfortunately, no..
> 
> However, if using DingTalk allows for better communication, I’m willing to install it!

Of course, email is also a great communication channel, but I would like to invite you to join the Seata contributors' DingTalk group if you're willing to install it. If you install it, please send your DingTalk ID to my email, and I will invite you to join the Seata contributors' group.

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: @funky-eyes 

Done in https://github.com/apache/incubator-seata/pull/7133/commits/fc28796f8009fdecb1dd554c91d461a623d1a050 !
PTAL ⚡️

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
```diff
@@ -204,9 +204,17 @@ public boolean isTimeout() {
...
+     *
+     * If the global session's status is "Rollbacked" or "TimeoutRollbacked", it returns the negative time since the session began,
+     * indicating that the session is already dead.
+     * For other statuses, it returns the remaining time until the session reaches the retry dead threshold.
+     *
      * @return time to dead session. if not greater than 0, then deadSession
      */
     public long timeToDeadSession() {
+        if (this.status == GlobalStatus.Rollbacked || this.status == GlobalStatus.TimeoutRollbacked) {
+            return beginTime - System.currentTimeMillis();
```
funky-eyes: The beginTime does not change after the creation of the global session, which will cause any global session to be added to the needDoRollbackedSessions list.

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,12 +197,15 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
...
-        GlobalStatus.TimeoutRollbackRetrying, GlobalStatus.RollbackRetrying};
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying
+    };
 
     private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
 
     private final GlobalStatus[] rollbackingStatuses = new GlobalStatus[] {GlobalStatus.Rollbacking};
     private final GlobalStatus[] committingStatuses = new GlobalStatus[] {GlobalStatus.Committing};
+    private final GlobalStatus[] rollbackedStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.TimeoutRollbacked};
```
funky-eyes: Why not process transactions in other end states together?

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -658,6 +724,8 @@ public void init() {
         rollbackingSchedule(0);
 
         committingSchedule(0);
+
+        rollbackedSchedule(0);
```
funky-eyes: The methods `rollbackingSchedule`, `committingSchedule`, and `rollbackedSchedule` need to use the `SessionHolder.distributedLockAndExecute` method to ensure thread safety in a distributed environment.

YongGoose:
```java
boolean called = SessionHolder.distributedLockAndExecute(ROLLBACKED, this::handleRollbackedByScheduled);
```

Aren’t we already using `SessionHolder.distributedLockAndExecute`?

Are you suggesting that the `rollbackingSchedule`, `committingSchedule`, and `rollbackedSchedule` methods should be explicitly called using the `SessionHolder.distributedLockAndExecute` API?
> It seems difficult because distributedLockAndExecute requires a parameterless function.

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
YongGoose: I implemented it this way to set a slightly shorter interval for sessions in the `end` state.

If possible, could you please provide a more detailed explanation? 🙂

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: I will include the `CommitFailed`, `RollbackFailed`, `TimeoutRollbackFailed`, `Finished`, `CommitRetryTimeout`, and `RollbackRetryTimeout` states :)

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
funky-eyes: The beginTime is set when the transaction is created. When the Seata server's business thread changes the transaction status from "Rollbacking" to "Rollbacked," your timeToDeadSession method directly subtracts the current timestamp from beginTime, which will inevitably result in a negative value. As a result, the business thread will delete the transaction, and your scheduled task will also delete the transaction, leading to a concurrency issue and thread safety problem. I believe that beginTime should be increased by the corresponding transaction timeout, plus an additional buffer time to ensure thread safety.

YongGoose: Then, would it be appropriate to create a `RETRY_DEAD_THRESHOLD` specifically for the end state?

> Of course, this timeToDeadSession method is more relevant for ongoing statuses like "rollbacking" or "committing." For transactions in the "end" status, I believe the interval can be shortened.

As suggested in the discussion, since we can reduce the interval in the end state, I believe decreasing the size of `RETRY_DEAD_THRESHOLD` could be a reasonable approach.

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
funky-eyes: Sorry, I misread it. The issue I mentioned does not exist.

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
YongGoose: > The beginTime is set when the transaction is created. When the Seata server's business thread changes the transaction status from "Rollbacking" to "Rollbacked," your timeToDeadSession method directly subtracts the current timestamp from beginTime, which will inevitably result in a negative value. As a result, the business thread will delete the transaction, and your scheduled task will also delete the transaction, leading to a concurrency issue and thread safety problem. I believe that beginTime should be increased by the corresponding transaction timeout, plus an additional buffer time to ensure thread safety.

Since our comments were written at the same time, I didn’t see yours when I was writing mine. 😅

Thank you for the additional explanation!
In that case, I will adjust the execution interval for sessions in the `end state` to be shorter using an `if` condition.

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: > Sorry, I misread it. The issue I mentioned does not exist.

No worries! 🙂

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
funky-eyes: > RETRY_DEAD_THRESHOLD

I think this can be done this way

YongGoose: > > RETRY_DEAD_THRESHOLD
> 
> I think this can be done this way

Then no changes required?

> Then, would it be appropriate to create a RETRY_DEAD_THRESHOLD specifically for the end state?

or you mean this way?

funky-eyes: Create a corresponding RETRY_DEAD_THRESHOLD for the end state

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/a9ec30f61e8d70815ca73e21f7dd22c54f5bb5ab !

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/997775c07c4e946b39e4cc85cd9a42dc79c3d145 !

PTAL ⚡️

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,12 +197,17 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
...
-        GlobalStatus.TimeoutRollbackRetrying, GlobalStatus.RollbackRetrying};
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying
+    };
 
     private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
 
     private final GlobalStatus[] rollbackingStatuses = new GlobalStatus[] {GlobalStatus.Rollbacking};
     private final GlobalStatus[] committingStatuses = new GlobalStatus[] {GlobalStatus.Committing};
+    private final GlobalStatus[] rollbackedStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.TimeoutRollbacked,
```
funky-eyes:
```suggestion
    private final GlobalStatus[] endStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.TimeoutRollbacked,
```

common/src/main/java/org/apache/seata/common/ConfigurationKeys.java
```diff
@@ -475,6 +475,11 @@ public interface ConfigurationKeys {
      */
     String ROLLBACKING_RETRY_PERIOD = RECOVERY_PREFIX + "rollbackingRetryPeriod";
 
+    /**
+     * The constant ROLLBACKED_RETRY_PERIOD.
+     */
+    String ROLLBACKED_RETRY_PERIOD = RECOVERY_PREFIX + "rollbackedRetryPeriod";
```
funky-eyes: https://github.com/apache/incubator-seata/blob/2.x/script/config-center/config.txt
https://github.com/apache/incubator-seata/blob/2.x/server/src/main/resources/application.example.yml
https://github.com/apache/incubator-seata/blob/2.x/server/src/main/resources/application.raft.example.yml
https://github.com/apache/incubator-seata/blob/2.x/seata-spring-autoconfigure/seata-spring-autoconfigure-server/src/main/java/org/apache/seata/spring/boot/autoconfigure/properties/server/ServerProperties.java
This configuration item needs to be supplemented in the above file, and I think it is no longer appropriate to name this configuration starting with `rollback`, because it is a configuration name for handling transactions in the `end` state..

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
```diff
@@ -203,13 +211,35 @@ public boolean isTimeout() {
...
+    private boolean isEndStatus() {
+        EnumSet<GlobalStatus> deadStatuses = EnumSet.of(
+            GlobalStatus.Rollbacked,
+            GlobalStatus.TimeoutRollbacked,
+            GlobalStatus.CommitFailed,
+            GlobalStatus.RollbackFailed,
+            GlobalStatus.TimeoutRollbackFailed,
+            GlobalStatus.Finished,
+            GlobalStatus.CommitRetryTimeout,
+            GlobalStatus.RollbackRetryTimeout
```
funky-eyes:
```suggestion
            GlobalStatus.Finished
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +645,60 @@ private void committingSchedule(long delay) {
             delay, TimeUnit.MILLISECONDS);
     }
 
+    /**
+     * Handle rollbacked by scheduled.
+     */
+    protected void handleRollbackedByScheduled() {
```
funky-eyes:
```suggestion
    protected void handleEndStatesByScheduled() {
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +645,60 @@ private void committingSchedule(long delay) {
...
+            try {
+                if (isRetryTimeout(now, MAX_ROLLBACK_RETRY_TIMEOUT, rollbackedSession.getBeginTime())) {
+
+                    // commit retry timeout event
+                    SessionHelper.endRollbackFailed(rollbackedSession, true, true);
+
+                    //The function of this 'return' is 'continue'.
+                    return;
+                }
+                core.doGlobalRollback(rollbackedSession, true);
```
funky-eyes: It should no longer perform rollback processing. For different states of transactions in the end phase, call the `SessionHelper.endRollbacked` and `SessionHelper.endCommitted` methods respectively.

common/src/main/java/org/apache/seata/common/ConfigurationKeys.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/932c51fe7dd96c38d47f06e4e2ee375a9705baa6 !

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/a7a9e5255b8f235758e9397ef29872f2386735d0 !

YongGoose: @funky-eyes 

When you have time, I would appreciate it if you could review this comment and the issue below.
- https://github.com/apache/incubator-seata/issues/6774

script/config-center/config.txt
```diff
@@ -153,6 +153,7 @@ store.redis.queryLimit=100
 server.recovery.committingRetryPeriod=1000
 server.recovery.asynCommittingRetryPeriod=1000
 server.recovery.rollbackingRetryPeriod=1000
+server.recovery.endstatusRetryPeriod=1000
```
funky-eyes, line 156:
```suggestion
server.recovery.endstatusRetryPeriod=10000
```

common/src/main/java/org/apache/seata/common/DefaultValues.java
```diff
@@ -397,6 +402,11 @@ public interface DefaultValues {
      */
     int DEFAULT_ROLLBACKING_RETRY_PERIOD = 1000;
 
+    /**
+     * the constant DEFAULT_END_STATUS_RETRY_PERIOD
+     */
+    int DEFAULT_END_STATUS_RETRY_PERIOD = 1000;
```
funky-eyes: I believe that for transactions in the end state, they should not be processed with such high frequency, because the probability of end state transactions not being processed by the business thread is very low. Thirty seconds might be more appropriate.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/05735298ecc9a2f49f05ec9f413289f98931ed49 :)

changes/zh-cn/2.x.md
```diff
@@ -9,6 +9,7 @@
 - [[#7037](https://github.com/apache/incubator-seata/pull/7037)] 支持UndoLog的fury序列化方式
 - [[#7069](https://github.com/apache/incubator-seata/pull/7069)] Raft集群模式支持地址转换
 - [[#7038](https://github.com/apache/incubator-seata/pull/7038)] 支持Fury序列化器
+- [[#7133](https://github.com/apache/incubator-seata/pull/7133)] 实现对已回滚和超时回滚事务的定时处理
```
```suggestion
- [[#7133](https://github.com/apache/incubator-seata/pull/7133)] 实现对残留的end状态事务定时处理
```

common/src/main/java/org/apache/seata/common/DefaultValues.java
```diff
@@ -397,6 +402,11 @@ public interface DefaultValues {
      */
     int DEFAULT_ROLLBACKING_RETRY_PERIOD = 1000;
 
+    /**
+     * the constant DEFAULT_END_STATUS_RETRY_PERIOD
+     */
+    int DEFAULT_END_STATUS_RETRY_PERIOD = 30;
```
funky-eyes:
```suggestion
    int DEFAULT_END_STATUS_RETRY_PERIOD = 30 * 1000;
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +644,64 @@ private void committingSchedule(long delay) {
...
+        SessionHelper.forEach(needDoRollbackedSessions, rollbackedSession -> {
+            try {
+                if (isRetryTimeout(now, MAX_ROLLBACK_RETRY_TIMEOUT, rollbackedSession.getBeginTime())) {
+                    handleEndStateSession(rollbackedSession);
+                }
+            } catch (TransactionException ex) {
+                LOGGER.error("Failed to handle rollbacked [{}] {} {}", rollbackedSession.getXid(), ex.getCode(), ex.getMessage());
+            }
+        });
+        rollbackedSchedule(delay);
```
funky-eyes:
```suggestion
        endSchedule(delay);
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +644,64 @@ private void committingSchedule(long delay) {
...
+    }
+
+    private void handleEndStateSession(GlobalSession globalSession) throws TransactionException {
+        GlobalStatus status = globalSession.getStatus();
+
+        if (status == GlobalStatus.CommitFailed) {
```
funky-eyes: The commit state is not limited to just one, and the rollback state is the same. I suggest putting the handling of various end states directly in the SessionHelper, and then calling it from handleEndStateSession. The only input parameter needed is a globalsession

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +644,64 @@ private void committingSchedule(long delay) {
...
+            SessionHelper.endCommitted(globalSession, true);
+            return;
+        }
+        SessionHelper.endRollbacked(globalSession, true);
+    }
+
+    private void rollbackedSchedule(long delay) {
```
funky-eyes:
```suggestion
    private void endSchedule(long delay) {
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -631,6 +644,64 @@ private void committingSchedule(long delay) {
...
+        syncProcessing.schedule(
+            () -> {
+                boolean called = SessionHolder.distributedLockAndExecute(ROLLBACKED, this::handleEndStatesByScheduled);
```
funky-eyes:
```suggestion
                boolean called = SessionHolder.distributedLockAndExecute(END, this::handleEndStatesByScheduled);
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: As I understand it, instead of directly managing the transaction end states with conditional statements in `handleEndStateSession`, the `responsibility` is being delegated to `SessionHelper`.

Is my understanding correct?

Below is the implementation based on my understanding.

```java
private void handleEndStateSession(GlobalSession globalSession) throws TransactionException {
    SessionHelper.processEndState(globalSession);
}
```

**The state values are just examples.**
```java
public class SessionHelper {

    public static void processEndState(GlobalSession globalSession) throws TransactionException {
        GlobalStatus status = globalSession.getStatus();

        switch (status) {
            case Committed:
            case CommitSuccess:
                endCommitted(globalSession, false);
                break;

            case CommitFailed:
            case CommitPending:
                endCommitted(globalSession, true);
                break;

            case Rollbacked:
            case RollbackSuccess:
                endRollbacked(globalSession, false);
                break;

            case RollbackFailed:
            case RollbackPending:
                endRollbacked(globalSession, true);
                break;

            default:
                log.warn("Unknown transaction state: {}", status);
        }
    }
}
```

funky-eyes: Yes, I think it should be this way, encapsulating the processing logic in the SessionHelper to hide the internal implementation from the outside.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7133/commits/5837ae6b0e090af8cbe435f47528a40f68d6feaf !

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
YongGoose: @funky-eyes 

Is there a specific reason why you used `state` instead of `status`?

YongGoose:
@funky-eyes 

You can see in the comment that the word `rollbacked` has been changed to `status`.  https://github.com/apache/incubator-seata/pull/7133/commits/113a7fd66cfad29f76c64dd577fb86b021619358

Please take a look and provide your feedback.

server/src/main/java/org/apache/seata/server/session/SessionHelper.java
```diff
@@ -454,6 +454,25 @@ public static void removeAllBranch(GlobalSession globalSession, boolean isAsync)
         }
     }
 
+    public static void processEndState(GlobalSession globalSession) throws TransactionException {
+        GlobalStatus globalStatus = globalSession.getStatus();
+
+        switch (globalStatus) {
+            case CommitFailed:
```
funky-eyes:
```suggestion
            case Committed:
```

server/src/main/java/org/apache/seata/server/session/SessionHelper.java
```diff
@@ -454,6 +454,25 @@ public static void removeAllBranch(GlobalSession globalSession, boolean isAsync)
...
+            case Finished:
+                endCommitted(globalSession, true);
+                return;
+            case Rollbacked:
+            case TimeoutRollbacked:
+            case RollbackFailed:
+            case TimeoutRollbackFailed:
```
funky-eyes:
```suggestion
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,12 +197,16 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
...
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying
+    };
 
     private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
 
     private final GlobalStatus[] rollbackingStatuses = new GlobalStatus[] {GlobalStatus.Rollbacking};
     private final GlobalStatus[] committingStatuses = new GlobalStatus[] {GlobalStatus.Committing};
+    private final GlobalStatus[] endStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.TimeoutRollbacked,
+        GlobalStatus.CommitFailed, GlobalStatus.RollbackFailed, GlobalStatus.TimeoutRollbackFailed, GlobalStatus.Finished};
```
funky-eyes:
```suggestion
    private final GlobalStatus[] endStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.Committed, GlobalStatus.Finished};
```

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,12 +197,16 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
 
     private final GlobalStatus[] retryRollbackingStatuses = new GlobalStatus[] {
         GlobalStatus.TimeoutRollbacking,
-        GlobalStatus.TimeoutRollbackRetrying, GlobalStatus.RollbackRetrying};
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying
+    };
 
     private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
```
funky-eyes:
```suggestion
    private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying};
```

funky-eyes (commented):
I believe the `committed` state should be handled within the `end` state task, and transactions in a failed state should not be retried. Failed transactions need to be addressed by the operations team, rather than being handled by scheduled tasks.

YongGoose:
> I believe the `committed` state should be handled within the `end` state task, and transactions in a failed state should not be retried. Failed transactions need to be addressed by the operations team, rather than being handled by scheduled tasks.

Good idea!
I agree with you

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
```diff
@@ -203,13 +211,33 @@ public boolean isTimeout() {
...
 
+    private boolean isEndStatus() {
+        EnumSet<GlobalStatus> deadStatuses = EnumSet.of(
+            GlobalStatus.Rollbacked,
+            GlobalStatus.TimeoutRollbacked,
+            GlobalStatus.CommitFailed,
+            GlobalStatus.RollbackFailed,
+            GlobalStatus.TimeoutRollbackFailed,
+            GlobalStatus.Finished
+        );
```
YongGoose:
```suggestion
        EnumSet<GlobalStatus> endStatuses = EnumSet.of(
            GlobalStatus.Rollbacked,
            GlobalStatus.TimeoutRollbacked,
            GlobalStatus.Committed,
            GlobalStatus.Finished
        );
```

@funky-eyes 

WDYT?

YongGoose:
> I believe the `committed` state should be handled within the `end` state task, and transactions in a failed state should not be retried. Failed transactions need to be addressed by the operations team, rather than being handled by scheduled tasks.

That sounds like a good point.

However, while going through your comment, a few questions came to mind:
1. Could you explain the situations where retrying the `endStatus` session would be necessary?
2. If that approach is taken, does it still fail to resolve the `issue` mentioned?
- https://github.com/apache/incubator-seata/issues/6875
3. Lastly, was this PR ultimately decided to simply retry the end status session, or was there another approach considered?

> I completely agree that failed transactions should not be retried but should instead be handled by the operations team. However, I’m curious about how the issue is addressed, so I wanted to ask.

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
funky-eyes: This is okay

funky-eyes:
For the reasons I mentioned earlier:
1. IO exceptions, such as disk or network issues
2. Process crashes (whether it's the seata-server or the database, etc.)

In the above points or some examples I did not mention, transactions are not deleted in the end state, leaving residual data.

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
funky-eyes: Sorry, I think I missed TimeoutRollbacked. Since TimeoutRollbacked is not a failed end state, it should be handled by scheduled tasks.

YongGoose: No worries 😉
I'll fix it.

server/src/main/java/org/apache/seata/server/coordinator/DefaultCoordinator.java
```diff
@@ -189,12 +197,15 @@ public class DefaultCoordinator extends AbstractTCInboundHandler implements Tran
...
+        GlobalStatus.TimeoutRollbackRetrying,
+        GlobalStatus.RollbackRetrying
+    };
 
-    private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying, GlobalStatus.Committed};
+    private final GlobalStatus[] retryCommittingStatuses = new GlobalStatus[] {GlobalStatus.CommitRetrying};
 
     private final GlobalStatus[] rollbackingStatuses = new GlobalStatus[] {GlobalStatus.Rollbacking};
     private final GlobalStatus[] committingStatuses = new GlobalStatus[] {GlobalStatus.Committing};
+    private final GlobalStatus[] endStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.Committed, GlobalStatus.Finished};
```
YongGoose:
```suggestion
    private final GlobalStatus[] endStatuses = new GlobalStatus[] {GlobalStatus.Rollbacked, GlobalStatus.TimeoutRollbacked, GlobalStatus.Committed, GlobalStatus.Finished};
```

server/src/main/java/org/apache/seata/server/session/SessionHelper.java
```diff
@@ -454,6 +454,22 @@ public static void removeAllBranch(GlobalSession globalSession, boolean isAsync)
...
+    public static void processEndState(GlobalSession globalSession) throws TransactionException {
+        GlobalStatus globalStatus = globalSession.getStatus();
+
+        switch (globalStatus) {
+            case Committed:
+            case Finished:
+                endCommitted(globalSession, true);
+                return;
+            case Rollbacked:
+                endRollbacked(globalSession, true);
```
YongGoose:
```suggestion
            case Rollbacked:
            case TimeoutRollbacked:
                endRollbacked(globalSession, true);
```

YongGoose (commented):
@funky-eyes 

Please take a look :)
If everything looks good, please proceed with committing the suggestion.

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
```diff
@@ -203,13 +211,31 @@ public boolean isTimeout() {
...
     }
 
+    private boolean isEndStatus() {
+        EnumSet<GlobalStatus> endStatuses = EnumSet.of(
+            GlobalStatus.Rollbacked,
+            GlobalStatus.TimeoutRollbacked,
+            GlobalStatus.Committed,
+            GlobalStatus.Finished
+        );
+        return deadStatuses.contains(this.status);
```
YongGoose:
```suggestion
        return endStatuses.contains(this.status);
```

YongGoose:
@funky-eyes 

Is there anything else left to do?

funky-eyes (approved):
LGTM

PR #7138, opened 2025-02-07
optimize: remove org.eclipse.jetty dependency

### LINKED ISSUE #7137: jetty 9.4.38 has 5 CVE security issues associated with it

### Ⅰ. Issue Description

https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-server/9.4.38.v20210224

https://github.com/apache/incubator-seata/blob/67548dfa69f988bd3544075f39078c3cc578bb36/dependencies/pom.xml#L136

It would be better to use the latest 9.4 release to avoid the security risks.

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- Database version:
- OS(e.g. `uname -a`):
- Others:

YongGoose:

fixes #7137

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -29,7 +29,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#7093](https://github.com/apache/incubator-seata/pull/7093)] add a test workflow for JDK 21
 - [[#7131](https://github.com/apache/incubator-seata/pull/7131)] Remove org.codehaus.jackson dependency
 - [[#7134](https://github.com/apache/incubator-seata/pull/7134)] upgrade tomcat-embed to 9.0.98
-- [[#7138]](https://github.com/apache/incubator-seata/pull/7138) upgrade jetty to 9.4.57.v20241219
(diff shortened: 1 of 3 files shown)


funky-eyes (commented):
I want to know where jetty is used, and whether we can remove it directly?

slievrly:
> I want to know where jetty is used, and whether we can remove it directly?

The org.eclipse.jetty:* dependency is only referenced in the seata-discovery-zk module, so it looks like it can be removed. It was probably needed back when org.apache.zookeeper:* hadn't been replaced with org.apache.curator:* yet.

YongGoose:
> > I want to know where jetty is used, and whether we can remove it directly?
> 
> The org.eclipse.jetty:* dependency is only referenced in the seata-discovery-zk module, so it looks like it can be removed. It was probably needed back when org.apache.zookeeper:* hadn't been replaced with org.apache.curator:* yet.

Done in https://github.com/apache/incubator-seata/pull/7138/commits/de5ba5b471e9710b653ee7e0b2c01554a5224e8b !

PTAL :)

funky-eyes (approved):
LGTM

PR #7157, opened 2025-02-11
feature: migrate the console to the naming server

funky-eyes:


YongGoose:
@funky-eyes 

If you're not planning to merge it right away, would it be okay if I take a look at the code and leave some comments?
Personally, I’m very interested in Spring. 🙂

funky-eyes:
> @funky-eyes
> 
> If you're not planning to merge it right away, would it be okay if I take a look at the code and leave some comments? Personally, I’m very interested in Spring. 🙂

Of course

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
```diff
@@ -113,14 +115,18 @@ public Result<String> changeGroup(@RequestParam String namespace,
...
-
         Result<String> addGroupResult = namingManager.changeGroup(namespace, vGroup, clusterName, unitName);
         if (!addGroupResult.isSuccess()) {
             return addGroupResult;
         }
         return new Result<>("200", "change vGroup " + vGroup + "to cluster " + clusterName + " successfully!");
     }
 
+    @GetMapping("/namespace")
+    public SingleResult<Map<String, NamespaceVO>> namespaces() {
```
YongGoose, line 140: [DTO](https://en.wikipedia.org/wiki/Data_transfer_object) can be a good option.

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
```diff
@@ -0,0 +1,75 @@
...
+		return new RestTemplate(requestFactory);
+	}
+
+    @Bean
+    public AsyncRestTemplate asyncRestTemplate(RestTemplate restTemplate) {
+        HttpComponentsAsyncClientHttpRequestFactory asyncClientHttpRequestFactory =
+            new HttpComponentsAsyncClientHttpRequestFactory();
+        asyncClientHttpRequestFactory.setConnectionRequestTimeout(5000); // Connection request timeout in milliseconds
+        asyncClientHttpRequestFactory.setConnectTimeout(5000); // Connection timeout in milliseconds
+        asyncClientHttpRequestFactory.setReadTimeout(5000); // Read timeout in milliseconds
```
YongGoose: If the same time needs to be used, defining it as a `constant` could be a good option.

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
```diff
@@ -0,0 +1,75 @@
...
+        return new AsyncRestTemplate(asyncClientHttpRequestFactory, restTemplate);
+    }
+
+    @Bean
+    public FilterRegistrationBean<Filter> consoleRemotingFilter(NamingManager namingManager, AsyncRestTemplate asyncRestTemplate) {
+        ConsoleRemotingFilter consoleRemotingFilter = new ConsoleRemotingFilter(namingManager, asyncRestTemplate);
+        FilterRegistrationBean<Filter> registration = new FilterRegistrationBean<>();
+        registration.setFilter(consoleRemotingFilter);
+        registration.addUrlPatterns("/api/v1/console/*");
```
YongGoose: Since `addUrlPatterns` accepts a variable-length array, it might be a good idea to define an `array` if the number of APIs increases to two or more.

- https://docs.spring.io/spring-boot/api/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBean.html#addUrlPatterns(java.lang.String...)

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
```diff
@@ -96,10 +100,8 @@ public MetaResponse discovery(@RequestParam String vGroup, @RequestParam String
     }
 
     @PostMapping("/addGroup")
-    public Result<String> addGroup(@RequestParam String namespace,
-                                      @RequestParam String clusterName,
-                                      @RequestParam String unitName,
-                                      @RequestParam String vGroup) {
+    public Result<String> addGroup(@RequestParam String namespace, @RequestParam String clusterName, String unitName,
+        @RequestParam String vGroup) {
```
YongGoose, line 118:
```suggestion
    public Result<String> addGroup(@RequestParam String namespace, @RequestParam String clusterName, @RequestParam String unitName,
        @RequestParam String vGroup) {
```

If `unitName` is not passed as a parameter, you can ignore this comment!

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
```diff
@@ -113,14 +115,18 @@ public Result<String> changeGroup(@RequestParam String namespace,
                                       @RequestParam String clusterName,
                                       @RequestParam String unitName,
                                       @RequestParam String vGroup) {
-
         Result<String> addGroupResult = namingManager.changeGroup(namespace, vGroup, clusterName, unitName);
         if (!addGroupResult.isSuccess()) {
             return addGroupResult;
         }
         return new Result<>("200", "change vGroup " + vGroup + "to cluster " + clusterName + " successfully!");
```
YongGoose, line 136: You can use the constants from `Result`.

- https://github.com/apache/incubator-seata/blob/205581771b46a1f5b72b47b008bc4f0663018141/common/src/main/java/org/apache/seata/common/result/Result.java#L27

Or how about creating a `success` API like `SingleResult`?

namingserver/src/main/java/org/apache/seata/namingserver/filter/ConsoleRemotingFilter.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.concurrent.ListenableFutureCallback;
+import org.springframework.web.client.AsyncRestTemplate;
+
+public class ConsoleRemotingFilter implements Filter {
+
+    NamingManager namingManager;
+
+    AsyncRestTemplate asyncRestTemplate;
```
YongGoose: If the variable does not change, adding the `final` keyword can be a good approach.

YongGoose (approved):
Since I’m not familiar with the overall code of the `namingserver` module, my comment might not be helpful.

I’d appreciate it if you take it lightly.

Great work! 👍

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
funky-eyes: > Since `addUrlPatterns` accepts a variable-length array, it might be a good idea to define an `array` if the number of APIs increases to two or more.
> 
> * https://docs.spring.io/spring-boot/api/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBean.html#addUrlPatterns(java.lang.String...)

I think using regular expressions might be better, so I will modify this part of the code

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
funky-eyes: ok

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
funky-eyes: > [DTO](https://en.wikipedia.org/wiki/Data_transfer_object) can be a good option.

I think DTO is a transfer object and VO is a presentation object, so I used VO. Maybe my understanding is incorrect, please point it out.

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
funky-eyes: The unitname is only used for the group in Seata's Raft mode. It is not required for other storage modes, so it is not necessary to provide it.

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
funky-eyes: Yes, but it is not within the scope of responsibility of the pr.

namingserver/src/main/java/org/apache/seata/namingserver/filter/ConsoleRemotingFilter.java
funky-eyes: ok

YongGoose:
@funky-eyes 

The `Result` object seemed to not effectively utilize the `type` received from the `generic`.
What do you think about this?

If you agree with my opinion, I will create an issue after resolving my current `PR`.

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
YongGoose: Thank you for the detailed explanation! :)
I think it’s okay to mark it as resolved.

funky-eyes:
> @funky-eyes
> 
> The `Result` object seemed to not effectively utilize the `type` received from the `generic`. What do you think about this?
> 
> If you agree with my opinion, I will create an issue after resolving my current `PR`.

I agree with you

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
YongGoose: My intention was not to replace `NamespaceVO` with a `DTO` but rather to replace `Map<String, NamespaceVO>` with a `DTO`.

As you mentioned, `DTO` is meant for transferring data between layers, so I thought it would be appropriate.

I think this website could be helpful for our discussion.
- https://stackoverflow.com/questions/1612334/difference-between-dto-vo-pojo-javabeans

funky-eyes: Since the controller layer does not perform any actions and the `VO` is the data that needs to be displayed on the front end, I believe it is sufficient to use `Map<String, NamespaceVO>` in the response. If we return a `DTO` to the controller side, the controller would need to convert it to `VO`, which I think is redundant.

YongGoose: > If we return a `DTO` to the controller side, the controller would need to convert it to `VO`, which I think is redundant.

Thank you for your opinion.

If the controller layer simply returns what the service layer provides, the DTO might seem unnecessary. However, when transmitting complex objects, I believe using a DTO is appropriate.

Discussions about DTOs and VOs never seem to have a definitive answer. 😅

This time, I think it would be better to follow your suggestion and use a `VO` to avoid redundancy.

namingserver/src/main/java/org/apache/seata/namingserver/controller/NamingController.java
YongGoose: > Yes, but it is not within the scope of responsibility of the pr.

I’ll take care of this part.

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
funky-eyes: done

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
funky-eyes: done

namingserver/src/main/java/org/apache/seata/namingserver/filter/ConsoleRemotingFilter.java
funky-eyes: done

console/src/main/resources/static/console-fe/src/pages/GlobalLockInfo/GlobalLockInfo.tsx
```diff
@@ -55,28 +71,67 @@ class GlobalLockInfo extends React.Component<GlobalProps, GlobalLockInfoState> {
...
 
   componentDidMount = () => {
     // @ts-ignore
     const { query } = this.props.history.location;
     if (query !== undefined) {
-      const { xid } = query;
-      if (xid !== undefined) {
+      const { xid,vgroup ,namespace,cluster} = query;
+      if (xid !== undefined && vgroup !== undefined) {
+        console.log('query:', query);
```
xjlgod: Is this line used to debug?

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
```diff
@@ -0,0 +1,79 @@
...
+        HttpComponentsAsyncClientHttpRequestFactory asyncClientHttpRequestFactory =
+            new HttpComponentsAsyncClientHttpRequestFactory();
+        asyncClientHttpRequestFactory.setConnectionRequestTimeout(DEFAULT_REQUEST_TIMEOUT); // Connection request timeout in milliseconds
+        asyncClientHttpRequestFactory.setConnectTimeout(DEFAULT_REQUEST_TIMEOUT); // Connection timeout in milliseconds
+        asyncClientHttpRequestFactory.setReadTimeout(DEFAULT_REQUEST_TIMEOUT); // Read timeout in milliseconds
+        return new AsyncRestTemplate(asyncClientHttpRequestFactory, restTemplate);
+    }
+
+    @Bean
+    public FilterRegistrationBean<Filter> consoleRemotingFilter(NamingManager namingManager,
```
xjlgod, line 69: what is the use of consoleRemotingFilter

console/src/main/resources/static/console-fe/src/pages/GlobalLockInfo/GlobalLockInfo.tsx
funky-eyes: Yes, I should delete it

namingserver/src/main/java/org/apache/seata/namingserver/config/WebConfig.java
funky-eyes: It is used to route console requests to the corresponding seata server

console/src/main/resources/static/console-fe/src/pages/GlobalLockInfo/GlobalLockInfo.tsx
funky-eyes:
```suggestion
```

namingserver/src/main/java/org/apache/seata/namingserver/contants/NamingConstant.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.seata.namingserver.contants;
+
+public interface NamingConstant {
+
+    String CONSOLE_PATTERN = "^/api/.*/console/.*";
+
+    int DEFAULT_REQUEST_TIMEOUT = 5000;
```
xjlgod, line 23: Should this config can be set by user?

funky-eyes: > Should this config can be set by user?

Theoretically, it's not necessary. I believe 5 seconds is already quite long, because the naming server and seata-server are on the same internal network, and since it's a paginated query, there isn't much data.

xjlgod:
LGTM

xjlgod (approved):
LGTM

PR #7162, opened 2025-02-13
optimize: Throwing IOException in sendSyncRequest 

### LINKED ISSUE #7161: RuntimeException in sendSyncRequest causing  missing logs
## Why you need it?

Image

Image

Image

When a RuntimeException is thrown, the outer logic (AbstractCore) fails to catch it because it only catches (IOException | TimeoutException e). As a result, a BranchTransactionException is not thrown. Since AbstractTCInboundHandler only catches (StoreException e), it also fails to catch the RuntimeException, and ultimately, the critical log is not printed.

---
当RuntimeException被抛出之后，外层逻辑（AbstractCore）无法catch，因为它只catch (IOException | TimeoutException e)，所以也没有抛出BranchTransactionException，直到AbstractTCInboundHandler也catch不到异常，它只catch (StoreException e)，最后没有打印出关键日志。

## How it could be?

Bughue:

When a RuntimeException is thrown, the outer logic (AbstractCore) fails to catch it because it only catches (IOException | TimeoutException e). As a result, a BranchTransactionException is not thrown. Since AbstractTCInboundHandler only catches (StoreException e), it also fails to catch the RuntimeException, and ultimately, the critical log is not printed.

fixes https://github.com/apache/incubator-seata/issues/7161


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingServer.java
```diff
@@ -68,28 +69,28 @@ public AbstractNettyRemotingServer(ThreadPoolExecutor messageExecutor, NettyServ
 
     @Override
     public Object sendSyncRequest(String resourceId, String clientId, Object msg, boolean tryOtherApp)
-        throws TimeoutException {
+            throws TimeoutException, IOException {
         Channel channel = ChannelManager.getChannel(resourceId, clientId, tryOtherApp);
         if (channel == null) {
-            throw new RuntimeException("rm client is not connected. dbkey:" + resourceId + ",clientId:" + clientId);
+            throw new IOException("rm client is not connected. dbkey:" + resourceId + ",clientId:" + clientId);
```
funky-eyes, line 75: Why is it considered an IO exception when a channel does not exist, rather than a runtime exception?

Bughue: 我认为channel失效也属于IO相关异常。而且调用send的地方也只捕捉IO和超时的异常，直接抛出runtime exception会无法转化到相应的异常
I think channel failure is also an IO related exception. Also the send call only catches IO and timeout exceptions, throwing a runtime exception directly will not translate to the appropriate exception.
image

funky-eyes: Then I believe it would be more appropriate to wrap it as a SeataException.

slievrly (approved):
LGTM

PR #7164, opened 2025-02-13
feature:Transfer global timeout and branch transaction begin time

xiaoxiangyeyu0:

XA模式用到的全局超时时间没有使用单独配置，需要传递全局超时时间和分支事务开始时间
The global timeout time used in the XA Mode is not configured separately, and the global timeout time and branch transaction start time need to be passed.

### DIFF
--- a/rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
+++ b/rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
@@ -27,6 +27,7 @@ import org.apache.seata.common.DefaultValues;
 import org.apache.seata.common.lock.ResourceLock;
 import org.apache.seata.common.util.StringUtils;
 import org.apache.seata.config.ConfigurationFactory;
+import org.apache.seata.core.context.RootContext;
 import org.apache.seata.core.exception.TransactionException;
 import org.apache.seata.core.model.BranchStatus;
 import org.apache.seata.core.model.BranchType;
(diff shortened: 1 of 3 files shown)


tm/src/main/java/org/apache/seata/tm/api/DefaultGlobalTransaction.java
```diff
@@ -113,6 +113,8 @@ public void begin(int timeout, String name) throws TransactionException {
         xid = transactionManager.begin(null, null, name, timeout);
         status = GlobalStatus.Begin;
         RootContext.bind(xid);
+        RootContext.setTimeout(timeout);
+        RootContext.setBranchBeignTime(createTime);
```
funky-eyes, line 117: integration模块中需要传递两个值
Two values need to be passed in the integration module

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
```diff
@@ -93,6 +94,15 @@ public void init() {
             if (!currentAutoCommitStatus) {
                 throw new IllegalStateException("Connection[autocommit=false] as default is NOT supported");
             }
+            Integer transactionTimeout = RootContext.getTimeout();
+            if (transactionTimeout == null) {
+                transactionTimeout = DefaultValues.DEFAULT_GLOBAL_TRANSACTION_TIMEOUT;
+            }
+            timeout = Math.max(BRANCH_EXECUTION_TIMEOUT, transactionTimeout);
+            branchBeginTime = RootContext.getBranchBeignTime();
```
slievrly, line 102: I don't see this branchBeginTime value being passed from the downstream service, and if it's not passed, it's no different from the previous function. Or what you're trying to say is timeout is the timeout of a branch transaction not the timeout of a global transaction?

tm/src/main/java/org/apache/seata/tm/api/DefaultGlobalTransaction.java
slievrly, line 117: If I set this value here I think it is ambiguous, the time here should be the start time of the global transaction and not the begin time of the branch transaction.

PR #7170, opened 2025-02-19
optimize: optimize seata client I/O processing by adjusting thread count

### LINKED ISSUE #7168: The client's NioEventLoopGroup thread is only 1 by default

### Ⅰ. Issue Description

Image

Image

The io reading and coding/decoding work of the seata client is in this thread, the default thread of 1 will affect the efficiency of its processing, we need to adjust it to the number of threads of cpu*2, you can change the default value to `WorkThreadMode.Default.getValue()`

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- Database version:
- OS(e.g. `uname -a`):
- Others:

YongGoose:

fixes #7168


funky-eyes (approved):
LGTM

funky-eyes:
https://github.com/apache/incubator-seata/blob/2.x/script/client/conf/file.conf#L44
https://github.com/apache/incubator-seata/blob/2.x/script/client/spring/application.properties#L74
https://github.com/apache/incubator-seata/blob/2.x/script/client/spring/application.yml#L85
https://github.com/apache/incubator-seata/blob/2.x/script/config-center/config.txt#L35
Since the corresponding IO thread count for the client in the above files is 1, I believe these configurations should be changed to -1. Then, in the code, handle the value of -1 by converting it to the number of CPU cores * 2.

core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientConfig.java
```diff
@@ -25,7 +25,6 @@
 import static org.apache.seata.common.DefaultValues.DEFAULT_RPC_RM_REQUEST_TIMEOUT;
 import static org.apache.seata.common.DefaultValues.DEFAULT_RPC_TM_REQUEST_TIMEOUT;
 import static org.apache.seata.common.DefaultValues.DEFAULT_SELECTOR_THREAD_PREFIX;
-import static org.apache.seata.common.DefaultValues.DEFAULT_SELECTOR_THREAD_SIZE;
```
funky-eyes, line 28: The value of DEFAULT_SELECTOR_THREAD_SIZE should be changed to -1

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7170/commits/f5d38a93014946cf08a563a4b37c44b9e8687b5c !

YongGoose:
> Then, in the code, handle the value of -1 by converting it to the number of CPU cores * 2.

What do you think about changing the thread count to CPU cores * 2 when the value is negative?

core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientConfig.java
funky-eyes: > Done in [f5d38a9](https://github.com/apache/incubator-seata/commit/f5d38a93014946cf08a563a4b37c44b9e8687b5c) !

In NettyClientBootstrap, it is necessary to handle the case where selectorThreadSizeThreadSize=-1.

YongGoose: > > Done in [f5d38a9](https://github.com/apache/incubator-seata/commit/f5d38a93014946cf08a563a4b37c44b9e8687b5c) !
> 
> In NettyClientBootstrap, it is necessary to handle the case where selectorThreadSizeThreadSize=-1.

- https://github.com/apache/incubator-seata/pull/7170#issuecomment-2667623948

What do you think?

funky-eyes:
> What do you think about changing the thread count to CPU cores * 2 when the value is negative?

If it is a negative number, directly assign the value of `WorkThreadMode.Default.getValue()` to it

core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientConfig.java
```diff
@@ -346,11 +345,13 @@ public static String getSocketAddressStartChar() {
...
      * Gets client selector thread size.
+     * If the configured thread size is less than or equal to 0, it returns the default value.
      *
-     * @return the client selector thread size
+     * @return the client selector thread size, or the default value if the configured size is invalid.
      */
     public int getClientSelectorThreadSize() {
-        return CONFIG.getInt(ConfigurationKeys.CLIENT_SELECTOR_THREAD_SIZE, DEFAULT_SELECTOR_THREAD_SIZE);
+        int threadSize = CONFIG.getInt(ConfigurationKeys.CLIENT_SELECTOR_THREAD_SIZE, WorkThreadMode.Default.getValue());
+        return threadSize > 0 ? threadSize : WorkThreadMode.Default.getValue();
```
YongGoose, line 354: @funky-eyes 

Handling this in `NettyClientBootstrap` is a good idea, but I believe it is more appropriate to process it here.
Because if we handle this in `NettyClientBootstrap`, we would need to process` -1` in every piece of code that retrieves the thread count using this method.

funky-eyes: You are right, handling it this way is better, good job.

YongGoose:
@funky-eyes 

I have resolved the failing test related to `getClientSelectorThreadSize`.
However, the `testRemoveConfig` test in `ZkConfigurationTest` is still failing, but only when running on `Java 8`.
Could you help me understand why this failure occurs specifically in `Java 8`?

funky-eyes:
> @funky-eyes
> 
> I have resolved the failing test related to `getClientSelectorThreadSize`. However, the `testRemoveConfig` test in `ZkConfigurationTest` is still failing, but only when running on `Java 8`. Could you help me understand why this failure occurs specifically in `Java 8`?

This is not caused by your PR; this issue is an occasional occurrence, and the exact cause is currently unknown. I have restarted the CI workflow for now.

PR #7174, opened 2025-02-20
feature:[loom] replace the usages of synchronized with ReentrantLock at saga module

### LINKED ISSUE #7173: feature: [loom] replace the usages of synchronized with ReentrantLock at sage module
## Why you need it?

## How it could be?

lightClouds917:

replace the usages of synchronized with ReentrantLock  at sage module

To use virtual threads in JDK 21, we need to remove the usage of synchronized to prevent the issue where "a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier."

There are two scenarios in which a virtual thread cannot be unmounted during blocking operations because it is pinned to its carrier:

When it executes code inside a synchronized block or method, or When it executes a native method or a [foreign function](https://openjdk.java.net/jeps/424).
Pinning does not make an application incorrect, but it might hinder its scalability. If a virtual thread performs a blocking operation such as I/O or BlockingQueue.take() while it is pinned, then its carrier and the underlying OS thread are blocked for the duration of the operation. Frequent pinning for long durations can harm the scalability of an application by capturing carriers.

The scheduler does not compensate for pinning by expanding its parallelism. Instead, avoid frequent and long-lived pinning by revising synchronized blocks or methods that run frequently and guard potentially long I/O operations to use [java.util.concurrent.locks.ReentrantLock](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html) instead. There is no need to replace synchronized blocks and methods that are used infrequently (e.g., only performed at startup) or that guard in-memory operations. As always, strive to keep locking policies simple and clear.

You can refer to the following documents:

https://openjdk.org/jeps/444
https://openjdk.org/projects/loom/

fixed #7173


compatible/src/main/java/io/seata/saga/proctrl/ProcessContext.java
```diff
@@ -97,4 +98,10 @@ public interface ProcessContext extends org.apache.seata.saga.proctrl.ProcessCon
      * @return the get instruction
      */
     <T extends Instruction> T getInstruction(Class<T> clazz);
+
+    /**
+     * Gets get lock.
+     * @return the lock of the current process context
+     */
+    ResourceLock getLock();
```
slievrly: The getLock method doesn't really fit a universal business process or template setup, so it's not quite suitable to include it in the interface.

slievrly:
Long time no progress, close it.

PR #7179, opened 2025-02-23
optimize: Use shared `EventLoop` for TM and RM clients to reduce thread overhead and improve performance

### LINKED ISSUE #6774: optimize seata-client TM&RM netty eventloop

YongGoose:
…ad overhead

fixes #6774


core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientBootstrap.java
```diff
@@ -56,18 +56,20 @@
...
+
+
     private final NettyClientConfig nettyClientConfig;
     private final Bootstrap bootstrap = new Bootstrap();
-    private final EventLoopGroup eventLoopGroupWorker;
-    private EventExecutorGroup defaultEventExecutorGroup;
     private final AtomicBoolean initialized = new AtomicBoolean(false);
-    private static final String THREAD_PREFIX_SPLIT_CHAR = "_";
     private final NettyPoolKey.TransactionRole transactionRole;
+    private final EventLoopGroup sharedEventLoopGroup;
```
funky-eyes: I think it should still be named eventLoopGroupWorker, because it is possible that this EventLoopGroup is not shared.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7179/commits/982f2a3f150fef43f1b735f59883293c973f96a4 !

YongGoose: I would appreciate it if you could also review the comments below. 🙂
- https://github.com/apache/incubator-seata/issues/6774#issuecomment-2676706690

core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientBootstrap.java
```diff
@@ -56,18 +56,20 @@
...
+
+
     private final NettyClientConfig nettyClientConfig;
     private final Bootstrap bootstrap = new Bootstrap();
-    private final EventLoopGroup eventLoopGroupWorker;
-    private EventExecutorGroup defaultEventExecutorGroup;
     private final AtomicBoolean initialized = new AtomicBoolean(false);
-    private static final String THREAD_PREFIX_SPLIT_CHAR = "_";
     private final NettyPoolKey.TransactionRole transactionRole;
+    private final EventLoopGroup eventLoopGroupWorker;
```
funky-eyes, line 71: This is not a static variable, the clientBootstrap held by the two instances of TmNettyRemotingClient and RmNettyRemotingClient are not the same, how does this variable manage to be shared use?

YongGoose: Could this be related to the comment?
- https://github.com/apache/incubator-seata/issues/6774#issuecomment-2676706690

In fact, I also believe that it should be a static variable to use the same value across multiple instances.  
However, if it becomes a static variable, there’s an issue where the previous `eventLoopGroupWorker` gets modified when a new one is created. I wanted to ask about that.

funky-eyes: The previous eventLoopGroupWorker should not have been created yet, so I don't think there should be an action to modify this.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7179/commits/f308be6069c0750ac8de34d905ce53a73681b484 !

PTAL ⚡️

core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientBootstrap.java
```diff
@@ -56,18 +56,20 @@
...
 /**
  * Rpc client.
- *
  */
 public class NettyClientBootstrap implements RemotingBootstrap {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NettyClientBootstrap.class);
+    private static final String THREAD_PREFIX_SPLIT_CHAR = "_";
+
+    private static EventLoopGroup eventLoopGroupWorker;
```
funky-eyes: I think it should be a static variable, sharedEventLoopGroupWorker, and eventLoopGroupWorker is an instance variable. When enableClientSharedEventLoop is true, sharedEventLoopGroupWorker is created and assigned to eventLoopGroupWorker.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7179/commits/a836f57ca0ec5a6105b0158795623a31f8d093f8 !

funky-eyes (approved):
LGTM

Issue #7082, opened 2024-12-28
bugfix: In raft mode, the service domain name is used as xid. The xid is too long and an error occurs in XA mode.


PeppaO:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
这个错误在跑XA测试用例中也出现过，所以手动执行XA start测试，发现就是字符太长导致的。
<img width="790" alt="image" src="https://github.com/user-attachments/assets/385d7ecf-9d7c-4f30-aba2-473f3e53ac3e" />

### Ⅱ. Describe what happened

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

Minimal yet complete reproducer code (or URL to code):

### Ⅵ. Environment:

- JDK version(e.g. `java -version`): 1.8
- Seata client/server version:  2.x
- Database version: mysql 8.0.27
- OS(e.g. `uname -a`):
- Others:

slievrly:
@PeppaO   Are you sure it's not a special character? This length limit is usually 132 characters, although it may vary depending on the version.

PeppaO:
> @PeppaO Are you sure it's not a special character? This length limit is usually 132 characters, although it may vary depending on the version.

error: xa start "seata-server-0.seata-svc-headless.default.svc.cluter.local:8091:8269208701573369857";
correct: xa start "s-0.svc-l.default.svc.cluster.local:8091:1784025255588331521";
I think its maximum length can be 64.

funky-eyes:
我认为应该将xa的xid设定为transactionId和branchid的组合，而不是xid和branchid，理论上transactionid和branchid都是唯一的，更加精简，也更合适作为xa xid。但是需要考虑如何向下兼容。
I believe the XA XID should be set as a combination of the transactionId and branchId, rather than the XID and branchId. In theory, both the transactionId and branchId are unique, making this combination more concise and better suited as the XA XID. However, it is necessary to consider how to maintain backward compatibility.
@slievrly

PR #7181, opened 2025-02-24
feature:  raft implements domain name resolution and selects peerId

PeppaO:

raft不设置SEATA_IP，raft实现域名解析并选择peerId

#7082


funky-eyes (commented):
I think this PR is not a bugfix PR because, according to the changes you listed, the first point is a feature and the second point is an optimization.

funky-eyes (approved):
LGTM

PR #7183, opened 2025-02-25
feature: client discovers raft nodes through the naming server

funky-eyes:

### DIFF
--- a/server/src/main/resources/application.raft.example.yml
+++ b/server/src/main/resources/application.raft.example.yml
@@ -32,10 +32,6 @@ logging:
     kafka-appender:
       bootstrap-servers: 127.0.0.1:9092
       topic: logback_to_logstash
-console:
-  user:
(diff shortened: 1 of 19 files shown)


server/src/main/resources/application.raft.example.yml
```diff
@@ -76,6 +72,15 @@ seata:
     # support: nacos 、 eureka 、 redis 、 zk  、 consul 、 etcd3 、 sofa
     type: file
     preferred-networks: 30.240.*
+    seata:
+      server-addr: 127.0.0.1:8081
+      cluster: default
+      namespace: public
+      heartbeat-period: 5000
+      metadata-max-age-ms: 30000
```
slievrly, line 80: If the type is not configured as seata here, how should the username and password for logging into the console be set?

slievrly (approved):
LGTM

PR #7184, opened 2025-02-25
optimize: using subqueries to reduce the retrieval of invalid content data.

MaoMaoandSnail:


saga/seata-saga-engine-store/src/main/java/org/apache/seata/saga/engine/store/db/StateLangStoreSqls.java
```diff
@@ -62,6 +62,10 @@ public String getQueryStateMachinesByNameAndTenantSql(String dbType) {
         return queryStateMachinesByNameAndTenantSql;
     }
 
+    public String getLastVersionStateMachine(String dbType) {
+        return getQueryStateMachinesByNameAndTenantSql(dbType) + " LIMIT 1";
```
slievrly: The LIMIT syntax is not part of the SQL standard, and there are several databases that do not support the LIMIT syntax.

MaoMaoandSnail: > The LIMIT syntax is not part of the SQL standard, and there are several databases that do not support the LIMIT syntax.

OK, I will optimize it.

saga/seata-saga-engine-store/src/main/java/org/apache/seata/saga/engine/store/db/DbStateLangStore.java
```diff
@@ -50,8 +50,12 @@ public StateMachine getStateMachineById(String stateMachineId) {
 
     @Override
     public StateMachine getLastVersionStateMachine(String stateMachineName, String tenantId) {
-        return selectOne(stateLangStoreSqls.getLastVersionStateMachine(dbType), RESULT_SET_TO_STATE_MACHINE,
-                stateMachineName, tenantId);
+        List<StateMachine> list = selectList(stateLangStoreSqls.getLastVersionStateMachine(dbType),
```
slievrly, line 53: Has this change been fully tested by saga? I remember that tenant_id is an optional field, what happens if the upper layer doesn't pass the tenant_id parameter?

MaoMaoandSnail: > Has this change been fully tested by saga? I remember that tenant_id is an optional field, what happens if the upper layer doesn't pass the tenant_id parameter?

I've performed local testing and executed the Saga test approach. When the tenant_id parameter is not provided, the system will automatically fill it with the defaultTenantId (000001).
image

PR #7223, opened 2025-03-15
optimize: Apply Spotless with Palantir Java Format

YongGoose:
### **🔹 Related Document**
https://lists.apache.org/thread/b4mmng4t0tog6bd58y7y9vnr02jn0q9f

### **🔹 Key Changes:**  
- **Configured Spotless Maven Plugin** to automatically format Java files using **Palantir Java Format**.  
  - https://github.com/palantir/palantir-java-format
- **JDK version-based profile activation:**  
  - JDK **8 to <11** → Uses `palantirJavaFormat.version = 1.1.0`  
  - JDK **11 to <21** → Uses `palantirJavaFormat.version = 2.28.0`  
  - JDK **21+** → Uses `palantirJavaFormat.version = 2.39.0`  
- **Excluded specific directories/files** from formatting:  
  - `script/`, `generated/`, `antlr/mysql/parser/`, `antlr/mysql/antlr/`, etc.  
- **Integrated with Maven’s `process-sources` phase**, ensuring formatting is applied during builds.  
- **Enabled unused import removal, trailing whitespace trimming, and newline enforcement at EOF.**  

### **🔹 How to Verify:**  
1️⃣ Run **Maven build** to ensure Spotless is applied automatically:  
   ```sh
   mvn clean install
   ```  
2️⃣ Manually apply formatting with:  
   ```sh
   mvn spotless:apply
   ```  
3️⃣ Verify excluded directories remain unchanged.  

### **🔹 Screen shots**
image
image

ps. I found that applying Spotless resulted in 1,914 files being modified!!
image


YongGoose:
@funky-eyes @xingfudeshi 

What do you think? 🤔

slievrly:
@YongGoose Can Spotless only work on the change files instead of all files?

YongGoose:
> @YongGoose Can Spotless only work on the change files instead of all files?

According to the link below, it seems possible using a Git pre-commit hook or `ratchetFrom`.
- https://github.com/diffplug/spotless/issues/178

YongGoose:
> Can Spotless only work on the change files instead of all files?

No, Spotless can be applied to all files as well.

YongGoose:
There are several things we need to discuss ❗️

1. Should we use both `Spotless` and `Checkstyle`, or just Spotless?  
2. Files that should be excluded from `Spotless` (currently, the same exclusions as `Checkstyle` have been applied).  
3. When should Spotless be automatically applied (in CI process)?

funky-eyes:
I think this approach is very effective. Although it involves modifying many files, I believe as long as the CI passes, there won't be any significant issues. Even though it may conflict with other PRs, I think this cost is worthwhile.

YongGoose:
> I think this approach is very effective. Although it involves modifying many files, I believe as long as the CI passes, there won't be any significant issues. Even though it may conflict with other PRs, I think this cost is worthwhile.

Thank you for your comment! 

However, after discussing with @slievrly on the mailing lists, we decided to apply spotless only to the modified files, so it looks like we won’t be modifying many files!

> Although there is the drawback that spotless won’t be applied to unmodified files, after weighing the trade-offs with the git history, we determined that it’s better to only modify the changed files.

funky-eyes:
> > I think this approach is very effective. Although it involves modifying many files, I believe as long as the CI passes, there won't be any significant issues. Even though it may conflict with other PRs, I think this cost is worthwhile.
> 
> Thank you for your comment!
> 
> However, after discussing with @slievrly on the mailing lists, we decided to apply spotless only to the modified files, so it looks like we won’t be modifying many files!
> 
> > Although there is the drawback that spotless won’t be applied to unmodified files, after weighing the trade-offs with the git history, we determined that it’s better to only modify the changed files.

It would be great if this can be achieved, but when I run the mvn spotless:apply command locally, how can I ensure that only the files I have modified are formatted?

YongGoose:
> It would be great if this can be achieved, but when I run the mvn spotless:apply command locally, how can I ensure that only the files I have modified are formatted?

I believe it can be done by using the `git diff` feature with `ratchet`. 
I haven’t actually implemented it yet, but it does seem feasible.

- https://github.com/diffplug/spotless/tree/main/plugin-maven#how-do-i-preview-what-mvn-spotlessapply-will-do

funky-eyes:
> > It would be great if this can be achieved, but when I run the mvn spotless:apply command locally, how can I ensure that only the files I have modified are formatted?
> 
> I believe it can be done by using the `git diff` feature with `pre-commit` hook. I haven’t actually implemented it yet, but it does seem feasible.
> 
> * https://github.com/diffplug/spotless/tree/main/plugin-maven#how-do-i-preview-what-mvn-spotlessapply-will-do

Ok, I hope this solution is feasible. If it's not, I don't mind modifying all the files; I don't think it will have too much of an impact.

pom.xml
```diff
@@ -289,6 +289,78 @@
...
+                        <configuration>
+                            <java>
+                                <excludes>
+                                    <exclude>**/script/**</exclude>
+                                    <exclude>**/generated/**</exclude>
+                                    <exclude>**/antlr/mysql/parser/*.*</exclude>
+                                    <exclude>**/antlr/mysql/antlr/*.*</exclude>
+                                    <exclude>**/antlr/mysql/stream/ANTLRNoCaseStringStream.java</exclude>
+                                </excludes>
+                                <ratchetFrom>HEAD</ratchetFrom>
```
YongGoose, line 343: @slievrly @funky-eyes @xingfudeshi 

I configured `<ratchetFrom>HEAD</ratchetFrom>` so that only the modified files are formatted by comparing them with the currently checked-out commit (HEAD).

funky-eyes: > @slievrly @funky-eyes @xingfudeshi
> 
> I configured `<ratchetFrom>HEAD</ratchetFrom>` so that only the modified files are formatted by comparing them with the currently checked-out commit (HEAD).

Good job! I will pull this PR to my local machine for related testing

YongGoose: @funky-eyes 

Currently, our baseline is set to `HEAD`, which means that only files changed since the last commit are considered. This creates a potential issue: if code with incorrect formatting gets committed, `checkstyle` will catch it, but `Spotless` won’t apply its fixes, and we lose the intended benefit.

To address this, we tried switching the baseline from HEAD to `origin/2.x`. However, the branch wasn’t recognized. 🥹

What do you think?

xingfudeshi: > @funky-eyes
> 
> Currently, our baseline is set to `HEAD`, which means that only files changed since the last commit are considered. This creates a potential issue: if code with incorrect formatting gets committed, `checkstyle` will catch it, but `Spotless` won’t apply its fixes, and we lose the intended benefit.
> 
> To address this, we tried switching the baseline from HEAD to `origin/2.x`. However, the branch wasn’t recognized. 🥹
> 
> What do you think?

Hi, I pulled your code and followed the steps. Seems to work!
```
1. git checkout feature/apply-spotless
2. Change the value of ratchetFrom from HEAD to feature/apply-spotless
3. Add some spaces in the method of Java class, like this
```
image 

```
4. mvn spotless:check
5. Then I got the outputs like this
```
image

funky-eyes: > @funky-eyes
> 
> Currently, our baseline is set to `HEAD`, which means that only files changed since the last commit are considered. This creates a potential issue: if code with incorrect formatting gets committed, `checkstyle` will catch it, but `Spotless` won’t apply its fixes, and we lose the intended benefit.
> 
> To address this, we tried switching the baseline from HEAD to `origin/2.x`. However, the branch wasn’t recognized. 🥹
> 
> What do you think?

I think it can be done this way

funky-eyes: Hi @YongGoose ,
image
I noticed that the constructor is formatted like this after formatting. I still prefer the format shown in the image below. Is there a way to adjust it?
image

YongGoose: It seems challenging to apply that format solely to the constructor. 

However, I believe we can set up some configurations based on commas.

funky-eyes: Sure, I don't think this issue is a big deal. I believe we can merge this PR first and then make additional adjustments later. Also, when the CI detects that the code style doesn't meet expectations, can it prompt the user to run mvn spotless:apply locally? I recall that in the Dubbo community's CI workflow, such a prompt is provided when code style checks fail.

YongGoose: I’ll take a look at Dubbo’s code to get a better understanding. It seems feasible.

xingfudeshi:
> > > It would be great if this can be achieved, but when I run the mvn spotless:apply command locally, how can I ensure that only the files I have modified are formatted?
> 
> > 
> 
> > I believe it can be done by using the `git diff` feature with `pre-commit` hook. I haven’t actually implemented it yet, but it does seem feasible.
> 
> > 
> 
> > * https://github.com/diffplug/spotless/tree/main/plugin-maven#how-do-i-preview-what-mvn-spotlessapply-will-do
> 
> 
> 
> Ok, I hope this solution is feasible. If it's not, I don't mind modifying all the files; I don't think it will have too much of an impact.

I agree with your point .

pom.xml
YongGoose: @funky-eyes 

After reviewing the suggested approach, I found that in Dubbo, the GitHub Actions workflow checks the code style and prompts the user to apply Spotless if any issues are detected.
```
run: |
  echo "## :negative_squared_cross_mark: Formatting issues found!" >> $GITHUB_STEP_SUMMARY
  echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
  cat mvn.log | grep "ERROR" | sed 's/Check if code needs formatting    Check if code aligns with code style   [0-9A-Z:.-]\+//' | sed 's/\[ERROR] //' | head -n -11 >> $GITHUB_STEP_SUMMARY
  echo "\`\`\`" >> $GITHUB_STEP_SUMMARY
  echo "Please run \`mvn spotless:apply\` to fix the formatting issues." >> $GITHUB_STEP_SUMMARY
```

However, this approach would not work in our current implementation.  
The reason is that for the workflow to run, the user must first commit and push their changes, triggering the workflow.  
But in our case, Spotless is applied only to the changes from the `latest commit` (`HEAD`).  
- https://javadoc.io/doc/org.eclipse.jgit/org.eclipse.jgit/5.6.1.202002131546-r/org/eclipse/jgit/lib/Repository.html#resolve-java.lang.String-

As a result, even if the workflow instructs the user to run Spotless, the problematic code has already been committed, meaning it falls outside Spotless's scope.  

To resolve this issue, we either need to remove the scope limitation of Spotless or find an alternative approach.

https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#how-can-i-enforce-formatting-gradually-aka-ratchet

funky-eyes: In other words, unless we format and commit all the code at once, we will not be able to prompt the user to run `mvn spotless:apply`. What if we change the `HEAD` to `2.x`?

YongGoose: > What if we change the `HEAD` to `2.x`?

I think this approach is a good one.  

However, when I tried running it locally, the behavior was different from what I expected, so I'm reconsidering it.  
To explain in more detail, switching to `origin/2.x` should apply changes only to the files that have been modified from `2.x`.  

Even though there were no changes in my local branch, some recently committed files in `2.x` were also modified.
image

YongGoose: Still, I think setting it to `2.x` is a better approach than setting it to `HEAD`.

YongGoose: In this commit https://github.com/apache/incubator-seata/pull/7223/commits/4f07580b4988f7012b7d4c455d3dfacc51dabfb1 i changed ratchet from to origin/2.x

Then below error occured.
```
Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.44.3:apply (default) on project seata-parent: Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.44.3:apply failed: No such reference 'origin/2.x' -> [Help 1]
```

funky-eyes: I noticed that after setting it to 2.x, the number of changed files is quite large, unless we are willing to accept changes to all files. Otherwise, I think we can initially set it to HEAD, merge this PR, and then create an issue to discuss the remaining problems.
@slievrly What do you think?

pom.xml
```diff
@@ -289,6 +289,78 @@
...
+                                    <version>${palantirJavaFormat.version}</version>
+                                </palantirJavaFormat>
+                                <removeUnusedImports/>
+                                <trimTrailingWhitespace/>
+                                <endWithNewline/>
+                            </java>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <goals>
```
slievrly, line 354: How it works in github action workflow, I didn't see this profile id included in workflow.In addition to checking if the format is as expected, is it possible to automatically format changes in the workflow and commit them to PR?

YongGoose: This doesn't work in GitHub Actions.

Automatically formatting and committing changes seems more suitable for a `git hook`, which would require users to modify their local `.git` folder.

For now, in this PR, I’ve applied `Spotless`. 
I’m planning to explore additional improvements in a separate PR.

slievrly: ok.

slievrly (approved):
LGTM

PR #7225, opened 2025-03-16
test: added unit tests for NacosRegistryServiceImple

mehedikhan72:
Unit tests added for NacosRegistryServiceImplementation


discovery/seata-discovery-nacos/src/test/java/org/apache/seata/discovery/registry/nacos/NacosRegistryServiceImplTest.java
```diff
@@ -40,5 +46,34 @@ public void testGetConfigProperties() throws Exception {
         Assertions.assertThat(properties.getProperty("contextPath")).isEqualTo("/bar");
     }
 
+    @Test
+    public void testRegister() throws Exception {
+        RegistryService registryService = mock(NacosRegistryServiceImpl.class);
+        InetSocketAddress inetSocketAddress = new InetSocketAddress("127.0.0.1", 8091);
+        registryService.register(inetSocketAddress);
+        verify(registryService).register(inetSocketAddress);
```
slievrly, line 88: Aside from the verify method being called, can we test more fine-grained expected results?

mehedikhan72: @slievrly I'm working on it.

mehedikhan72:
Hi @slievrly . Can you take a look? Since these are mock tests, how else would we improve these? The actual functionalities are tightly coupled(Depends on the discovery services) which makes me think these are not good cases for mock tests, since we are mocking the core functionalities. 

Please let me know if I'm missing any key points, thanks.

slievrly:
> Hi @slievrly . Can you take a look? Since these are mock tests, how else would we improve these? The actual functionalities are tightly coupled(Depends on the discovery services) which makes me think these are not good cases for mock tests, since we are mocking the core functionalities.
> 
> Please let me know if I'm missing any key points, thanks.

Some third-party configuration registries themselves provide mock server capabilities. If they don't, then we'll consider launching a registry container in the github action pipeline to do real testing. If that's not possible, we'll consider mock test to simulate the registry's behavior to see if Seata behaves as expected.

funky-eyes:
https://github.com/apache/incubator-seata/blob/2.x/.github/workflows/build.yml#L38
I don't think mocking is necessary because the Nacos container is already running in the CI environment on Linux. Therefore, we can directly perform the related tests, as long as we add a condition to run only in the Linux environment

slievrly:
Long time no progress, close it.

PR #7234, opened 2025-03-19
optimize: discover the raft leader node from the naming server

funky-eyes:

When a raft re-election occurs, there may be two leaders on the naming server side: the old leader that unexpectedly went down (since the heartbeat has not yet expired, it has not been removed) and the newly elected leader. It is necessary to filter out the latest leader based on the term to ensure the client correctly connects to the leader node.
当raft发生重选举时，namingserver侧可能存在两个leader，一个是意外宕机的老leader(由于心跳还未到期，所以还没被摘除)，一个是新选举的leader，需要根据term去筛选出最新leader，以便client正确连接到leader节点

### DIFF
--- a/discovery/seata-discovery-namingserver/src/test/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImplTest.java
+++ b/discovery/seata-discovery-namingserver/src/test/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImplTest.java
@@ -60,6 +60,7 @@ class NamingserverRegistryServiceImplTest {
 
     @BeforeAll
     public static void beforeClass() throws Exception {
+/*
         System.setProperty("registry.seata.namespace", "dev");
         System.setProperty("registry.seata.cluster", "cluster1");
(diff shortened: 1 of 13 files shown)


discovery/seata-discovery-namingserver/src/test/java/org/apache/seata/discovery/registry/namingserver/NamingserverRegistryServiceImplTest.java
```diff
@@ -157,6 +203,7 @@ public void testRegister2() throws Exception {
 
 
     @Test
+    @Disabled
```
xjlgod, line 206: why there are some many unit test disabled?

funky-eyes: Because these test cases need to be coordinated with the namingserver node, they can only be used for local testing. The test cases in this test class were all initially disabled. I have modified and enabled some of them to run in GitHub CI.

xjlgod (approved):
LGTM

PR #7259, opened 2025-03-30
optimize: seata-server log configuration

YoWuwuuuw:

Optimisation Points:
If you need to collect seata-server logs to Kafka and logstash, you need to modify the logback-spring.xml configuration file (uncomment the corresponding appender), and you need to revise the specific configuration of Kafka-appender by revising kafka-appender.xml and application.yml.
It's a hassle.

Optimisation Scheme: Establish SpringProperty mapping
The configuration of open Kafka, logstash and metric's appender will be transferred to application.yml.
Transfer the specific configuration of Kafka to application.yml.

About dynamic updates to configuration changes
in logback-spring.xml, 
`configuration scan="true" scanPeriod="60 seconds" debug="false"`
The configuration is scaned every minute, so if you introduce a configuration center to change the yml configuration, it is visible to logback originally.

#4648


slievrly (commented):
It looks like these configuration files need to be modified as well.
https://github.com/apache/incubator-seata/blob/2.x/server/src/main/resources/application.example.yml
https://github.com/apache/incubator-seata/blob/2.x/server/src/main/resources/application.raft.example.yml

xingfudeshi (approved):
LGTM

slievrly (approved):
LGTM

Issue #7158, opened 2025-02-11
Refactor: Refactoring the `Result` Class


YongGoose:
## Why you need it?
Please check the comments on the PR.
- https://github.com/apache/incubator-seata/pull/7157#issuecomment-2650089338

## How it could be?
I want to use `type` effectively in the `Result` object, just like in `SimpleResult`.

One idea I have is to move APIs like `success` from `SimpleResult` to `Result`.

YongGoose:
@funky-eyes 

I’d like to proceed once both tasks are completed.
Please assign me to this issue. 

Thank you! ☺️

- https://github.com/apache/incubator-seata/issues/6774
- https://github.com/apache/incubator-seata/pull/7133

funky-eyes:
Well done, looking forward to your PR.

YongGoose:
@xingfudeshi 

If you're interested, please review PR I created 🙂
- https://github.com/apache/incubator-seata/pull/7176

xingfudeshi:
> @xingfudeshi 
> 
> If you're interested, please review PR I created 🙂
> - https://github.com/apache/incubator-seata/pull/7176

Sure!

funky-eyes:
I want to know which modules are currently involved in refactoring and the classes involved, as we need to understand the scope of this issue.

YongGoose:
> I want to know which modules are currently involved in refactoring and the classes involved, as we need to understand the scope of this issue.

I refactored the `Result` class along with its subclasses.  
Consequently, the `common`, `console`, `namingserver`, and `server` modules, which depend on these classes, were also part of the refactoring.

funky-eyes:
OK, I believe the impact of this change is within a controllable range. You can complete this change through multiple PRs, either by module or by dependency.

YongGoose:
> OK, I believe the impact of this change is within a controllable range. You can complete this change through multiple PRs, either by module or by dependency.

Is it correct to address the impact of this PR through multiple PRs?  
If I’ve misunderstood, could you please provide further clarification?

funky-eyes:
> > OK, I believe the impact of this change is within a controllable range. You can complete this change through multiple PRs, either by module or by dependency.
> 
> Is it correct to address the impact of this PR through multiple PRs? If I’ve misunderstood, could you please provide further clarification?

Yes, you can gradually complete the refactoring through multiple PRs.

YongGoose:
> Yes, you can gradually complete the refactoring through multiple PRs.

[...]

PR #7261, opened 2025-03-31
feature: enforce account initialization and disable default credentials

### LINKED ISSUE #6396: Prohibiting Default Username and Password Logins.
After the initial setup of the console, the first login should redirect to a page for initializing the account and password, where data or configuration can be persisted. Subsequently, due to security concerns, the login using default accounts and passwords should be disabled.

### LINKED ISSUE #7307: seata-server 需要配置加密
应安全要求，seata server 的配置文件中不能出现明文密码，application.prpperties中console，nacos注册中心和nacos配置中心的连接信息等需要加密

YongGoose:

fix #6396 
fix #7307

### DIFF
--- a/console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
+++ b/console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
@@ -16,12 +16,11 @@
  */
 package org.apache.seata.console.security;
 
-import java.io.BufferedReader;
-import java.io.Console;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.util.Arrays;
+import java.util.UUID;
 import javax.annotation.PostConstruct;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
@@ -34,47 +33,38 @@ import org.springframework.stereotype.Service;
 @Service
 public class CustomUserDetailsServiceImpl implements UserDetailsService {
 
-    private Console console;
+    private static final Logger LOGGER = LoggerFactory.getLogger(CustomUserDetailsServiceImpl.class);
+
+    @Value("${console.user.username:seata}")
+    private String username;
+
+    @Value("${console.user.password:}")
+    private String password;
+
     private User user;
 
     /**
      * Init.
      */
     @PostConstruct
-    public void init() throws IOException {
-        String envUsername = System.getenv("SEATA_CONSOLE_USERNAME");
-        String envPassword = System.getenv("SEATA_CONSOLE_PASSWORD");
-
-        if (envUsername != null && envPassword != null) {
-            user = new User(envUsername, envPassword);
+    public void init() {
+        if (!password.isEmpty()) {
+            user = new User(username, password);
             return;
         }
 
-        console = System.console();
-        if (console == null) {
-            // In an IDE, 'System.console()' returns 'null', so 'BufferedReader' is used instead.
-            BufferedReader reader = new BufferedReader(new InputStreamReader(System.in));
-            String username = promptInput(reader, "Username: ");
-            String password = promptInput(reader, "Password: ");
-            user = new User(username, password);
-        } else {
-            String username = getUsername();
-            String password = getUserPassword();
-            user = new User(username, password);
-        }
+        password = generateRandomPassword();
+        LOGGER.info(
+                "No password was configured. A random password has been generated for security purposes. You may either:\n"
+                        + "1. Use the auto-generated password: [{}]\n"
+                        + "2. Set a custom password in the configuration.",
+                password);
+
+        user = new User(username, password);
     }
 
-    private String promptInput(BufferedReader reader, String message) throws IOException {
-        System.out.print(message);
-        return reader.readLine();
-    }
-
-    private String getUserPassword() {
-        return Arrays.toString(console.readPassword("Password: "));
-    }
-
-    private String getUsername() {
-        return console.readLine("Username: ");
+    private String generateRandomPassword() {
+        return UUID.randomUUID().toString().replace("-", "").substring(0, 8);
     }
 
     @Override
--- a/console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
+++ b/console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
@@ -22,12 +22,9 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.Arrays;
 import javax.annotation.PostConstruct;
-
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
-import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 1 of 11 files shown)


console/src/main/java/org/apache/seata/console/security/User.java
YongGoose, line 1: If there are no plans to change the user's `username` and `password`, it would be better to change the `User` class to a `record`.

What do you think? @slievrly

YongGoose:
@slievrly 

Also, we need to consider how to handle this logic in the build workflow.  
Currently, it expects username and password input, but input is not possible in a build workflow.

That's why the workflow below does not terminate.
https://github.com/apache/incubator-seata/actions/runs/14169655810/job/39690431239?pr=7261

YongGoose:
@slievrly 

Can you give some feedback on this PR? 🤔

console/src/main/java/org/apache/seata/console/security/User.java
YongGoose: > If there are no plans to change the user's `username` and `password`, it would be better to change the `User` class to a `record`.
> 

@funky-eyes

What do you think about this issue?

YongGoose:
> @slievrly
> 
> Also, we need to consider how to handle this logic in the build workflow. Currently, it expects username and password input, but input is not possible in a build workflow.
> 
> That's why the workflow below does not terminate. https://github.com/apache/incubator-seata/actions/runs/14169655810/job/39690431239?pr=7261

I resolved the problem in https://github.com/apache/incubator-seata/pull/7261/commits/78dd208a0d3061743a1e8a67b70fab0cb24e1aa3 by adding environment variables during the CI stage.

console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
```diff
@@ -32,23 +34,47 @@
...
      * Init.
      */
     @PostConstruct
-    public void init() {
-        // TODO: get userInfo by db
-        user = new User();
-        user.setUsername(username);
-        user.setPassword(new BCryptPasswordEncoder().encode(password));
+    public void init() throws IOException {
+        String envUsername = System.getenv("SEATA_CONSOLE_USERNAME");
```
funky-eyes: Requiring users to read credentials (username and password) directly from environment variables would impose significant changes to their workflow and create a poor user experience. Instead, we can default to a predefined username (e.g., seata) and remove any default password. If no password is configured, the system will automatically generate a random password, log it for user reference, and notify them:

"No password was configured. A random password has been generated for security purposes. You may either:
1. Use the auto-generated password (see logs for details), or
2. Set a custom password in the configuration."

This approach ensures backward compatibility while guiding users toward secure practices.

YongGoose: > Requiring users to read credentials (username and password) directly from environment variables would impose significant changes to their workflow and create a poor user experience. Instead, we can default to a predefined username (e.g., seata) and remove any default password. If no password is configured, the system will automatically generate a random password, log it for user reference, and notify them:
> 
> "No password was configured. A random password has been generated for security purposes. You may either:
> 
> 1. Use the auto-generated password (see logs for details), or
> 2. Set a custom password in the configuration."
> 
> This approach ensures backward compatibility while guiding users toward secure practices.

Thank you for your comment!

As you pointed out, suddenly requiring users to provide credentials would lead to a poor user experience.
Following your suggestion, I will implement it so that if no password is set in the `YAML` configuration file, the system will generate a random password, log it, and provide guidance to the user.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7261/commits/a02be9533a983c03c212e1ea1b865f19a85385b1 !

YongGoose: The modified result is now being logged as shown below!

```
22:56:58.549  INFO --- [                     main] [y.CustomUserDetailsServiceImpl] [                init]  [] : No password was configured. A random password has been generated for security purposes. You may either:
1. Use the auto-generated password: [424e287f]
2. Set a custom password in the configuration.
```

YongGoose:
The problem currently occurs only with JDK 17, and I'm in the process of analyzing it.

```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.apache.catalina.startup.Tomcat
Caused by: java.lang.ExceptionInInitializerError
```

The most likely cause at this point appears to be a compatibility issue between `JDK 17` and `Tomcat version 9.0.99`.
If this turns out to be the case, we plan to upgrade Tomcat to version `9.1.00` with separate PR.

- https://bz.apache.org/bugzilla/show_bug.cgi?id=69576

YongGoose:
> The problem currently occurs only with JDK 17, and I'm in the process of analyzing it.
> 
> ```
> Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.apache.catalina.startup.Tomcat
> Caused by: java.lang.ExceptionInInitializerError
> ```
> 
> The most likely cause at this point appears to be a compatibility issue between `JDK 17` and `Tomcat version 9.0.99`. If this turns out to be the case, we plan to upgrade Tomcat to version `9.1.00` with separate PR.
> 
> * https://bz.apache.org/bugzilla/show_bug.cgi?id=69576

See https://github.com/apache/incubator-seata/pull/7329

console/src/main/java/org/apache/seata/console/security/CustomUserDetailsServiceImpl.java
funky-eyes: Well done

funky-eyes (approved):
LGTM @slievrly PTAL

YongGoose:
@funky-eyes @slievrly 

I don't think this should be merged right now.
It could cause issues with login.

I'll make the necessary changes and provide a more detailed explanation afterward.

YongGoose:
> @funky-eyes @slievrly
> 
> I don't think this should be merged right now. It could cause issues with login.
> 
> I'll make the necessary changes and provide a more detailed explanation afterward.

During the login process, the password is compared using `PasswordEncoder` internally by the `AuthenticationManager`.
However, in the previous implementation, passwords were stored in plain text, which could have caused login failures.

This has now been corrected by updating the code to encrypt passwords using `PasswordEncoder`.
I ran the integration tests shown below, and since everything works as expected, I believe it's safe to proceed with the merge now.'

PTAL 🚀

`Random password test`

```java
@Test
public void loginSuccess_shouldReturnTokenAndAddToHeader(CapturedOutput output) throws Exception {
    String logs = output.getOut();

    Pattern pattern = Pattern.compile("Use the auto-generated password: \\[(.+?)\\]");
    Matcher matcher = pattern.matcher(logs);

    assertTrue(matcher.find(), "captured password not found in logs");

    String extractedPassword = matcher.group(1);
    User user = new User("seata", extractedPassword);

    String userJson = objectMapper.writeValueAsString(user);

    MvcResult result = mockMvc.perform(post("/api/v1/auth/login")
                    .contentType(MediaType.APPLICATION_JSON)
                    .content(userJson))
            .andExpect(status().isOk())
            .andExpect(jsonPath("$.success").value(true))
            .andExpect(jsonPath("$.data").isNotEmpty())
            .andExpect(header().exists(WebSecurityConfig.AUTHORIZATION_HEADER))
            .andReturn();

    String authHeader = result.getResponse().getHeader(WebSecurityConfig.AUTHORIZATION_HEADER);
    assertNotNull(authHeader);
    assert (authHeader.startsWith(WebSecurityConfig.TOKEN_PREFIX));
}
```
`Property test`
```java
@Test
public void loginSuccess_shouldReturnTokenAndAddToHeader() throws Exception {
    User user = new User("seata", "foo");
    String userJson = objectMapper.writeValueAsString(user);

    MvcResult result = mockMvc.perform(post("/api/v1/auth/login")
                    .contentType(MediaType.APPLICATION_JSON)
                    .content(userJson))
            .andExpect(status().isOk())
            .andExpect(jsonPath("$.success").value(true))
            .andExpect(jsonPath("$.data").isNotEmpty())
            .andExpect(header().exists(WebSecurityConfig.AUTHORIZATION_HEADER))
            .andReturn();

    String authHeader = result.getResponse().getHeader(WebSecurityConfig.AUTHORIZATION_HEADER);
    assertNotNull(authHeader);
    assert (authHeader.startsWith(WebSecurityConfig.TOKEN_PREFIX));
}
```

YongGoose:
Since an integration test for login has been added, I believe it's okay to resume work on the issue below.
What do you think?

- #7158

slievrly (approved):
LGTM

PR #7280, opened 2025-04-11
feature: add oceanbase oracle support

### LINKED ISSUE #7279: [feature]:add oceanbase oracle mode support
[desc]
Oceanbase's MySQL mode is directly driven by MySQL connections. Due to the coding model, it is not convenient to distinguish between the specific patterns of querying the database according to connect, and only the oracle mode is adapted here. The code logic of Oracle is basically reused, and the reason why this part of the code is added is that the Oracle mode of OceanBase can only be connected through the OceanBase protocol. Due to the driver problem, the oracle mode of OceanBase cannot obtain the value of the auto-increment primary key in the oracle mode through getGenerateKeys, so the transaction control of the auto-increment SQL statement by sequence number is not supported.

xxsc0529:

Oceanbase's MySQL mode is directly driven by MySQL connections. Due to the coding model, it is not convenient to distinguish between the specific patterns of querying the database according to connect, and only the oracle mode is adapted here. The code logic of Oracle is basically reused, and the reason why this part of the code is added is that the Oracle mode of OceanBase can only be connected through the OceanBase protocol. Due to the driver problem, the oracle mode of OceanBase cannot obtain the value of the auto-increment primary key in the oracle mode through getGenerateKeys, so the transaction control of the auto-increment SQL statement by sequence number is not supported.

close #7279

Test code that basically reuses the oracle schema
org.apache.seata.rm.datasource.exec.OceanBaseInsertExecutorTest
org.apache.seata.rm.datasource.sql.struct.TableMetaCacheFactoryTest

use https://github.com/apache/incubator-seata-samples

oceanbase is ant group open source products.
the oracle mode of OceanBase cannot obtain the value of the auto-increment primary key in the oracle mode through getGenerateKeys, so the transaction control of the auto-increment SQL statement by sequence number is not supported.


rm-datasource/src/test/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactoryTest.java
```diff
@@ -16,6 +16,16 @@
  */
 package org.apache.seata.rm.datasource.sql.struct;
 
+<<<<<<< 2.x
```
slievrly: It seems that the conflict was handled incorrectly.

PR #7282, opened 2025-04-12
optimize: prevent unexpected NullPointerException in lookup method in FileRegistryServiceImpl class

wjwang00:
…ileRegistryServiceImpl class.

prevent unexpected NullPointerException in lookup method in FileRegistryServiceImpl class

Due to the design characteristics of `NetUtil.splitIPPortStr(endpoint)`:

- If the `endpoint` does not contain a ":", it returns `null`.
- If the `endpoint` contains a ":", it returns an array of fixed length 2 (even if the content is empty, e.g., `":"` → `["", ""]`).

As a result:
1. The condition `if (ipAndPort.length != 2)` will **never be triggered**.
2. A `NullPointerException` will be thrown when the `endpoint` does not contain a ":".

just fix bug

### DIFF
--- a/common/src/main/java/org/apache/seata/common/util/NetUtil.java
+++ b/common/src/main/java/org/apache/seata/common/util/NetUtil.java
@@ -148,6 +148,15 @@ public class NetUtil {
             if (StringUtils.isBlank(hostAddress) || StringUtils.isBlank(portStr)) {
                 throw new IllegalArgumentException("Invalid endpoint format: " + address + ". Endpoint should be in the format ip:port.");
             }
+            try {
+                int port = Integer.parseInt(portStr);
+                if (port < 1 || port > 65535) {
+                    throw new IllegalArgumentException("Invalid endpoint format: " + address + ". Port must be between 1 and 65535.");
+                }
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException("Invalid endpoint format: " + address + ". Port must be a numeric value.", e);
+            }
+
             return new String[]{hostAddress, portStr};
         } else {
             throw new IllegalArgumentException("Invalid endpoint format: " + address + ". Endpoint should be in the format ip:port.");
--- a/common/src/main/java/org/apache/seata/common/util/NetUtil.java
+++ b/common/src/main/java/org/apache/seata/common/util/NetUtil.java
@@ -138,18 +138,20 @@ public class NetUtil {
         if (address.charAt(0) == '[') {
             address = removeBrackets(address);
         }
-        String[] serverAddArr = null;
         int i = address.lastIndexOf(Constants.IP_PORT_SPLIT_CHAR);
         if (i > -1) {
-            serverAddArr = new String[2];
-            String hostAddress = address.substring(0,i);
+            String hostAddress = address.substring(0, i);
             if (hostAddress.contains("%")) {
                 hostAddress = hostAddress.substring(0, hostAddress.indexOf("%"));
             }
-            serverAddArr[0] = hostAddress;
-            serverAddArr[1] = address.substring(i + 1);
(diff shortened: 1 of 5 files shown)


discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/FileRegistryServiceImpl.java
```diff
@@ -89,7 +89,7 @@ public List<InetSocketAddress> lookup(String key) throws Exception {
         List<InetSocketAddress> inetSocketAddresses = new ArrayList<>();
         for (String endpoint : endpoints) {
             String[] ipAndPort = NetUtil.splitIPPortStr(endpoint);
-            if (ipAndPort.length != 2) {
+            if (ipAndPort == null || ipAndPort[0].isEmpty() || ipAndPort[1].isEmpty() ) {
```
slievrly: It is suggested that the judgments of ipAndPort[0] and ipAndPort[1] converge to the NetUtil.splitIPPortStr method. It is necessary to determine whether the outer caller ipAndPort is null and its length (there is a reference to the array below).

wjwang00: Sure, I have adjusted the `NetUtil.splitIPPortStr` method to ensure that the returned result is definitely a string array of length 2 with the correct format. The caller no longer needs to validate the returned result.

slievrly (commented):
You need to adjust the UT of NetUtilTest to make it as expected.

https://github.com/apache/incubator-seata/actions/runs/14420166446/job/40441706978?pr=7282

https://github.com/apache/incubator-seata/blob/2.x/common/src/test/java/org/apache/seata/common/util/NetUtilTest.java

wjwang00:
> You need to adjust the UT of NetUtilTest to make it as expected.
> 
> https://github.com/apache/incubator-seata/actions/runs/14420166446/job/40441706978?pr=7282
> 
> https://github.com/apache/incubator-seata/blob/2.x/common/src/test/java/org/apache/seata/common/util/NetUtilTest.java

ok,I appreciate your reminder.

common/src/main/java/org/apache/seata/common/util/NetUtil.java
```diff
@@ -138,18 +138,20 @@ public static String[] splitIPPortStr(String address) {
...
-            String hostAddress = address.substring(0,i);
+            String hostAddress = address.substring(0, i);
             if (hostAddress.contains("%")) {
                 hostAddress = hostAddress.substring(0, hostAddress.indexOf("%"));
             }
-            serverAddArr[0] = hostAddress;
-            serverAddArr[1] = address.substring(i + 1);
+            String portStr = address.substring(i + 1);
+            if (StringUtils.isBlank(hostAddress) || StringUtils.isBlank(portStr)) {
+                throw new IllegalArgumentException("Invalid endpoint format: " + address + ". Endpoint should be in the format ip:port.");
```
slievrly, line 149: Pls. add UT case for this exception.

common/src/main/java/org/apache/seata/common/util/NetUtil.java
```diff
@@ -138,18 +138,20 @@ public static String[] splitIPPortStr(String address) {
...
+            }
+            return new String[]{hostAddress, portStr};
```
slievrly, line 160: It is best to verify the numeric type of the port and the range of the port.

wjwang00: OK, I have verified the numeric type of the port and the range of the port, and I have also added some test cases to cover the new code. Thank you.

slievrly (approved):
LGTM

PR #7287, opened 2025-04-13
test: Refactored testGetErrorMsgWithValidCodeReturnsExpectedMsg to use parameterized unit testing

Monilnarang:

**Summary**: 
- Parameterized `testGetErrorMsgWithValidCodeReturnsExpectedMsg `

**Elaboration**: 
- The same method calls (`getMsg()`) is repeated multiple times with different inputs, making the test harder to maintain and extend. 
- When a test fails, JUnit only shows which assertion failed, but not which specific input caused the failure.
- Adding new test cases requires copying and pasting another Assertions.assertEquals(...) statement instead of simply adding new data.

To accomplish this, I have retrofitted the test into a parameterized unit test. This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions.

**Test execution results after changes:**
Screenshot 2025-04-13 at 11 24 21 AM
 

Successful tests run


slievrly (approved):
LGTM

PR #7294, opened 2025-04-14
test: improved test SqlServerInsertRecognizerTest

Monilnarang:

**Summary**: 
- Separated independent logics inside the aforementioned test, i.e. checking for explicit values and default value. 
- Parameterized the test with explicit values

**Elaboration**: 
- The same method calls (`getSQLStatement`, creating recognizer) were repeated multiple times with different inputs, making the test harder to maintain and extend. 
- The test also had 2 independent logics of checking for explicit values and default value under single test. 
- When a test fails, JUnit only shows which assertion failed, but not which specific input caused the failure.
- Adding new test cases requires copying and pasting another block of statements instead of simply adding new data.

To accomplish this, I separated the 2 logics into different test and then retrofitted the explicit values test into a parameterized unit test. This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions. The separated tests also increase the fault detection capabilities of the suit. 

**Test execution results after changes:** 
Screenshot 2025-04-13 at 11 10 49 PM

Successful tests run


slievrly (approved):
LGTM

PR #7311, opened 2025-04-23
Optimize: Remove `CompressUtil` and use `byte flag` for compression detection

### LINKED ISSUE #4071: optimize compressUtil
## Why you need it?
1. io.seata.common.util.CompressUtil 和 io.seata.compressor.gzip.GzipCompressor 重合了，CompressUtil可以直接去掉
2. CompressUtil比GzipCompressor多了个isCompressData的方法，有两种方案，a. 为Compressor的每个实现都添加isCompress的方法，然后直接把使用了isCompressData的地方迁移过去；b. 把用到了isCompressData的地方优化掉，可以把根据isCompressData判断数据是否被加密，改成加一个字节直接保存数据是否已加密会更好。

针对第2点的还有个问题，当前的支持的Compressor的对于是否被加密了，都是只能通过判断前2-4位的magic number来判断，不排除有一定的误判的可能性，虽然在当前的使用场景中不会出现问题，但是在客观上依然属于一个小问题，个人建议采用b方案。

如果有兴趣优化的，可以在下方留言`please assign it to me`

## How it could be?

YongGoose:
fix #4071


YongGoose:
@funky-eyes 

When the data size is small, **the cost of compression** itself can exceed that of simply storing the original data. 
To **avoid** unnecessary compression overhead, i introduce a `threshold`.
Additionally, if `compression` doesn't reduce the data size, no compression flag is added.

Could this approach potentially lead to any unforeseen issues?

compressor/seata-compressor-gzip/src/main/java/org/apache/seata/compressor/gzip/GzipUtil.java
```diff
@@ -64,4 +64,74 @@ public static byte[] decompress(byte[] bytes) {
...
+
+    /**
+     * Check if the data has the compression flag and is marked as compressed.
+     * This is the preferred method to check compression status when using the new format.
+     *
+     * @param bytes the bytes to check
+     * @return true if the first byte indicates the data is compressed
+     */
+    public static boolean hasCompressionFlag(byte[] bytes) {
+        return bytes != null && bytes.length > 0 && bytes[0] == COMPRESSED_FLAG;
```
slievrly: If a user upgrades to this version but uses existing data files, the new seata-server will attempt to restore incomplete transactions using these files upon startup. When `bytes[0] == COMPRESSED_FLAG`, it does not necessarily mean that compression is enabled, which could lead to decoding errors.

YongGoose: I think that's a good idea, and I agree with you.

Considering users who may continue using existing data files, I modified the logic in `BranchSession` to first check the header against `GZIPInputStream.GZIP_MAGIC` to determine whether compression is enabled. If that doesn't match, it then checks against the newly added `COMPRESSED_FLAG`.

```java
private String decodeLockKey(byte[] lockKeyBytes) {
    if (GzipUtil.isCompressData(lockKeyBytes)) {
        try {
            return new String(GzipUtil.uncompress(lockKeyBytes));
        } catch (IOException e) {
            throw new RuntimeException("decompress lockKey error", e);
        }
    }

    if (GzipUtil.hasCompressionFlag(lockKeyBytes)) {
        return new String(GzipUtil.decompressWithFlag(lockKeyBytes));
    }
    return new String(lockKeyBytes);
}
```

However, there's still a potential issue here. If an existing data file wasn't compressed, but its header coincidentally matches the new `COMPRESSED_FLAG`, the system could mistakenly identify the file as compressed.

I did increase the size of the COMPRESSED_FLAG to `2 bytes`, considering this case.

But, I believe this problem still could lead to problems—what are your thoughts on this?

slievrly: Compression implementations typically have unique magic numbers to identify their formats, so why do we need additional independent flag bits? Is it solely to accommodate merging two compression types at the cost of complexity? This approach seems counterproductive. If necessary, I would prefer retaining existing logic rather than introducing redundant flags.

YongGoose: I agree with your opinion as well.
This approach is quite complex and could potentially introduce bugs for users who rely on existing data.
Therefore, I'm planning to close this PR — what do you think?

Also, if possible, it would be helpful to note in the issue that we won't be proceeding with this change.
Otherwise, another contributor might come across the issue later and open a similar PR. 😅

PR #7333, opened 2025-05-06
bugfix: Add scheduled tasks to clean up futures

### LINKED ISSUE #5283: ClientOnResponseProcessor#mergeMsgMap&futures is a risk of memory leaks

### Ⅰ. Issue Description
timeout发生如tc网络出现故障或宕机，消息实际上已达到tc时，但无法将response响应时候没有去清除mergemsgmap和futures里的数据，而是单纯的靠server的response下来再remove（详见ClientOnResponseProcessor#process），这带来一个问题，server非优雅下线的时候，这个futures和mergemsgmap无法清理，残留了数据
AbstractNettyRemotingClient#MergedSendRunnable 中也仅对发送失败时清理了futures（该内存泄露问题的前提是client发送成功，tc响应失败），mergemsgmap并没有清理
image

ps：该情况出现概率非常非常低，通过阅读这块代码逻辑偶然发现

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- Seata version:  
- OS :
- Others:

Tsukilc:

Use scheduled tasks to detect whether futures have timeout information before sending messages. If there is, clear it to avoid memory leaks

fixes #5283

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
@@ -16,6 +16,16 @@
  */
 package org.apache.seata.core.rpc.netty;
 
+import static org.apache.seata.common.exception.FrameworkErrorCode.NoAvailableService;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandler.Sharable;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+import io.netty.util.concurrent.EventExecutorGroup;
 import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -242,6 +249,12 @@ public void destroy() {
         if (mergeSendExecutorService != null) {
             mergeSendExecutorService.shutdown();
         }
+
+        futures.clear();
```
funky-eyes: Why is it necessary to clean up these maps? When a RemotingClient executes the destroy method, this instance should no longer be used. In theory, it should be garbage collected, or the application is shutting down.

Tsukilc: Yes. manually clearing the maps is indeed unnecessary.I have removed it.

funky-eyes (commented):
I believe the following PR can solve this issue:
https://github.com/apache/incubator-seata/pull/7095

PR #7344, opened 2025-05-18
optimize: raft mode performs transaction size check in advance

funky-eyes:

### DIFF
--- a/server/src/main/java/org/apache/seata/server/session/BranchSession.java
+++ b/server/src/main/java/org/apache/seata/server/session/BranchSession.java
@@ -446,7 +446,7 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
             byte[] clientIdBytes,
             byte[] applicationDataBytes,
             byte[] xidBytes) {
-        // trascationId
+        // transactionId
         // branchId
         // resourceIdBytes.length
         // lockKeyBytes.length
--- a/server/src/main/java/org/apache/seata/server/session/BranchSession.java
+++ b/server/src/main/java/org/apache/seata/server/session/BranchSession.java
@@ -446,15 +446,6 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
             byte[] clientIdBytes,
             byte[] applicationDataBytes,
             byte[] xidBytes) {
-        // transactionId
-        // branchId
-        // resourceIdBytes.length
-        // lockKeyBytes.length
-        // clientIdBytes.length
-        // applicationDataBytes.length
-        // xidBytes.size
-        // statusCode
-        // branchType
         return 8 // trascationId
                 + 8 // branchId
                 + 4 // resourceIdBytes.length
--- a/server/src/main/java/org/apache/seata/server/session/BranchSession.java
+++ b/server/src/main/java/org/apache/seata/server/session/BranchSession.java
@@ -391,6 +391,15 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
         return result;
     }
 
+    /**
+     * Checks if the serialized size of the branch session exceeds the configured maximum limit.
+     * This method calculates the size based on the current state of the branch session's fields
+     * (resourceId, lockKey, clientId, applicationData, xid). If the size exceeds the limit,
+     * it attempts to compress the lockKey and re-checks the size.
+     *
+     * @throws TransactionException if the calculated size of the branch session (after potential lockKey compression)
+     *                              exceeds the {@link StoreConfig#getMaxBranchSessionSize()}.
+     */
     public void checkSize() throws TransactionException {
         byte[] resourceIdBytes = resourceId != null ? resourceId.getBytes() : null;
 
@@ -409,6 +418,19 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
         }
     }
 
+    /**
+     * Checks if the serialized size of the branch session exceeds the configured maximum limit.
+     * This method calculates the size based on the current state of the branch session's fields
+     * (resourceId, lockKey, clientId, applicationData, xid). If the size exceeds the limit,
+     * it attempts to compress the lockKey and re-checks the size.
+     * @param resourceIdBytes      Byte array representation of the resource ID.
+     * @param lockKeyBytes         Byte array representation of the lock key.
+     * @param clientIdBytes        Byte array representation of the client ID.
+     * @param applicationDataBytes Byte array representation of the application data.
+     * @param xidBytes             Byte array representation of the XID.
+     * @throws TransactionException if the calculated size of the branch session (after potential lockKey compression)
+     *                              exceeds the {@link StoreConfig#getMaxBranchSessionSize()}.
+     */
     private void checkSize(
             byte[] resourceIdBytes,
             byte[] lockKeyBytes,
@@ -440,6 +462,17 @@ public class BranchSession implements Lockable, Comparable<BranchSession>, Sessi
         }
     }
 
+    /**
+     * Calculates the total size of the branch session in bytes based on its constituent parts.
+     * This includes fixed-size fields and the variable lengths of string fields represented as byte arrays.
+     *
+     * @param resourceIdBytes      Byte array representation of the resource ID.
+     * @param lockKeyBytes         Byte array representation of the lock key.
+     * @param clientIdBytes        Byte array representation of the client ID.
+     * @param applicationDataBytes Byte array representation of the application data.
+     * @param xidBytes             Byte array representation of the XID.
+     * @return The calculated total size of the branch session in bytes.
+     */
     private int calBranchSessionSize(
             byte[] resourceIdBytes,
             byte[] lockKeyBytes,
--- a/server/src/main/java/org/apache/seata/server/session/BranchSession.java
+++ b/server/src/main/java/org/apache/seata/server/session/BranchSession.java
@@ -16,19 +16,21 @@
  */
 package org.apache.seata.server.session;
 
+import static org.apache.seata.core.model.LockStatus.Locked;
+
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-
 import org.apache.seata.common.util.BufferUtils;
 import org.apache.seata.common.util.CompressUtil;
 import org.apache.seata.core.exception.TransactionException;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/seata/server/session/BranchSession.java
```diff
@@ -405,31 +383,92 @@ public byte[] encode() {
...
+        }
+    }
+
+    private int calBranchSessionSize(
+            byte[] resourceIdBytes,
+            byte[] lockKeyBytes,
+            byte[] clientIdBytes,
+            byte[] applicationDataBytes,
+            byte[] xidBytes) {
+        // trascationId
```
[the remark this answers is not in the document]
YongGoose: I’ve got to say, this is a nice feature of Copilot. 👍🏻😁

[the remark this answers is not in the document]
funky-eyes: > I’ve got to say, this is a nice feature of Copilot. 👍🏻😁

You're right

server/src/main/java/org/apache/seata/server/session/BranchSession.java
```diff
@@ -405,31 +383,92 @@ public byte[] encode() {
...
+            byte[] xidBytes) {
+        // transactionId
+        // branchId
+        // resourceIdBytes.length
+        // lockKeyBytes.length
+        // clientIdBytes.length
+        // applicationDataBytes.length
+        // xidBytes.size
+        // statusCode
+        // branchType
```
YongGoose: What is the purpose of this comment?

server/src/main/java/org/apache/seata/server/session/BranchSession.java
```diff
@@ -405,31 +383,92 @@ public byte[] encode() {
...
-            + 4 // xidBytes.size
-            + 1 // statusCode
-            + (resourceIdBytes == null ? 0 : resourceIdBytes.length)
-            + (lockKeyBytes == null ? 0 : lockKeyBytes.length)
-            + (clientIdBytes == null ? 0 : clientIdBytes.length)
-            + (applicationDataBytes == null ? 0 : applicationDataBytes.length)
-            + (xidBytes == null ? 0 : xidBytes.length)
-            + 1; //branchType
-        return size;
+    public void checkSize() throws TransactionException {
```
YongGoose, line 403: Having a `Javadoc` comment for this method would make it easier for community members to understand its purpose later on. 😁

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
```diff
@@ -682,29 +677,65 @@ public byte[] encode() {
...
+        byteBuffer.put((byte) status.getCode());
         BufferUtils.flip(byteBuffer);
         byte[] result = new byte[byteBuffer.limit()];
         byteBuffer.get(result);
         return result;
     }
 
-    private int calGlobalSessionSize(byte[] byApplicationIdBytes, byte[] byServiceGroupBytes, byte[] byTxNameBytes,
-        byte[] xidBytes, byte[] applicationDataBytes) {
+    public void checkSize() throws TransactionException {
```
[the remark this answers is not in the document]
YongGoose: IMHO, it's good idea
- https://github.com/apache/incubator-seata/pull/7344#discussion_r2094436319

WDYT?

server/src/test/java/org/apache/seata/server/session/GlobalSessionTest.java
```diff
@@ -76,6 +77,31 @@ public void beginTest(GlobalSession globalSession) throws Exception {
...
+        for (int i = 0; i < size; i++) {
+            sb.append('a');
+        }
+        String str = sb.toString();
+        globalSession.setApplicationData(str);
+        Assertions.assertThrows(TransactionException.class, globalSession::checkSize);
+        globalSession.setApplicationData(null);
+        globalSession.setXid(str);
+        Assertions.assertThrows(TransactionException.class, globalSession::checkSize);
+        GlobalSession globalSession1 = new GlobalSession(null,str,null,60000,true);
```
YongGoose:
```suggestion
        GlobalSession globalSession1 = new GlobalSession(null, str, null, 60000, true);
```

Just minor things 😅

YongGoose (changes requested):
Left few comments!
Please take a look :)

server/src/main/java/org/apache/seata/server/session/BranchSession.java
funky-eyes: This is indeed unnecessary content, and I will delete it.

server/src/main/java/org/apache/seata/server/session/BranchSession.java
funky-eyes: ok

server/src/main/java/org/apache/seata/server/session/GlobalSession.java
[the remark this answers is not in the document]
funky-eyes: I believe the check methods of these two classes are not the same, because GlobalSession does not contain the lockKey property.

server/src/main/java/org/apache/seata/server/session/BranchSession.java
funky-eyes:
```suggestion
```

server/src/main/java/org/apache/seata/server/session/BranchSession.java
funky-eyes: done

server/src/main/java/org/apache/seata/server/session/BranchSession.java
```diff
@@ -336,33 +336,11 @@ public byte[] encode() {
...
-                size += lockKeyBytes.length;
-            }
-
-            if (size > MAX_BRANCH_SESSION_SIZE) {
-                throw new RuntimeException(
-                    "compress branch session size exceeded, compressSize : " + size + " maxBranchSessionSize : "
-                        + MAX_BRANCH_SESSION_SIZE);
-            }
+        if (!RaftServerManager.isRaftMode()) {
+            checkSize(resourceIdBytes, lockKeyBytes, clientIdBytes, applicationDataBytes, xidBytes);
```
YongGoose, line 340: I have one question.

The title of the PR says that the size check for transactions is performed in advance in Raft mode, but the relevant code seems to perform the check in advance even in modes other than Raft.

If I’ve misunderstood something, I apologize in advance :)

YongGoose (commented):
I’ve written one question above!

Also, I think it would be great to add a unit test to verify that the `checkSize` method is called when in Raft mode.

If that unit test is added, I believe a full integration test may not be strictly necessary.

funky-eyes:
> I’ve written one question above!
> 
> Also, I think it would be great to add a unit test to verify that the `checkSize` method is called when in Raft mode.
> 
> If that unit test is added, I believe a full integration test may not be strictly necessary.

In "file" mode, because GlobalSession and BranchSession instances need to be written to disk, their encode method is inevitably called when these instances are added. At this stage, the transaction size is validated. Conversely, "db" and "redis" modes do not require writing transaction information to disk, so the encode method is not called at all; consequently, these modes do not perform a transaction size check.

In "raft" mode, however, transaction information is stored entirely in memory. It's only when a Raft snapshot needs to be created that these in-memory transactions are flushed to disk. If the encode method is executed at this point, it's too late. This can lead to snapshot creation failure because if a transaction in memory exceeds the maximum size limit, a runtime exception is thrown, causing the snapshot process to fail. Therefore, for "raft" mode, the transaction size check is performed belatedly. The purpose of this pull request (PR) is to implement this check at an earlier stage to prevent the aforementioned issue.

funky-eyes:
> I’ve written one question above!
> 
> Also, I think it would be great to add a unit test to verify that the `checkSize` method is called when in Raft mode.
> 
> If that unit test is added, I believe a full integration test may not be strictly necessary.

It's not possible to construct multiple Raft nodes to form a Raft cluster within unit tests. Therefore, I believe that calling the checkSize method in RaftSessionManager is not fundamentally different from directly calling the checkSize method of GlobalSession and BranchSession, unless we use integration tests instead of unit tests.

YongGoose:
> > I’ve written one question above!
> > Also, I think it would be great to add a unit test to verify that the `checkSize` method is called when in Raft mode.
> > If that unit test is added, I believe a full integration test may not be strictly necessary.
> 
> In "file" mode, because GlobalSession and BranchSession instances need to be written to disk, their encode method is inevitably called when these instances are added. At this stage, the transaction size is validated. Conversely, "db" and "redis" modes do not require writing transaction information to disk, so the encode method is not called at all; consequently, these modes do not perform a transaction size check.
> 
> In "raft" mode, however, transaction information is stored entirely in memory. It's only when a Raft snapshot needs to be created that these in-memory transactions are flushed to disk. If the encode method is executed at this point, it's too late. This can lead to snapshot creation failure because if a transaction in memory exceeds the maximum size limit, a runtime exception is thrown, causing the snapshot process to fail. Therefore, for "raft" mode, the transaction size check is performed belatedly. The purpose of this pull request (PR) is to implement this check at an earlier stage to prevent the aforementioned issue.

Oh, I see !!
Thank you for the detailed explanation. 🙇🏻

Goose-Bot-max (approved):
LGTM 👍🏻
Integration tests can be added here or in a separate PR—both options are fine :)

YongGoose (approved):
LGTM 👍🏻
Integration tests can be added here or in a separate PR—both options are fine :)

YongGoose:
> LGTM 👍🏻 Integration tests can be added here or in a separate PR—both options are fine :)

I think it would be good to proceed with the integration test in the following way.
- https://github.com/YongGoose/incubator-seata/commit/16e477eaea90cef711af1fd7a65b0e6e9f32e66a

While working locally to create a PR for your branch in order to help you, I got stuck on the part where the RaftServerManager is initialized. 😭

Have a good night!

funky-eyes:
> > LGTM 👍🏻 Integration tests can be added here or in a separate PR—both options are fine :)
> 
> I think it would be good to proceed with the integration test in the following way.
> 
> * [YongGoose@16e477e](https://github.com/YongGoose/incubator-seata/commit/16e477eaea90cef711af1fd7a65b0e6e9f32e66a)
> 
> While working locally to create a PR for your branch in order to help you, I got stuck on the part where the RaftServerManager is initialized. 😭
> 
> Have a good night!

Thank you for your assistance. However, I believe that while the mock approach is meaningful for testing and verification in a single-instance deployment, it may not fully and accurately validate the process of adding transactions in Raft mode, which is a cluster-based setup.

PR #7315, opened 2025-04-25
test: Refactor log testing use `ListAppender` for accurate log capture

YongGoose:
- [x] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did
Currently, I'm encountering an unexpected issue during the CI stage.
image

I suspect it might be related to the `LOGGER`, so to address this, I'm planning to use `ListAppender` to **capture** the logs and use the captured output for assertions in the test.

> For more details, please refer to the conversation in DingTalk.

### Ⅱ. Does this pull request fix one issue?

No

### Ⅲ. Why don't you add test cases (unit test/integration test)? 
Don't have to

### Ⅳ. Describe how to verify it
X

### Ⅴ. Special notes for reviews
There’s one thing we need to discuss. ‼️

In the original test, we used `mockStatic` to control the static methods of `LOGGER`.  
> `mockStatic` allows us to mock static methods.

However, in the refactored code, we no longer use `mockStatic`, which means we can't control the log level of `LOGGER` when it's printing logs. As a result, we’re unable to verify that logs are not printed at certain levels.

To work around this, I configured the log level when collecting logs, and instead verified that appropriate logs are captured at the expected level.

What do you think?

PR #7345, opened 2025-05-19
optimize: Add some checks to RegistryFactory

YoWuwuuuw:

1. Use TreeSet to Increase the MultiRegistryFactory check for repeated registryType
2. Increase the inspection of RegistryFactory to air registryTypeName(align MultiRegistryFactory)
3. Put the logic of throwing NotSupportYetException into RegistryType.getType ()

https://github.com/apache/incubator-seata/discussions/7342

Covers two factory

### DIFF
--- a/discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
+++ b/discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
@@ -49,7 +49,6 @@ public class MultiRegistryFactory {
 
     private static List<RegistryService> buildRegistryServices() {
         List<RegistryService> registryServices = new ArrayList<>();
-        Set<String> processedRegistryTypes = new HashSet<>();
         String registryTypeNamesStr = ConfigurationFactory.CURRENT_FILE_INSTANCE.getConfig(
                 ConfigurationKeys.FILE_ROOT_REGISTRY + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR + ConfigurationKeys.FILE_ROOT_TYPE);
 
@@ -70,15 +69,14 @@ public class MultiRegistryFactory {
                 throw new NotSupportYetException("not support registry type: " + registryTypeName);
             }
 
-            if (processedRegistryTypes.contains(registryType.name())) {
+            RegistryService registryService = EnhancedServiceLoader
+                    .load(RegistryProvider.class, Objects.requireNonNull(registryType).name()).provide();
+
+            if(registryServices.contains(registryService)) {
                 LOGGER.warn("The duplicate registration center type '{}' was found in the configuration and has been skipped.", registryType.name());
                 continue;
             }
-
-            RegistryService registryService = EnhancedServiceLoader
-                    .load(RegistryProvider.class, Objects.requireNonNull(registryType).name()).provide();
             registryServices.add(registryService);
-            processedRegistryTypes.add(registryType.name());
         }
         return registryServices;
     }
--- a/discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
+++ b/discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
@@ -17,8 +17,10 @@
 package org.apache.seata.discovery.registry;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 import org.apache.seata.common.ConfigurationKeys;
 import org.apache.seata.common.Constants;
@@ -48,6 +50,7 @@ public class MultiRegistryFactory {
 
     private static List<RegistryService> buildRegistryServices() {
         List<RegistryService> registryServices = new ArrayList<>();
+        Set<String> processedRegistryTypes = new HashSet<>();
         String registryTypeNamesStr =
             ConfigurationFactory.CURRENT_FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_REGISTRY
                 + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR + ConfigurationKeys.FILE_ROOT_TYPE);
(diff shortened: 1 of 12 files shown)


discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
```diff
@@ -48,26 +49,36 @@ public static List<RegistryService> getInstances() {
...
             RegistryType registryType;
             try {
                 registryType = RegistryType.getType(registryTypeName);
             } catch (Exception exx) {
                 throw new NotSupportYetException("not support registry type: " + registryTypeName);
             }
+
+            if (processedRegistryTypes.contains(registryType.name())) {
+                LOGGER.warn("The duplicate registration center type '{}' was found in the configuration and has been skipped.", registryType.name());
+                continue;
```
YongGoose: I saw a test that verifies a method was called once by checking if `instances.size()` equals 1. 
How about also adding a test to verify that the corresponding log message is properly output?

What do you think?

discovery/seata-discovery-core/src/test/java/org/apache/seata/discovery/registry/RegistryFactoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    public void testGetInstancesWithBlankRegistryType() throws Throwable {
+        System.setProperty(REGISTRY_TYPE_KEY, "");
+
+        RegistryService instance = invokeBuildRegistryService();
+        Assertions.assertEquals(instance.getClass(), FileRegistryServiceImpl.class);
+    }
+
+    /**
+     * Use reflection to call the buildRegistryServices method
```
YongGoose: It looks like the comment should refer to `buildRegistryService` instead of `buildRegistryServices`.

YongGoose (commented):
It looks good 👍🏻

Left some comments!

And also, If the implementation is created through `RegistryType`, it would be better to verify that the intended implementation is created (e.g., using `assertEquals`) rather than using `assertNotNull`.

discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
YoWuwuuuw: Ok, learned and used the method you wrote before in #7315 (it's awesome)

YongGoose:
@slievrly 

Please take a look :)
image

discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/MultiRegistryFactory.java
```diff
@@ -48,25 +47,33 @@ public static List<RegistryService> getInstances() {
...
-        String registryTypeNamesStr =
-            ConfigurationFactory.CURRENT_FILE_INSTANCE.getConfig(ConfigurationKeys.FILE_ROOT_REGISTRY
-                + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR + ConfigurationKeys.FILE_ROOT_TYPE);
+        String registryTypeNamesStr = ConfigurationFactory.CURRENT_FILE_INSTANCE.getConfig(
+                ConfigurationKeys.FILE_ROOT_REGISTRY + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR + ConfigurationKeys.FILE_ROOT_TYPE);
+
         if (StringUtils.isBlank(registryTypeNamesStr)) {
             registryTypeNamesStr = RegistryType.File.name();
         }
         String[] registryTypeNames = registryTypeNamesStr.split(Constants.REGISTRY_TYPE_SPLIT_CHAR);
```
funky-eyes: Why not convert this directly to a TreeSet? This would allow for case-insensitivity and also handle deduplication

YongGoose: +1
Also, it would be great to add a test case that covers this scenario.

discovery/seata-discovery-core/src/test/java/org/apache/seata/discovery/registry/MultiRegistryFactoryTest.java
YongGoose, line 1: Since it’s been changed to a `TreeSet`, it might be good to add a few more test cases
1. When the same `RegistryType` is provided with different casing (e.g., `File`, `FILE`), does it recognize them as the same?
    - This could be verified by checking that no log is printed.

2. When more than one RegistryType is provided, does **the log output as expected**?
3. When a value not defined in RegistryType is passed, is the appropriate exception thrown? (After removing the try-catch block, this test should be added.)

discovery/seata-discovery-core/src/main/java/org/apache/seata/discovery/registry/RegistryFactory.java
```diff
@@ -45,8 +45,12 @@ public static RegistryService getInstance() {
...
-            ConfigurationKeys.FILE_ROOT_REGISTRY + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR
-                + ConfigurationKeys.FILE_ROOT_TYPE);
+                ConfigurationKeys.FILE_ROOT_REGISTRY + ConfigurationKeys.FILE_CONFIG_SPLIT_CHAR + ConfigurationKeys.FILE_ROOT_TYPE);
+
+        if (StringUtils.isBlank(registryTypeName)) {
+            registryTypeName = RegistryType.File.name();
+        }
+
         LOGGER.info("use registry center type: {}", registryTypeName);
         try {
```
YongGoose: This try-catch block could be removed, as it simply catches the exception and rethrows it without adding any additional handling.

YongGoose (commented):
We’re almost there!
Just a few more things to add and it should be good to go.

discovery/seata-discovery-core/src/test/java/org/apache/seata/discovery/registry/MultiRegistryFactoryTest.java
YoWuwuuuw: done

YongGoose:
@YoWuwuuuw 

I noticed a very minor change was needed, so I opened a PR to your branch.
https://github.com/YoWuwuuuw/incubator-seata/pull/10

Feel free to ping me when you have questions :)

YongGoose (approved):
LGTM

funky-eyes (approved):
LGTM

PR #7349, opened 2025-05-19
bugfix: Resolve `NullPointerException` in `EtcdRegistryServiceImplMockTest`

### LINKED ISSUE #7347: EtcdRegistryServiceImplMockTest.testUnsubscribe:237 » NullPointer
Image

Image

YongGoose:

fixes #7347

### DIFF
--- a/discovery/seata-discovery-etcd3/src/main/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImpl.java
+++ b/discovery/seata-discovery-etcd3/src/main/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImpl.java
@@ -16,6 +16,7 @@
  */
 package org.apache.seata.discovery.registry.etcd3;
 
+import static io.netty.util.CharsetUtil.UTF_8;
 
 import io.etcd.jetcd.ByteSequence;
 import io.etcd.jetcd.Client;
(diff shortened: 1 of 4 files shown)


discovery/seata-discovery-etcd3/src/main/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImpl.java
```diff
@@ -178,7 +181,11 @@ public void unsubscribe(String cluster, Watch.Listener listener) throws Exceptio
                     .collect(Collectors.toSet());
             listenerMap.put(cluster, newSubscribeSet);
         }
-        watcherMap.remove(cluster).stop();
+
+        EtcdWatcher watcher = watcherMap.remove(cluster);
+        if (watcher != null) {
+            watcher.stop();
```
YongGoose, line 187: @funky-eyes 

I added a null check here, so even if an element is not found in the map, it won't cause a `NullPointerException`.

discovery/seata-discovery-etcd3/src/main/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImpl.java
```diff
@@ -401,20 +411,24 @@ public EtcdWatcher(String cluster, Watch.Listener listener) {
...
         }
 
         /**
          * stop this task
          */
         public void stop() {
-            this.watcher.close();
+            if (this.watcher != null) {
+                this.watcher.close();
+            }
```
YongGoose, line 431: @funky-eyes 

I add a null check here. What do you think?

YongGoose:
image

discovery/seata-discovery-etcd3/src/main/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImpl.java
funky-eyes: I don't quite understand why the NPE in the unsubscribe is related to this.

YongGoose: Currently, the `watcher` is initialized not in the `constructor`, but within the `run` method of the `EtcdWatcher` class.
In the test code, we used `thenAnswer` to return a `mockWatcher`, ensuring that the watcher is properly initialized in the run method. However, when running the test `10,000` times using `@RepeatedTest`, we encountered 7 failures.

All 7 failures were caused by a `NullPointerException` triggered when the `stop` method was called while the watcher was still null.

image


It appears that, intermittently, the unsubscribe method is executed before the run method, leading to this issue.

- After adding a null check in the `stop` method, the issue has not occurred again.\
That's why i believe this piece of code is related to the NPE issue.

funky-eyes: > Currently, the `watcher` is initialized not in the `constructor`, but within the `run` method of the `EtcdWatcher` class. In the test code, we used `thenAnswer` to return a `mockWatcher`, ensuring that the watcher is properly initialized in the run method. However, when running the test `10,000` times using `@RepeatedTest`, we encountered 7 failures.
> 
> All 7 failures were caused by a `NullPointerException` triggered when the `stop` method was called while the watcher was still null.
> 
> image
> 
> It appears that, intermittently, the unsubscribe method is executed before the run method, leading to this issue.
> 
> * After adding a null check in the `stop` method, the issue has not occurred again.
>   That's why i believe this piece of code is related to the NPE issue.

I understand now, thank you for your explanation.

funky-eyes (approved):
LGTM

PR #7356, opened 2025-05-21
bugfix: fix codecov bug

YongGoose:

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,6 +17,7 @@
 
 codecov:
   require_ci_to_pass: true
+  branch: 2.x
 coverage:
   status:
     patch: true
@@ -24,7 +25,6 @@ coverage:
       default:
         threshold: 1%
         if_not_found: success
-        base: 2.x
     changes: true
   precision: 2
   range: "50...100"
--- a/codecov.yml
+++ b/codecov.yml
@@ -16,16 +16,16 @@
 #
 
 codecov:
-  require_ci_to_pass: yes
+  require_ci_to_pass: true
 coverage:
   status:
-    patch: yes
+    patch: true
     project:
       default:
         threshold: 1%
         if_not_found: success
-        target: 2.x
-    changes: yes  #only care the changed files
+        base: 2.x
+    changes: true
   precision: 2
   range: "50...100"
 ignore:
@@ -42,4 +42,4 @@ ignore:
 comment:
   layout: "diff,flags,files,reach"
   behavior: default
-  require_changes: no
+  require_changes: false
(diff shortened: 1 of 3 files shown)


jsbxyyx (approved):
LGTM

jsbxyyx:
The `base` property is Deprecated. suggest use `branch` 

https://docs.codecov.com/docs/codecov-yaml#default-branch

https://docs.codecov.com/docs/commit-status#branches

funky-eyes (commented):
Please register this PR in the 2.x.md

YongGoose:
> Please register this PR in the 2.x.md

I just got home! 🥹
I’ll review that comment first and then make the addition.
- https://github.com/apache/incubator-seata/pull/7356#issuecomment-2897072109

YongGoose:
@funky-eyes @jsbxyyx 

PTAL ⚡️

YongGoose:
I think we can merge this PR once we confirm that `Codecov` is working properly.

codecov.yml
```diff
@@ -16,16 +16,16 @@
...
   status:
-    patch: yes
+    patch: true
     project:
       default:
         threshold: 1%
         if_not_found: success
-        target: 2.x
-    changes: yes  #only care the changed files
+        base: 2.x
```
YongGoose: It looks like this setting was added to limit the comparison to only the modified files, but in reality, it behaves differently.
It actually detects changes in code coverage even for lines that weren’t directly modified in the PR or commit.

So I think we need to decide whether to keep it or remove it. What are your thoughts?

codecov.yml
```diff
@@ -16,16 +16,16 @@
...
   status:
-    patch: yes
+    patch: true
     project:
       default:
         threshold: 1%
         if_not_found: success
-        target: 2.x
-    changes: yes  #only care the changed files
+    changes: true
```
YongGoose: It looks like this setting was added to limit the comparison to only the modified files, but in reality, it behaves differently.
- https://github.com/apache/incubator-seata/pull/7350/files

It actually detects changes in code coverage even for lines that weren’t directly modified in the PR or commit.

So I think we need to decide whether to keep it or remove it. What are your thoughts?

funky-eyes: Since this prompt keeps appearing, I decided to delete the codecov/changes stage. I think we should revert it back to "no". I don't believe we need this stage as it doesn't seem to provide any positive value.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7356/commits/bf14d7483e96ad6453cd92cbe8162b47bd132b72 !

funky-eyes (approved):
LGTM

PR #7375, opened 2025-05-28
optimize: enhance close() logic of discovery module

YoWuwuuuw:

1. **Discovery module** uses thread pool in many registration center modules. If the shutdown logic is not explicitly invoked, resource leakage may occur. So, **shutdown** the thread pool in `close()` method.
2. Similarly, **explicitly close** the client in `close()` as best practice, If the client's lifecycle is controlled by the underlying layer, keep the original logic.
3. `RegistryHeartBeats.close()` is only called in **discovery-redis module**, adding the logic to shutdown `RegistryHeartBeats` for **discovery-consul module** and **discovery-etcd module**(use `RegistryHeartBeats` in the same way).


discovery/seata-discovery-etcd3/src/test/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImplMockTest.java
YongGoose, line 1: Personally, I think tests should ideally not depend on execution order.
Would you mind sharing the reason why you chose to specify the order in this case? 🤔

YoWuwuuuw: RegistryService is a single instance mode. When testing the whole life cycle, it is generally required to follow the order of register-> lookup (subscribe) -> unregister-> close. Otherwise, if the test method is executed concurrently or the order is not fixed, the required resources will be destroyed.

YoWuwuuuw: Or is there a better way to do it, and I do see lower efficiency in the test than in the normal test

YongGoose: Thanks for the explanation!

Since it’s implemented as a singleton and shares internal static fields, it does make sense that tests could affect each other.

While it might be possible to make each test fully isolated, I agree that using `@Order` seems like a more practical solution in this case.

I don’t think any changes are needed in this case!

YongGoose:
I’ll take a look at the more urgent PRs first, and then come back to review this one!
Also time to start paying attention to the first coding phase of GSoC. 😅

Let’s keep it up and make GSoC a success together! 🙌

YoWuwuuuw:
> 我会先看看那些比较紧急的 PR，然后再回来审核这个！ 现在也该开始关注 GSoC 的第一个编码阶段了。😅
> 
> 让我们继续努力，共同推动 GSoC 取得成功！🙌

Ok.😄
In fact, I'm pushing GSoC every day!👍

xingfudeshi (approved):
LGTM

PR #7346, opened 2025-05-19
optimize: replace tomcat with netty-http on the server

funky-eyes:
- [ ] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did
https://github.com/apache/incubator-seata/pull/7328
The original author no longer has the energy to maintain it, and I have taken over the maintenance after communicating with them.

todo:
1. Support for HttpServletRequest parameters is required, or alternatively, refactor the interfaces that depend on this parameter.
2. An independent HTTP business processing thread pool is needed, as the current performance of controllers running in the Netty IO thread is suboptimal.
3. The watch interface of Raft needs to be enhanced with support for HTTP/2.

PR #7385, opened 2025-05-29
optimize: remove HttpServletRequest from the watch API

funky-eyes:

The code from #7346 needs to be merged first before reviewing this PR. This PR addresses the first item in the TODO list.


xjlgod:
CI Test is not passed,  this PR is a bit similar to 7346?

YongGoose:
> CI Test is not passed, this PR is a bit similar to 7346?

It looks like the checkstyle check has failed.

```
Error:  /home/runner/work/incubator-seata/incubator-seata/server/src/main/java/org/apache/seata/server/controller/ClusterController.java:117:9: 'if' is not followed by whitespace. [WhitespaceAround]
Error:  /home/runner/work/incubator-seata/incubator-seata/server/src/main/java/org/apache/seata/server/controller/ClusterController.java:117:19: '==' is not followed by whitespace. [WhitespaceAround]
Error:  /home/runner/work/incubator-seata/incubator-seata/server/src/main/java/org/apache/seata/server/controller/ClusterController.java:117:19: '==' is not preceded with whitespace. [WhitespaceAround]
Error:  /home/runner/work/incubator-seata/incubator-seata/server/src/main/java/org/apache/seata/server/controller/ClusterController.java:117:26: '{' is not preceded with whitespace. [WhitespaceAround]
Error:  /home/runner/work/incubator-seata/incubator-seata/server/src/main/java/org/apache/seata/server/cluster/manager/ClusterWatcherManager.java:99: 'if' has incorrect indentation level 9, expected level should be 8. [Indentation]
```

YongGoose:
If it’s not urgent to merge right away, would it be okay if I leave a review around this evening?

funky-eyes:
> If it’s not urgent to merge right away, would it be okay if I leave a review around this evening?

ok

xjlgod (approved):
LGTM

core/src/main/java/org/apache/seata/core/rpc/netty/http/ParameterParser.java
```diff
@@ -56,51 +58,53 @@ public static ObjectNode convertParamMap(Map<String, List<String>> paramMap) thr
...
+            return httpContext;
         } else if (ParamMetaData.ParamConvertType.MODEL_ATTRIBUTE.equals(paramConvertType)) {
             JsonNode param = paramMap.get("param");
             return OBJECT_MAPPER.convertValue(param, parameterType);
         } else if (ParamMetaData.ParamConvertType.REQUEST_BODY.equals(paramConvertType)) {
             JsonNode body = paramMap.get("body");
             return OBJECT_MAPPER.convertValue(body, parameterType);
         } else {
-            return paramMap.get("param").get(parameterName).asText(null);
+            JsonNode jsonNode = paramMap.get("param").get(parameterName);
```
YongGoose: If paramMap.get("param") returns null, it will cause a `NullPointerException`.

server/src/main/java/org/apache/seata/server/cluster/manager/ClusterWatcherManager.java
```diff
@@ -81,19 +83,38 @@ public void onChangeEvent(ClusterChangeEvent event) {
...
+                    ctx.writeAndFlush(response);
+                }
+            } else {
+                logger.warn("Netty channel is not active for watcher on group {}, cannot send response.",
+                    watcher.getGroup());
+            }
+        } else {
+            logger.warn("Unsupported context type for watcher on group {}: {}", watcher.getGroup(),
+                context != null ? context.getClass().getName() : "null");
         }
```
YongGoose: How about adding a return at the end of the if block to reduce the depth?

server/src/main/java/org/apache/seata/server/controller/ClusterController.java
```diff
@@ -123,13 +108,16 @@ public MetadataResponse cluster(String group) {
...
 
     @PostMapping("/watch")
-    public void watch(HttpServletRequest request, @RequestParam Map<String, Object> groupTerms,
-        @RequestParam(defaultValue = "28000") int timeout) {
-        AsyncContext context = request.startAsync();
-        context.setTimeout(0L);
+    public void watch(HttpContext context, @RequestBody Map<String, Object> groupTerms,
+        @RequestParam(defaultValue = "28000") Integer timeout) {
+        context.setAsync(true);
+        if (timeout == null) {
```
YongGoose, line 114: I think this code can be removed.

It would be good to add a test to ensure that defaultValue is correctly applied when null is passed, in case the behavior changes of the Spring Framework version.

If you feel that this test is unnecessary, feel free to leave it out.

image

YongGoose (commented):
It would be great if you could add both unit and integration tests.

server/src/main/java/org/apache/seata/server/controller/ClusterController.java
funky-eyes: > I think this code can be removed.
> 
> It would be good to add a test to ensure that defaultValue is correctly applied when null is passed, in case the behavior changes of the Spring Framework version.
> 
> If you feel that this test is unnecessary, feel free to leave it out.
> 
> image

https://github.com/apache/incubator-seata/issues/7407 We are not using spring-web features; these annotations are only for compatibility with previous spring-web usage patterns. Therefore, what's described in Spring's documentation does not apply to this code.

server/src/main/java/org/apache/seata/server/cluster/manager/ClusterWatcherManager.java
funky-eyes: done

core/src/main/java/org/apache/seata/core/rpc/netty/http/ParameterParser.java
funky-eyes: done

server/src/test/java/org/apache/seata/server/raft/RaftServerTest.java
```diff
@@ -92,4 +93,11 @@ public void initRaftServerFailByRaftPortNull() {
         Assertions.assertThrows(IllegalArgumentException.class, RaftServerManager::init);
     }
 
+    @AfterAll
+    public static void afterAll() {
+        LockerManagerFactory.destroy();
+        SessionHolder.destroy();
+        RaftServerManager.destroy();
```
YongGoose: Wouldn’t destory the `RaftServerManager` in `@AfterAll` potentially cause test failures if the test execution order changes?

funky-eyes: Not executing the destroy method actually affects other unit tests; each unit test should reset the modifications when it finishes. The exceptions caused by previous changes in test sequence happened because Seata's SPI doesn't support environment isolation. The Dubbo community solved this issue by allowing each unit test to have its own environment where there is a singleton instance, avoiding the mutual interference caused by globally sharing a single instance.

YongGoose: I agree with you about avoiding side effects in unit tests.
However, what I meant was not to use `@AfterAll`, but rather to use `@AfterEach` in order to improve isolation between test methods within the same class.

Sorry for the confusion — I realize I didn’t explain it clearly earlier! 😅

YongGoose: There was actually a similar issue recently.
- https://github.com/apache/incubator-seata/pull/7397/files

YongGoose:
> Error:  Errors: 
Error:    ClusterControllerTest.watch:85 » HttpHostConnect Connect to 127.0.0.1:8091 [/1...
Error:    ClusterControllerTest.watchTimeoutTest:55 » HttpHostConnect Connect to 127.0.0...
[INFO] 
Error:  Tests run: 173, Failures: 0, Errors: 2, Skipped: 12

To send actual HTTP requests, you’ll likely need to set the webEnvironment property in `@SpringBootTest`.
> If this doesn’t turn out to be the solution, I’ll look into it again later this afternoon. 😊

- https://docs.spring.io/spring-boot/reference/testing/spring-boot-applications.html
image

funky-eyes:
> > Error:  Errors:
> > Error:    ClusterControllerTest.watch:85 » HttpHostConnect Connect to 127.0.0.1:8091 [/1...
> > Error:    ClusterControllerTest.watchTimeoutTest:55 » HttpHostConnect Connect to 127.0.0...
> > [INFO]
> > Error:  Tests run: 173, Failures: 0, Errors: 2, Skipped: 12
> 
> To send actual HTTP requests, you’ll likely need to set the webEnvironment property in `@SpringBootTest`.
> 
> > If this doesn’t turn out to be the solution, I’ll look into it again later this afternoon. 😊
> 
> * https://docs.spring.io/spring-boot/reference/testing/spring-boot-applications.html
>   image

We don't have a WebApplicationContext because seata-server no longer uses spring-web features. HTTP requests are actually sent to the transaction port 8091, which works in my local testing.
image

YongGoose:
> We don't have a WebApplicationContext because seata-server no longer uses spring-web features. HTTP requests are actually sent to the transaction port 8091, which works in my local testing. 

I also ran it locally and it worked fine 🤨
Could it possibly be a port conflict issue?

YongGoose:
@funky-eyes 

It looks like the original issue has been resolved.

```
Error:  Failures: 
Error:    ClusterControllerTest.watchTimeoutTest:61 expected: <304> but was: <200>
```

What do you think?

funky-eyes:
> @funky-eyes
> 
> It looks like the original issue has been resolved.
> 
> ```
> Error:  Failures: 
> Error:    ClusterControllerTest.watchTimeoutTest:61 expected: <304> but was: <200>
> ```
> 
> What do you think?


It looks like the order wasn't set, which caused unit tests that should return 304 to run together with unit tests that should return 200.

funky-eyes:
> > We don't have a WebApplicationContext because seata-server no longer uses spring-web features. HTTP requests are actually sent to the transaction port 8091, which works in my local testing.
> 
> I also ran it locally and it worked fine 🤨 Could it possibly be a port conflict issue?

I think this issue is caused by port conflicts. This port might have been modified by other test cases and is no longer 8091, so I created a separate configuration for this test case, specifying the port as 18091 to avoid conflicts.

YongGoose (approved):
LGTM 👍🏻

Issue #7379, opened 2025-05-28
add UT for TccAnnotationProcessor class


slievrly:
https://app.codecov.io/gh/apache/incubator-seata/blob/2.x/spring%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fseata%2Fspring%2Ftcc%2FTccAnnotationProcessor.java

OmCheeLin:
Please assign it to me, I would like to work on this.

slievrly:
@OmCheeLin Looking forward to your PR.

PR #7398, opened 2025-05-31
test: add UT for TccAnnotationProcessor class

OmCheeLin:

add UT for TccAnnotationProcessor class

test #7379

### DIFF
--- a/spring/pom.xml
+++ b/spring/pom.xml
@@ -31,6 +31,10 @@
     <name>seata-spring ${project.version}</name>
     <description>spring for Seata built with Maven</description>
 
+    <properties>
+        <logcaptor.version>2.11.0</logcaptor.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
@@ -79,6 +83,12 @@
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
             <artifactId>kotlin-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
+            <version>${logcaptor.version}</version>
             <scope>test</scope>
         </dependency>
 
--- a/spring/pom.xml
+++ b/spring/pom.xml
@@ -31,10 +31,6 @@
     <name>seata-spring ${project.version}</name>
     <description>spring for Seata built with Maven</description>
 
-    <properties>
-        <logcaptor.version>2.11.0</logcaptor.version>
-    </properties>
-
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
@@ -88,7 +84,6 @@
         <dependency>
             <groupId>io.github.hakky54</groupId>
             <artifactId>logcaptor</artifactId>
-            <version>${logcaptor.version}</version>
             <scope>test</scope>
         </dependency>
 
--- a/spring/pom.xml
+++ b/spring/pom.xml
@@ -82,9 +82,17 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>io.github.hakky54</groupId>
-            <artifactId>logcaptor</artifactId>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


spring/src/test/java/org/apache/seata/spring/tcc/TccAnnotationProcessorTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        Object newValue = field.get(bean);
+        assertNotEquals(originalValue, newValue, "Proxy should replace original field");
+    }
+
+    @Test
+    public void testAddTccAdviseFieldValueNull() throws Exception {
+        TestBean bean = new TestBean();
+        Field nullField = TestBean.class.getField("nullService");
+        // 应该安全返回，不抛异常
```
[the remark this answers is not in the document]
YongGoose: Agree.

YongGoose:
Please take a look :)

> when submitting a PR, please add a brief description of the current changes to the X.X.X.md file under the 'changes/' folder

- https://github.com/apache/incubator-seata/blob/2.x/CONTRIBUTING.md#commit-content

spring/src/test/java/org/apache/seata/spring/tcc/TccAnnotationProcessorTest.java
```diff
@@ -0,0 +1,171 @@
...
+                            return Mockito.spy(new MockTccService());
+                        }
+                        return arg;
+                    });
+
+            processor.addTccAdvise(bean, "testBean", field, MockTccService.class);
+        }
+
+        Object newValue = field.get(bean);
+        assertNotEquals(originalValue, newValue, "Proxy should replace original field");
```
YongGoose, line 114: I think it might also be helpful to check whether the logs are being printed correctly.
WDYT?

spring/src/test/java/org/apache/seata/spring/tcc/TccAnnotationProcessorTest.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Test
+    public void testAddTccAdviseFieldValueNull() throws Exception {
+        TestBean bean = new TestBean();
+        Field nullField = TestBean.class.getField("nullService");
+        // 应该安全返回，不抛异常
+        processor.addTccAdvise(bean, "testBean", nullField, MockTccService.class);
+        assertNull(nullField.get(bean));
```
YongGoose, line 126: Since fieldValue is null, I assume the log won’t be printed.
It might also be a good idea to add a test to verify that the log is not printed in this case!!

spring/src/test/java/org/apache/seata/spring/tcc/TccAnnotationProcessorTest.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    @Test
+    public void testProcessWithNullAnnotation() throws Exception {
+        processor.process(new TestBean(), "testBean", null);
+        Field proxiedField = TccAnnotationProcessor.class.getDeclaredField("PROXIED_SET");
+        proxiedField.setAccessible(true);
+        Set<String> proxied = (Set<String>) proxiedField.get(null);
```
YongGoose: What do you think about extracting proxied as a global variable?

We could assign it in a `@BeforeEach` and reset it in an `@AfterEach` — that might help improve clarity and maintainability.

YongGoose (commented):
Great work on the test code.
I’ve added a few comments for your consideration.

spring/pom.xml
```diff
@@ -81,6 +85,12 @@
             <artifactId>kotlin-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.github.hakky54</groupId>
+            <artifactId>logcaptor</artifactId>
+            <version>${logcaptor.version}</version>
+            <scope>test</scope>
```
YongGoose: I think it would be better to standardize the libraries we use.
- https://github.com/apache/incubator-seata/pull/7320

image

YongGoose (commented):
We're almost there!
Left comment. please take a look!

spring/pom.xml
OmCheeLin: Do you mean to define {logcaptor. version} in dependencies/pom.xml ?

spring/src/test/java/org/apache/seata/spring/tcc/TccAnnotationProcessorTest.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @Test
+    public void testAddTccAdviseFieldValueNull() throws Exception {
+        TestBean bean = new TestBean();
+        Field nullField = TestBean.class.getField("nullService");
+        processor.addTccAdvise(bean, "testBean", nullField, MockTccService.class);
+        assertNull(nullField.get(bean));
+        boolean hasLogs = listAppender.list.stream()
+                .anyMatch(event -> event.getLevel() == Level.INFO);
```
YongGoose: It would be better to verify that the log was not printed, rather than assuming that there is no log at the INFO level.

Also, please consider using noneMatch for this check.

funky-eyes (approved):
LGTM

PR #7400, opened 2025-06-01
test: add test for SpringProxyUtils and OrderUtils

### LINKED ISSUE #7380: add UT for org.apache.seata.spring.util package
https://app.codecov.io/gh/apache/incubator-seata/tree/2.x/spring%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fseata%2Fspring%2Futil

Oscarcheng0312:

Added comprehensive unit tests for SpringProxyUtils and supplemented missing test cases in OrderUtilTest. Improved overall test coverage for the org.apache.seata.spring.util package to 91%. Increase test coverage for OrderUtil to 100%. Created class SpringProxyUtilsTest with test coverage of 81%.
 
fixed #7380 

This PR adds missing unit tests.

Run SpringProxyUtilsTest and OrderUtilTest, and check the coverage report. All new test cases should pass.

### DIFF
--- a/spring/src/test/java/org/apache/seata/spring/util/OrderUtilTest.java
+++ b/spring/src/test/java/org/apache/seata/spring/util/OrderUtilTest.java
@@ -25,7 +25,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * The type OrderUtil test
- *
  */
 public class OrderUtilTest {
 
@@ -65,6 +64,20 @@ public class OrderUtilTest {
         assertThat(OrderUtil.lowerOrEquals(advisor21, advisor11)).isTrue();
         assertThat(OrderUtil.lowerOrEquals(advisor12, advisor11)).isTrue();
         assertThat(OrderUtil.lowerOrEquals(advisor22, advisor21)).isTrue();
+
+        // Case: source is null -> treated as Integer.MAX_VALUE, so MAX > 0 => should return true
+        assertThat(OrderUtil.lowerThan(null, 0)).isTrue();
+        // Case: target is null -> treated as Integer.MAX_VALUE, so 0 < MAX => should return false
+        assertThat(OrderUtil.lowerThan(0, null)).isFalse();
+        // Case: both null -> both treated as MAX => not lower than
+        assertThat(OrderUtil.lowerThan((Integer) null, null)).isFalse();
+    }
+
+    @Test
+    public void test_lowerOrEquals_nullCases() {
+        assertThat(OrderUtil.lowerOrEquals(null, 0)).isTrue(); // MAX >= 0
+        assertThat(OrderUtil.lowerOrEquals(0, null)).isFalse(); // 0 >= MAX → false
+        assertThat(OrderUtil.lowerOrEquals((Integer) null, null)).isTrue(); // MAX >= MAX
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


slievrly:
@Oscarcheng0312 Pls add ASF license header.

YongGoose:
I believe that unit tests should be kept as small and focused as possible!
That way, you don’t need to explain each case in comments — the test method names themselves will clearly indicate what each case is testing.

The following document might be helpful.
- https://www.browserstack.com/guide/how-to-write-junit-test-cases

image

Oscarcheng0312:
> @Oscarcheng0312 Pls add ASF license header.

OK, I will.

slievrly:
> I believe that unit tests should be kept as small and focused as possible! That way, you don’t need to explain each case in comments — the test method names themselves will clearly indicate what each case is testing.
> 
> The following document might be helpful.
> 
> * https://www.browserstack.com/guide/how-to-write-junit-test-cases
> 
> image

@Oscarcheng0312 This is a good suggestion. What do you think?

Oscarcheng0312:
> > I believe that unit tests should be kept as small and focused as possible! That way, you don’t need to explain each case in comments — the test method names themselves will clearly indicate what each case is testing.
> > The following document might be helpful.
> > 
> > * https://www.browserstack.com/guide/how-to-write-junit-test-cases
> > 
> > image
> 
> @Oscarcheng0312 This is a good suggestion. What do you think?

I totally agree, you can check my second modification in this PR, I not only added the ASF header, but also removed most of the comments.

slievrly:
@Oscarcheng0312 Pls add change  in the following file.

spring/src/test/java/org/apache/seata/spring/util/SpringProxyUtilsTest.java
```diff
@@ -0,0 +1,195 @@
...
+                (proxy, method, args) -> null
+        );
+        assertEquals(jdkProxy.getClass(), SpringProxyUtils.findTargetClass(jdkProxy));
+
+        ProxyFactory factory = new ProxyFactory(new TestServiceImpl());
+        factory.addInterface(TestService.class);
+        TestService springJdkProxy = (TestService) factory.getProxy();
+        assertTrue(AopUtils.isAopProxy(springJdkProxy) && AopUtils.isJdkDynamicProxy(springJdkProxy));
+        assertEquals(TestServiceImpl.class, SpringProxyUtils.findTargetClass(springJdkProxy));
+    }
```
YongGoose: What I meant was not simply removing comments, but splitting the tests by case.

For example, in this test, there are cases for `null`, `plain`, `proxy`, and `proxy created by proxyFactory`.
Instead of handling all these cases within a single test, they should be covered in four separate tests.

Oscarcheng0312: Thanks for your suggestions, I modified the test class, splitting the tests by case.

spring/src/test/java/org/apache/seata/spring/util/OrderUtilTest.java
YongGoose, line 1: It looks like the OrderUtilTest class should also be refactored in the same way.
You’re welcome to include it in this PR, or handle it separately in another PR—whichever you prefer.

YongGoose (approved):
LGTM

spring/src/test/java/org/apache/seata/spring/util/SpringProxyUtilsTest.java
```diff
@@ -0,0 +1,302 @@
...
+        MultiInterfaceImpl multi = new MultiInterfaceImpl();
+        Class<?>[] ifaces = SpringProxyUtils.getAllInterfaces(multi);
+        assertNotNull(ifaces);
+        assertEquals(2, ifaces.length);
+        assertTrue(Arrays.asList(ifaces).contains(TestService.class));
+        assertTrue(Arrays.asList(ifaces).contains(Runnable.class));
+    }
+
+
+    // Tests for getTargetClass
```
YongGoose, line 238: [Nit] In cases like this, using `@Nested` could be a good option.

spring/src/test/java/org/apache/seata/spring/util/OrderUtilTest.java
Oscarcheng0312: I prefer to handle it in another PR.

slievrly (approved):
LGTM

PR #7397, opened 2025-05-31
bugfix: Resolve NullPointer and port binding errors

YongGoose:
This pull request includes several updates to improve test reliability, fix bugs, and update documentation for the `2.x` branch. Key changes include resolving NullPointer and port binding errors, improving test setup and teardown processes, and adding missing mock configurations.

### Bug Fixes:
* Added a new entry in the English and Chinese `2.x.md` changelogs to document the resolution of NullPointer and port binding errors (`#7397`). [[1]](diffhunk://#diff-ad044a1f48999a633bcdb6209a68785b1dafe98ce0f884346f87a7056e2cfb75L32-R32) [[2]](diffhunk://#diff-150bf4cbc3fa9dcea0e61fd42e7dc4e56b65f3f1c0c8eb1d2415b3a94d22624dR32-R33)

### Test Improvements:
* Updated `ChannelEventHandlerIntegrationTest` to use a different server port (`8919`) to avoid potential conflicts.
* Enhanced `ChannelEventListenerTest` by adding a mock for `ChannelId` and using `when` to simulate `channel.id()` behavior. [[1]](diffhunk://#diff-1a65aed251f068b700d616f8a064d51554d34f89ff016cb0a7784751a370a774R20) [[2]](diffhunk://#diff-1a65aed251f068b700d616f8a064d51554d34f89ff016cb0a7784751a370a774R31) [[3]](diffhunk://#diff-1a65aed251f068b700d616f8a064d51554d34f89ff016cb0a7784751a370a774R41-R43) [[4]](diffhunk://#diff-1a65aed251f068b700d616f8a064d51554d34f89ff016cb0a7784751a370a774R87) [[5]](diffhunk://#diff-1a65aed251f068b700d616f8a064d51554d34f89ff016cb0a7784751a370a774R98)
[...]

PR #7401, opened 2025-06-01
bugfix: fix port binding

OmCheeLin:

Resolve potential port conflict issues in `ChannelEventHandlerIntegrationTest.java`

### DIFF
--- a/core/src/test/java/org/apache/seata/core/rpc/netty/ChannelEventHandlerIntegrationTest.java
+++ b/core/src/test/java/org/apache/seata/core/rpc/netty/ChannelEventHandlerIntegrationTest.java
@@ -35,7 +35,10 @@ import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.GlobalEventExecutor;
+
+import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.net.ServerSocket;
 import java.net.SocketAddress;
 import java.util.Iterator;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


funky-eyes (commented):
The problem still exists.

YongGoose:
It would be great to first resolve the issue in this PR, and then gradually improve things by applying this approach step by step.
- https://github.com/apache/incubator-seata/pull/7397

I’d appreciate it if you could explain in the issue how this approach addresses the current problem.
Also, it would be helpful to include how it handles `flaky tests`, `race conditions`, or cases where a port is already in use.

OmCheeLin:
> It would be great to first resolve the issue in this PR, and then gradually improve things by applying this approach step by step.
> 
> * [bugfix: Resolve NullPointer and port binding errors #7397](https://github.com/apache/incubator-seata/pull/7397)
> 
> I’d appreciate it if you could explain in the issue how this approach addresses the current problem. Also, it would be helpful to include how it handles `flaky tests`, `race conditions`, or cases where a port is already in use.

There is currently an issue with `FileSessionManagerTest.java`, 
I see that you have solved both `ChannelEventHandlerIntegrationTest.java` and `ChannelEventListenerTest.java`, but I think there may be better ways to modify these two files. You can refer to my file changes.

YongGoose:
> There is currently an issue with `FileSessionManagerTest.java`, I see that you have solved both `ChannelEventHandlerIntegrationTest.java` and `ChannelEventListenerTest.java`, but I think there may be better ways to modify these two files. You can refer to my file changes.

Your approach looks promising!

We’re currently experiencing intermittent `BindExceptions` in our tests.
If this method can effectively address the issues I mentioned in my previous comment, I think it would be a great idea to refactor all the test cases that currently use statically assigned ports.

That’s why I suggested starting a discussion with the community through an issue — to explore this direction further!

Have a great day :)

YongGoose:
> > There is currently an issue with `FileSessionManagerTest.java`, I see that you have solved both `ChannelEventHandlerIntegrationTest.java` and `ChannelEventListenerTest.java`, but I think there may be better ways to modify these two files. You can refer to my file changes.
> 
> Your approach looks promising!
> 
> We’re currently experiencing intermittent `BindExceptions` in our tests. If this method can effectively address the issues I mentioned in my previous comment, I think it would be a great idea to refactor all the test cases that currently use statically assigned ports.
> 
> That’s why I suggested starting a discussion with the community through an issue — to explore this direction further!
> 
> Have a great day :)

@OmCheeLin

What do you think?

OmCheeLin:
@YongGoose 
You have already solved the problem, that's great. Perhaps we can wait until there are problems in the next test to optimize it.

PR #7312, opened 2025-04-24
optimize: spotless skip

jsbxyyx:
- [ ] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did

optimize spotless skip

PR #7408, opened 2025-06-04
optimize: Apply Spotless to the entire codebase.

YongGoose:


YongGoose:
@funky-eyes 

I’m wondering if there are more areas where this type of work is needed.
- https://github.com/apache/incubator-seata/pull/7312

YongGoose:
You can see how the AOSP style is applied in this PR
- https://github.com/YongGoose/incubator-seata/pull/6

You can see how the Palantier style is applied in this PR
- https://github.com/YongGoose/incubator-seata/pull/7

Currently, both styles are failing `Checkstyle` validation.
One alternative would be to remove the `Checkstyle` dependency and use `Spotless / Check` for style validation instead.

I’d appreciate it if you could let me know which of the two styles you think is better.
Also, if there are any options that should be added, please don’t hesitate to let me know.

funky-eyes:
> You can see how the AOSP style is applied in this PR
> 
> * [Google java format (AOSP style)  YongGoose/incubator-seata#6](https://github.com/YongGoose/incubator-seata/pull/6)
> 
> You can see how the Palantier style is applied in this PR
> 
> * [palantier format YongGoose/incubator-seata#7](https://github.com/YongGoose/incubator-seata/pull/7)
> 
> Currently, both styles are failing `Checkstyle` validation. One alternative would be to remove the `Checkstyle` dependency and use `Spotless / Check` for style validation instead.
> 
> I’d appreciate it if you could let me know which of the two styles you think is better. Also, if there are any options that should be added, please don’t hesitate to let me know.

Can you tell me which one is currently the most popular?

funky-eyes:
I agree with removing the current Checkstyle and then using Spotless/Check instead.

YongGoose:
> Can you tell me which one is currently the most popular?

Google seems to be the most popular.

However, since Palantir addresses some of the shortcomings of the Google formatter, I personally prefer Palantir.

funky-eyes:
> > Can you tell me which one is currently the most popular?
> 
> Google seems to be the most popular.
> 
> However, since Palantir addresses some of the shortcomings of the Google formatter, I personally prefer Palantir.

Perhaps we should send an email to introduce the differences between the two options, and then conduct a vote with a 3-day deadline to decide on the final choice.

YongGoose:
> Perhaps we should send an email to introduce the differences between the two options, and then conduct a vote with a 3-day deadline to decide on the final choice.

I’ll send the email by tomorrow and notice on Dingtalk!

funky-eyes:
The voting should have already ended. We should continue with the work related to this PR.

YongGoose:
> The voting should have already ended. We should continue with the work related to this PR.

I’ll likely be able to work on it later tonight.
Once I’ve completed the work, I’ll request a review!

YongGoose:
Our workflow requires running the code formatter with Java 8. However, the current version of the Spotless plugin doesn't support the palantirJavaFormat configuration when using Java 8.

We should also reconsider using `Checkstyle`. 
If we go that route, it might take some time since we'll need to adjust the `Spotless` configuration accordingly.

jsbxyyx:
add importOrder
```
<importOrder>
    <order>,javax|java,\#</order>
</importOrder>
```

this is idea default import style

funky-eyes:
> Our workflow requires running the code formatter with Java 8. However, the current version of the Spotless plugin doesn't support the palantirJavaFormat configuration when using Java 8.
> 
> We should also reconsider using `Checkstyle`. If we go that route, it might take some time since we'll need to adjust the `Spotless` configuration accordingly.

We can disable code style checks on JDK 8 and instead perform code style checks on JDK 17, or have a dedicated pipeline specifically for code style checks.

YongGoose:
> > Our workflow requires running the code formatter with Java 8. However, the current version of the Spotless plugin doesn't support the palantirJavaFormat configuration when using Java 8.
> > We should also reconsider using `Checkstyle`. If we go that route, it might take some time since we'll need to adjust the `Spotless` configuration accordingly.
> 
> We can disable code style checks on JDK 8 and instead perform code style checks on JDK 17, or have a dedicated pipeline specifically for code style checks.

Splitting it sounds like a good idea.

Since there are many file conflicts at the moment, I’ll close this PR and open a new one.
I’ve been quite busy lately, so I’ll likely be able to work on it around the end of this week. 🥲

PR #7421, opened 2025-06-07
optimize: Add pmd-check reports to the workflow

YoWuwuuuw:

#### Cause: 
When contributing code, even if the pmd-check plugin is successfully checked in a local win or mac environment, it may fail in the GitHub Actions ci workflow and no useful information can be found in the log.
#### Solve:
Therefore, add pmd-check reports to the workflow so that developers can clearly see why pmd is false, even if it cannot be reproduced locally.
#### Performance loss:
Less than a second

#### **pmd-check success scenario:**
url: https://github.com/YoWuwuuuw/incubator-seata/actions/runs/15506243350/job/43661466222
image

#### **pmd-check false scenario:**
url: https://github.com/YoWuwuuuw/incubator-seata/actions/runs/15506208318/job/43661597897
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,11 +71,11 @@ jobs:
         uses: actions/cache/restore@v4
         id: cache-maven-repository
         with:
-         path: ~/.m2/repository
-         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-${{ env.TODAY }}
-         restore-keys: |
-           ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-           ${{ runner.os }}-maven-
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-${{ env.TODAY }}
+          restore-keys: |
+            ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+            ${{ runner.os }}-maven-
       # step 6.1
       - name: "Test, Check style, Check PMD, Check license with Maven and Java8"
         if: matrix.java == '8'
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -71,11 +71,11 @@ jobs:
...
-         path: ~/.m2/repository
-         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-${{ env.TODAY }}
-         restore-keys: |
-           ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-           ${{ runner.os }}-maven-
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}-${{ env.TODAY }}
+          restore-keys: |
+            ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+            ${{ runner.os }}-maven-
```
YongGoose, line 78: Just checking — is the indentation intentional?

YoWuwuuuw: > Just checking — is the indentation intentional?

It's just a uniform format, these lines are indented differently than the overall format

YongGoose (approved):
Nice work :)

.github/workflows/build.yml
```diff
@@ -84,6 +84,32 @@ jobs:
...
+            if grep -q "<violation" "$report_file"; then
+              FOUND_VIOLATIONS=true
+              echo "--- PMD XML Report with Violations for: $report_file ---"
+              cat "$report_file"
+              echo "--------------------------------------------------------"
+              echo ""
+            fi
+          done
+
+          if [ "$FOUND_VIOLATIONS" = false ]; then
```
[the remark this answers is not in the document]
YoWuwuuuw: It doesn't matter, but it makes sense

.github/workflows/build.yml
```diff
@@ -84,6 +84,32 @@ jobs:
                  -Dcheckstyle.skip=false -Dpmd.skip=false -Dlicense.skip=false -DredisCaseEnabled=true \
                  -e -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn;
       # step 6.2
+      - name: "Print PMD Violation XML Details"
+        if: matrix.java == '8' && always()
+        run: |
+          echo "--- Searching for and printing PMD XML reports with violations across all modules ---"
```
slievrly, line 90: all modules? When PMD check encounters the first module that does not meet expectations, it will fail on error and print the error message. And the remaining modules will not continue to be executed.

slievrly: https://github.com/apache/incubator-seata/blob/2.x/pom.xml#L419-L462

YoWuwuuuw: > all modules? When PMD check encounters the first module that does not meet expectations, it will fail on error and print the error message. And the remaining modules will not continue to be executed.

Yes, the loop here is to check if the pmd.xml in each target directory is empty, and the final print is the pmd report of the failed module. 

If all modules are empty, it will prompt that the PMD test is successful.

slievrly: The current PMd check follows a fail-fast approach. Suppose a project has 4 modules. In a single PR submission, if modules 2, 3, and 4 all introduce PMD issues, the workflow will only detect and report issues in Module 2, while Modules 3 and 4 will not be processed or analyzed.

YoWuwuuuw: My understanding is that when the previous module fails to build, the next module will not generate the target directory, so the traversal ends (only the first PMD will fail to print)

YoWuwuuuw: I understand, we should modify the prompt words to remind developers that the pmd failure may not be the only one

slievrly: If the workflow only outputs the first failing module's PMD issues instead of all modules', what is the practical value of this PR improvement? Without this improvement, the original workflow still logs the first failing module's PMD issues in the log.

YoWuwuuuw: I remember the problem was that I couldn't find any PMD failure information in the log. I will verify tomorrow.

slievrly: You can import a useless import in the "import" of the java source code for testing.

YoWuwuuuw: This is indeed my problem. By downloading the log and filtering the keywords, it can be found that the log is printed as INFO level!
image

I think we can optimize the mvn command to make the pmd failure log ERROR level, as follows:
```
run: |
          ./mvnw -T 4C clean test \
                 -Dcheckstyle.skip=false -Dpmd.skip=false -Dlicense.skip=false -DredisCaseEnabled=true \
                 -e -B \
                 -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
                 -Dorg.slf4j.simpleLogger.log.net.sourceforge.pmd=error;
```
image


What do you think @slievrly 
Finally, this pr will be closed.

slievrly: I agree with you. go ahead.

PR #7428, opened 2025-06-10
optimize: pmd-check log as ERROR level

YoWuwuuuw:

If the build fails, pmd-check will only output logs at the info level, which is difficult to see in the large number of logs generated by the ci build process. 
Therefore, it is changed to the error level and highlighted to make it easier for developers to view

Based on: https://github.com/apache/incubator-seata/pull/7421#discussion_r2135050547


slievrly (approved):
LGTM

PR #7443, opened 2025-06-16
optimize: Replace @LocalTCC with @SagaTransactional in the saga annotation pattern

### LINKED ISSUE #7439: Remove the dependency of Saga annotations on @LocalTCC
### Why you need it?

1. 如果saga注解化的分支不是远程分支时,必须要带上`@LocalTCC`注解才会被代理,才能注册分支事务.而该注解在使用saga注解化场景下使用会非常困惑,因为它是saga,而不是tcc模式.
2. 你可以实现一个新的注解来替代`@LocalTCC`或者为TCC和SAGA注解化提供一个通用的注解,而不是带有某个特定事务模式标识的注解.
注: 需要向下兼容`@LocalTCC`的使用方式

- When the branch in Saga annotation mode is not a remote branch, it must be annotated with `@LocalTCC` to be proxied and to register the branch transaction. However, using this annotation in the Saga annotation scenario can be very confusing, since it is a Saga, not a TCC mode.

- You could implement a new annotation to replace `@LocalTCC`, or provide a general annotation for both TCC and Saga annotation modes, instead of using an annotation that indicates a specific transaction pattern.

Note: The new solution should be backward compatible with the current usage of `@LocalTCC`.

### How it could be?

_No response_

### Other related information

_No response_

WangzJi:

This PR implements a dedicated `@SagaTransactional` annotation to decouple Saga transaction mode from the TCC-specific `@LocalTCC` annotation, addressing the semantic confusion in Saga scenarios.

**Key Changes:**

1. **New `@SagaTransactional` Annotation**
   - Created in `saga-annotation` module for semantic clarity
   - Designed specifically for Saga transaction participants
   - Supports both interface and implementation class annotation
   - Includes comprehensive JavaDoc with usage examples

2. **Dedicated `SagaTransactionalRemotingParser`**
   - Independent parser handling only `@SagaTransactional` annotations
   - Registered separately via SPI mechanism
   - Optimized for Saga-specific processing logic

3. **SPI Integration**
   - Added `META-INF/services/org.apache.seata.integration.tx.api.remoting.RemotingParser`
   - Enables automatic discovery and registration of the Saga parser

4. **Enhanced Documentation**
   - Detailed usage patterns and best practices
   - Clear migration guidance for existing users
   - Comprehensive API documentation

5. **Backward Compatibility**
   - All existing `@LocalTCC` functionality remains unchanged
   - Both annotations can coexist in the same application
   - Zero breaking changes for existing implementations

Closes #7439

1. Run the test suite:
```bash
./mvnw test -Dtest=SagaTransactionalTest,SagaTransactionalRemotingParserTest -pl saga/seata-saga-annotation
```

2. Verify backward compatibility:
- Existing `@LocalTCC` services continue to work unchanged
- New services can use `@SagaTransactional` for better semantic clarity
- Both annotations can coexist without conflicts

1. Backward Compatibility:
   - Maintains 100% compatibility with existing @LocalTCC usage
   - No breaking changes to existing services

2. Code Quality:
   - All tests pass
   - Follows project coding standards
 
3. Documentation:
   - Added clear documentation for @LocalTransactional usage
   - Updated migration guide for existing users


compatible/src/main/java/io/seata/rm/tcc/remoting/parser/LocalTCCRemotingParser.java
funky-eyes, line 1: The new features do not necessitate any changes to the compatible module. Its sole responsibility is to maintain compatibility with Seata versions prior to 2.1.

common/src/main/java/org/apache/seata/common/transaction/api/TransactionParticipant.java
```diff
@@ -0,0 +1,41 @@
...
+ * In distributed transaction terminology, a "participant" is any component that takes part
+ * in the transaction protocol, making this name semantically accurate across different
+ * transaction patterns (TCC, Saga, XA, etc.).
+ * 
+ * @see org.apache.seata.rm.tcc.api.LocalTCC // the original TCC-specific annotation, still supported for backward compatibility
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface TransactionParticipant {
```
funky-eyes: Firstly, it is only applied in the TCC and SAGA annotation modes. Secondly, this naming is rather ambiguous and doesn't clearly relate to its functionality.

WangzJi: @funky-eyes, thanks for your review! I'd like to address your concerns:

**Regarding " it is only applied in the TCC and SAGA annotation modes":**

The  annotation aims to solve this limitation by being more generic. Here are the location options:

**Option A: Current location** `common/transaction/api`
- ✅ Generic design - not limited to specific transaction modes  
- ✅ Future extensibility for other transaction patterns
- ✅ Resolves the scope limitation you mentioned

**Option B: More focused location** `integration-tx-api/annotation`
- ✅ Co-located with other transaction annotations like `AspectTransactional`
- ✅ More focused on annotation-based transaction integration
- ✅ Still maintains proper module dependencies

**Regarding the naming:** 
How about `LocalTransactionResource` provides clearer semantics than `TransactionParticipant`:
- "Local" - clearly indicates it's for local (non-remote) resources
- "TransactionResource" - directly relates to its functionality
- Avoids the ambiguity you pointed out

Which approach do you prefer? Or do you have other suggestions for the design?

funky-eyes: > @funky-eyes, thanks for your review! I'd like to address your concerns:
> 
> **Regarding " it is only applied in the TCC and SAGA annotation modes":**
> 
> The annotation aims to solve this limitation by being more generic. Here are the location options:
> 
> **Option A: Current location** `common/transaction/api`
> 
> * ✅ Generic design - not limited to specific transaction modes
> * ✅ Future extensibility for other transaction patterns
> * ✅ Resolves the scope limitation you mentioned
> 
> **Option B: More focused location** `integration-tx-api/annotation`
> 
> * ✅ Co-located with other transaction annotations like `AspectTransactional`
> * ✅ More focused on annotation-based transaction integration
> * ✅ Still maintains proper module dependencies
> 
> **Regarding the naming:** How about `LocalTransactionResource` provides clearer semantics than `TransactionParticipant`:
> 
> * "Local" - clearly indicates it's for local (non-remote) resources
> * "TransactionResource" - directly relates to its functionality
> * Avoids the ambiguity you pointed out
> 
> Which approach do you prefer? Or do you have other suggestions for the design?

Regarding the naming, I lean towards TransactionParticipant instead of LocalTransactionResource. We could enhance TransactionParticipant with a parameter for the transaction mode. This would also allow it to handle more common settings in the future, like AT mode's isolation level. Consequently, configuring the lock retry interval via @GlobalTransaction would become unnecessary, since @GlobalTransaction is primarily for TM, not for RM. This new annotation could also enable local transaction features, like connection reuse until the aspect finishes and a unified commit. Or could it be changed to LocalTransactional? Or SeataLocalTransactional?
@slievrly @xingfudeshi @YongGoose @xjlgod PTAL

compatible/src/main/java/io/seata/rm/tcc/remoting/parser/LocalTCCRemotingParser.java
```diff
@@ -63,27 +66,37 @@ public RemotingDesc getServiceDesc(Object bean, String beanName) throws Framewor
...
+     * Determine whether there is a transaction participant annotation ({@link LocalTCC} or {@link TransactionParticipant}) 
+     * on the given class or its interfaces
+     * @param classType the class type to check
+     * @return true if the class has either LocalTCC or TransactionParticipant annotation
+     */
+    private boolean isTransactionParticipant(Class<?> classType) {
         Set<Class<?>> interfaceClasses = ReflectionUtil.getInterfaces(classType);
         for (Class<?> interClass : interfaceClasses) {
-            if (interClass.isAnnotationPresent(LocalTCC.class)) {
+            if (interClass.isAnnotationPresent(LocalTCC.class) || interClass.isAnnotationPresent(TransactionParticipant.class)) {
```
pengten:
```
interClass.isAnnotationPresent(LocalTCC.class) || interClass.isAnnotationPresent(TransactionParticipant.class)
```
This condition determines whether it would be more comfortable to encapsulate a method separately 🤔
It seems that this conditional judgment may continue to evolve in the future. Now it appears in multiple locations. Perhaps it would be better to converge?

common/src/main/java/org/apache/seata/common/transaction/api/TransactionParticipant.java
YongGoose: If it can also enable local transaction features like a unified commit, I'd cast my vote for `LocalTransactional`.

compatible/src/main/java/io/seata/rm/tcc/remoting/parser/LocalTCCRemotingParser.java
WangzJi: Thanks for the suggestion! 

I agree we should encapsulate the annotation check logic, but I suggest handling it directly in the existing `isLocalTCC()` method rather than creating a new one. This approach has several benefits:

1. Maintains backward compatibility
2. Method name still reflects its core purpose
3. Reduces changes needed in other code

If we need to add more annotation types in the future, we can simply extend the condition within this method.

What do you think about this approach?

pengten: > Thanks for the suggestion!
> 
> I agree we should encapsulate the annotation check logic, but I suggest handling it directly in the existing `isLocalTCC()` method rather than creating a new one. This approach has several benefits:
> 
> 1. Maintains backward compatibility
> 2. Method name still reflects its core purpose
> 3. Reduces changes needed in other code
> 
> If we need to add more annotation types in the future, we can simply extend the condition within this method.
> 
> What do you think about this approach?

This is one way, but we need to confirm whether this method needs to change its name.

compatible/src/main/java/io/seata/rm/tcc/remoting/parser/LocalTCCRemotingParser.java
WangzJi: Fix in [aa9a8eb](https://github.com/apache/incubator-seata/pull/7443/commits/aa9a8ebabb6dedb0f11dd57f536de0a4fe82e436)

common/src/main/java/org/apache/seata/common/transaction/api/TransactionParticipant.java
WangzJi: @funky-eyes @YongGoose Thanks for the feedback! I've adopted the `@LocalTransactional` naming as suggested by @YongGoose.

**Regarding transaction mode parameters:**
@funky-eyes Your suggestion about enhancing with transaction mode parameters and AT isolation levels is excellent for future iterations. For this PR, I focused on the core issue #7439 - removing Saga's semantic dependency on `@LocalTCC`. The current simple annotation provides a clean foundation that can be enhanced later.
This keeps the current PR focused while leaving room for those valuable enhancements in follow-up work.

funky-eyes: > @funky-eyes @YongGoose Thanks for the feedback! I've adopted the `@LocalTransactional` naming as suggested by @YongGoose.
> 
> **Regarding transaction mode parameters:** @funky-eyes Your suggestion about enhancing with transaction mode parameters and AT isolation levels is excellent for future iterations. For this PR, I focused on the core issue #7439 - removing Saga's semantic dependency on `@LocalTCC`. The current simple annotation provides a clean foundation that can be enhanced later. This keeps the current PR focused while leaving room for those valuable enhancements in follow-up work.

I agree with your opinion. However, the name of this annotation may need to be approved by the majority of the community at the next weekly meeting before this PR can be merged.

funky-eyes: The community's final decision is to separate the Saga annotation mode into a distinct annotation similar to LocalTCC, which will be dedicated exclusively to the Saga annotation mode.

WangzJi: Thank you for the clarification!

I understand the community's decision to create a dedicated annotation for Saga mode. I can easily adapt my current @LocalTransactional implementation to create a Saga-specific annotation.

What do you think about @SagaTransactional as the naming?

WangzJi:
Hi all,@funky-eyes @YongGoose @pengten.
Based on community feedback, I propose the following approach for decoupling Saga annotation from TCC:
1. Create `SagaTransactional` annotation in saga-annotation module (independent from @LocalTCC)
2. Implement `SagaTransactionalRemotingParser` to handle only `SagaTransactional`
3. Register parser separately via SPI

Does this approach look good to the community? I'll proceed with implementation once confirmed.
Thanks!

pengten:
> Hi all,@funky-eyes @YongGoose @pengten. Based on community feedback, I propose the following approach for decoupling Saga annotation from TCC:
> 
> 1. Create `SagaTransactional` annotation in saga-annotation module (independent from @LocalTCC)
> 2. Implement `SagaTransactionalRemotingParser` to handle only `SagaTransactional`
> 3. Register parser separately via SPI
> 
> Does this approach look good to the community? I'll proceed with implementation once confirmed. Thanks!

Sounds good to me.

funky-eyes:
> Hi all,@funky-eyes @YongGoose @pengten. Based on community feedback, I propose the following approach for decoupling Saga annotation from TCC:
> 
> 1. Create `SagaTransactional` annotation in saga-annotation module (independent from @LocalTCC)
> 2. Implement `SagaTransactionalRemotingParser` to handle only `SagaTransactional`
> 3. Register parser separately via SPI
> 
> Does this approach look good to the community? I'll proceed with implementation once confirmed. Thanks!



> Hi all,@funky-eyes @YongGoose @pengten. Based on community feedback, I propose the following approach for decoupling Saga annotation from TCC:
> 
> 1. Create `SagaTransactional` annotation in saga-annotation module (independent from @LocalTCC)
> 2. Implement `SagaTransactionalRemotingParser` to handle only `SagaTransactional`
> 3. Register parser separately via SPI
> 
> Does this approach look good to the community? I'll proceed with implementation once confirmed. Thanks!

I agree

tcc/src/test/java/org/apache/seata/rm/tcc/remoting/parser/LocalTCCRemotingParserTest.java
```diff
@@ -16,57 +16,204 @@
...
+import org.apache.seata.integration.tx.api.remoting.Protocols;
 import org.apache.seata.integration.tx.api.remoting.RemotingDesc;
-import org.apache.seata.rm.tcc.TccAction;
-import org.apache.seata.rm.tcc.TccActionImpl;
-import org.junit.jupiter.api.Assertions;
+import org.apache.seata.rm.tcc.api.LocalTCC;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.*;
```
funky-eyes: Do not use wildcard imports.

WangzJi: Done

funky-eyes (commented):
Please register your PR in those two files :

WangzJi:
> Please register your PR in those two files : https://github.com/apache/incubator-seata/blob/2.x/changes/zh-cn/2.x.md https://github.com/apache/incubator-seata/blob/2.x/changes/en-us/2.x.md

Done in [10acfb5](https://github.com/apache/incubator-seata/pull/7443/commits/10acfb56faf514b83a18854c0c5f765bae67edac).

changes/zh-cn/2.x.md
```diff
@@ -48,6 +48,7 @@
 - [[#7608](https://github.com/seata/seata/pull/7608)]  修改refreshToken方法中的参数名称
 - [[#7603](https://github.com/seata/seata/pull/7603)] 将Apache Tomcat依赖项从9.0.106升级到9.0.108 
 - [[#7614](https://github.com/seata/seata/pull/7614)] 更新 README.md
+- [[#7443](https://github.com/seata/seata/pull/7443)] 移除Saga注解对 @LocalTCC 的依赖
```
```suggestion
- [[#7443](https://github.com/seata/seata/pull/7443)] 将saga注释模式中的@LocalTCC替换为@SagaTransactional
```

changes/en-us/2.x.md
```diff
@@ -49,6 +49,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#7608](https://github.com/seata/seata/pull/7608)]  modify the parameter name in refreshToken method
 - [[#7603](https://github.com/seata/seata/pull/7603)] upgrade Apache Tomcat dependency from 9.0.106 to 9.0.108
 - [[#7614](https://github.com/seata/seata/pull/7614)] update README.md
+- [[#7443](https://github.com/seata/seata/pull/7443)] remove Saga annotation dependency on @LocalTCC
```
funky-eyes:
```suggestion
- [[#7443](https://github.com/seata/seata/pull/7443)] Replace @LocalTCC with @SagaTransactional in the saga annotation pattern
```
Wouldn't a description like "Replace @LocalTCC with @SagaTransactional in the saga annotation pattern" be more appropriate?

funky-eyes (approved):
LGTM

Issue #7406, opened 2025-06-04
Feature: Refactor Raft watch API to support HTTP/2


funky-eyes:
### Why you need it?

当前raft的元数据更新通知，通过http1的longpolling方式感知变化。而目前Seata Server已经支持http2协议，理论上稍加改造便可以支持http2方式，并且相对于longpolling方式而言，实时性更高。
需求：
1. 保留原本http1方式的watch接口
2. 增加新的http2方式的watch接口
3. client通过识别server版本号，决定使用http1或http2从而保证多版本兼容。
你可以提交多个pr来逐步完成这个特性。

Currently, Raft metadata update notifications rely on the long polling mechanism over HTTP/1 to detect changes. However, Seata Server now supports the HTTP/2 protocol, which, with some modifications, could enable support for HTTP/2-based notifications. Compared to long polling, HTTP/2 could provide better real-time performance.

Requirements:  
1. Retain the original watch interface based on HTTP/1.  
2. Add a new watch interface that supports HTTP/2.  
3. Clients should determine whether to use HTTP/1 or HTTP/2 for the watch interface based on the server version, ensuring compatibility across different versions.

You can submit multiple PRs to gradually implement this feature.

https://github.com/apache/incubator-seata/blob/2.x/server/src/main/java/org/apache/seata/server/controller/ClusterController.java

https://github.com/apache/incubator-seata/blob/2.x/discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java

YongGoose:
May I take on this task?
It might take a bit of time due to other issues, but I believe I can handle it :)

funky-eyes:
> May I take on this task? It might take a bit of time due to other issues, but I believe I can handle it :)

OK, no problem. Since we already have the long polling method in place, this task is not urgent.

YongGoose:
@funky-eyes 

If there are any additional tasks that need to be done, please feel free to create sub-issues as needed.

YongGoose:
Also, I’d appreciate it if you could take a look at the PR. (#7492)
I’ve been quite busy lately, so the progress has been a bit slow.
I’ll make sure to allocate more time for the next tasks. 🚀

YvCeung:
[...]

PR #7451, opened 2025-06-19
feature: support http2

funky-eyes:

Supports the HTTP/2 protocol and allows upgrading from HTTP/1 to HTTP/2, but only when the content-type is not grpc.
image


core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
```diff
@@ -0,0 +1,162 @@
...
+    protected void channelRead0(ChannelHandlerContext ctx, Http2StreamFrame msg) throws Exception {
+        if (requestHandled) {
+            return;
+        }
+        if (msg instanceof Http2HeadersFrame) {
+            Http2HeadersFrame headersFrame = (Http2HeadersFrame) msg;
+            this.http2Headers = headersFrame.headers();
+            headersReceived = true;
+            headersEndStream = headersFrame.isEndStream();
+            // 如果 header 已经 endStream，且没有 data，则直接处理
```
YongGoose: [nit] Please translate it to english :)

funky-eyes: > [nit] Please translate it to english :)

ok

core/src/main/java/org/apache/seata/core/protocol/detector/HttpDetector.java
```diff
@@ -52,10 +67,59 @@ private boolean startsWith(ByteBuf buffer, String prefix) {
...
+            @Override
+            public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+                if (cause instanceof java.io.IOException) {
+                    LOGGER.trace("Connection closed by client: {}", cause.getMessage());
+                } else {
+                    LOGGER.error("Exception caught in HTTP pipeline: ", cause);
+                }
+                ctx.close();
+            }
+        };
```
YongGoose, line 95: Is it okay not to override methods like `channelInactive` and `channelUnregistered`?

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                if (headersEndStream) {
+                    handleRequest(ctx);
+                }
+            } else if (msg instanceof Http2DataFrame) {
+                Http2DataFrame dataFrame = (Http2DataFrame) msg;
+                bodyBuffer.writeBytes(dataFrame.content());
+                if (dataFrame.isEndStream()) {
+                    handleRequest(ctx);
+                }
+            }
```
YongGoose, line 69: [nit] If there’s a possibility that the type of msg could be extended in the future, using a `switch` statement could be a good option.

core/src/main/java/org/apache/seata/core/rpc/netty/http/HttpDispatchHandler.java
```diff
@@ -82,13 +56,13 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpRequest httpRequest)
...
-                sendErrorResponse(ctx, HttpResponseStatus.NOT_FOUND, keepAlive);
+                sendErrorResponse(ctx, HttpResponseStatus.NOT_FOUND, false);
                 return;
             }
 
-            HttpContext httpContext = new HttpContext(httpRequest, ctx, keepAlive);
+            HttpContext<HttpRequest> httpContext = new HttpContext<>(httpRequest, ctx, keepAlive,HttpContext.HTTP_1_1);
             ObjectNode requestDataNode = OBJECT_MAPPER.createObjectNode();
-            requestDataNode.putIfAbsent("param", ParameterParser.convertParamMap(queryStringDecoder.parameters()));
+            requestDataNode.put("param", ParameterParser.convertParamMap(queryStringDecoder.parameters()));
```
xjlgod: The origin method in here is putIfAbsent

core/src/test/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+    @Test
+    void testHttp2GetRequestWithParameters() throws Exception {
+        // 构造 GET 请求
+        Http2Headers headers = new DefaultHttp2Headers();
+        headers.method("GET");
+        headers.path("/test?param=testValue");
+        Http2HeadersFrame headersFrame = new DefaultHttp2HeadersFrame(headers, true);
+        channel.writeInbound(headersFrame);
+
+        // 使用等待方法获取响应
```
xjlgod: Or we should use it in English

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
```diff
@@ -0,0 +1,174 @@
...
+        } catch (Exception e) {
+            if (bodyBuffer != null) {
+                bodyBuffer.release();
+                bodyBuffer = null;
+            }
+            throw e;
+        }
+    }
+
+    private void handleRequest(ChannelHandlerContext ctx) {
```
YongGoose, line 79: The method looks well written overall.

However, it might be a good idea to break it down into smaller methods.
Repeated try blocks, for example, could make the code harder to understand for someone reading it for the first time during future maintenance.

YongGoose (commented):
The code looks well written overall!

However, I noticed that there are quite a few comments written in Chinese in the test code. It would be great if you could update them to English :)

Also, for `BaseHttpChannelHandler`, it might be helpful to add some information about the role of this class, and which classes extend it.

Thank you for your work, always 👍🏻

core/src/main/java/org/apache/seata/core/rpc/netty/http/HttpDispatchHandler.java
YongGoose: Since the `requestDataNode` object is created just above, using `putIfAbsent` doesn’t seem strictly necessary.
image
> This is the constructor that is invoked by createObjectNode.

However, put has been marked as deprecated, so using `set` is recommended instead.

core/src/test/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandlerTest.java
YongGoose: +1

core/src/main/java/org/apache/seata/core/rpc/netty/http/HttpDispatchHandler.java
funky-eyes: > The origin method in here is putIfAbsent

There is no need to use the putIfAbsent method, so I will use the set method instead.

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
funky-eyes: > [nit] If there’s a possibility that the type of msg could be extended in the future, using a `switch` statement could be a good option.

Can you give an example?

core/src/main/java/org/apache/seata/core/protocol/detector/HttpDetector.java
funky-eyes: > Is it okay not to override methods like `channelInactive` and `channelUnregistered`?

Since most HTTP connections are short-lived, if you do not override the exceptionCaught method, a large number of unexpected connection closed exceptions will be output. For example, if you use the curl command to simulate a request, you will encounter such exceptions.

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
funky-eyes: ok

core/src/main/java/org/apache/seata/core/protocol/detector/HttpDetector.java
funky-eyes: `.\curl.exe -i -X POST  127.0.0.1:8091/health --http2`
```
23:23:16.110 ERROR --- [tyServerNIOWorker_1_10_16] [protocol.detector.HttpDetector] [     exceptionCaught]  [] : Connection closed by client: 
```

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
funky-eyes: done

core/src/test/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandlerTest.java
funky-eyes: done

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
YongGoose: Oh, pattern matching is only available starting from JDK 17.
With JDK 8, using if-else is probably the best option. 😅

server/src/main/java/org/apache/seata/server/cluster/manager/ClusterWatcherManager.java
```diff
@@ -87,37 +87,38 @@ public void onChangeEvent(ClusterChangeEvent event) {
...
-        HttpContext httpContext = (HttpContext)context;
-        ChannelHandlerContext ctx = httpContext.getContext();
-        if (ctx.channel().isActive()) {
-            HttpResponse response =
-                new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, nettyStatus, Unpooled.EMPTY_BUFFER);
-            response.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0);
+        ChannelHandlerContext ctx = context.getContext();
+        if (!context.isHttp2()) {
+            if (ctx.channel().isActive()) {
+                HttpResponse response =
```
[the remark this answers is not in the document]
funky-eyes: https://github.com/apache/incubator-seata/issues/7406 The task of this issue is to add support for HTTP/2.

core/src/main/java/org/apache/seata/core/rpc/netty/http/HttpDispatchHandler.java
```diff
@@ -82,13 +56,13 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpRequest httpRequest)
             HttpInvocation httpInvocation = ControllerManager.getHttpInvocation(path);
 
             if (httpInvocation == null) {
-                sendErrorResponse(ctx, HttpResponseStatus.NOT_FOUND, keepAlive);
+                sendErrorResponse(ctx, HttpResponseStatus.NOT_FOUND, false);
```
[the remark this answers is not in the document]
funky-eyes: A non-200 status code requires closing the connection, otherwise, the curl request will hang.

core/src/main/java/org/apache/seata/core/rpc/netty/http/Http2HttpHandler.java
funky-eyes: > Oh, pattern matching is only available starting from JDK 17. With JDK 8, using if-else is probably the best option. 😅

ok

core/src/main/java/org/apache/seata/core/protocol/detector/Http2Detector.java
```diff
@@ -50,17 +54,31 @@ public boolean detect(ByteBuf in) {
...
-            Http2FrameCodecBuilder.forServer().build(),
+        return new ChannelHandler[] {Http2FrameCodecBuilder.forServer().build(),
             new Http2MultiplexHandler(new ChannelInitializer<Http2StreamChannel>() {
                 @Override
                 protected void initChannel(Http2StreamChannel ch) {
                     final ChannelPipeline p = ch.pipeline();
-                    p.addLast(new GrpcDecoder());
-                    p.addLast(new GrpcEncoder());
-                    p.addLast(serverHandlers);
+                    p.addLast(new ChannelInboundHandlerAdapter() {
```
xjlgod: Should we not use an inner class here as we did before?

funky-eyes: done

core/src/main/java/org/apache/seata/core/protocol/detector/HttpDetector.java
```diff
@@ -52,7 +67,59 @@ private boolean startsWith(ByteBuf buffer, String prefix) {
...
+                return new Http2ServerUpgradeCodec(
+                        Http2FrameCodecBuilder.forServer().build(),
+                        new Http2MultiplexHandler(new ChannelInitializer<Http2StreamChannel>() {
+                            @Override
+                            protected void initChannel(Http2StreamChannel ch) {
+                                ch.pipeline().addLast(new Http2HttpHandler());
+                            }
+                        }));
+            } else {
+                return null;
```
YongGoose, line 119: Wouldn’t passing null cause a `NullPointerException`?

https://github.com/netty/netty/blob/dd42323c07db572265607cfad3c43995b6846a19/codec-http/src/main/java/io/netty/handler/codec/http/HttpServerUpgradeHandler.java#L243

funky-eyes: image
What returns null is not the factory class, but the return value of the newUpgradeCodec method.

YongGoose (approved):
LGTM

PR #7415, opened 2025-06-05
optimize: seata-server independent HTTP thread pool

YoWuwuuuw:
- [x] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did
Currently, seata-server uses netty to manage network requests, and HTTP request processing is executed synchronously with the io thread pool of netty itself. At the same time, due to the mechanism of multiple protocols on a single port, blocking may occur under high concurrency.
Therefore, an independent thread pool is added to process HTTP requests for processing of HTTP requests asynchronously. (align seata, grpc protocol asynchronous processing)

### Ⅱ. Does this pull request fix one issue?

#7405

PR #7461, opened 2025-06-21
Refactor: server netty config changed to use CONFIG format

YoWuwuuuw:

The original configuration method of secretly loading JVM startup parameters using `System.getProperty` has been changed to use a unified `CONFIG.getInt` format to align with the configuration format in other locations.

base on: https://github.com/apache/incubator-seata/pull/7415#discussion_r2134523446

At the beginning, `System.getProperty()` was introduced to open the configuration entry but not to be discovered by users. I think this is an unnecessary worry because users are unlikely to modify configuration items that they do not need. Moreover, this configuration method affects readability and hinders some needs.


YongGoose:
Run 'mvn spotless:apply' to fix these violations.

funky-eyes (approved):
LGTM

PR #7477, opened 2025-06-27
XAUtils add DmdbXAConnection

wangyjay:

XAUtils add DmdbXAConnection

in DM8.0+


rm-datasource/src/main/java/org/apache/seata/rm/datasource/util/XAUtils.java
funky-eyes, line 1: You can introduce a DM  database driver dependency and set its scope to test, then write a test case to verify whether the generated XAConnection meets expectations.

YongGoose:
I don't mean to be harsh, but i'm not a fan of code that violates the OCP (Open/Closed Principle).

PR #7492, opened 2025-07-02
feature : Upgrade HTTP client in common module to support HTTP/2

### LINKED ISSUE #7543: Upgrade HTTP Client in common Module to Support HTTP/2
### Check Ahead

### Why you need it?

Upgrade the HTTP client utility class in the `common` module to use HTTP Client 5.x. This change is the prerequisite to enabling HTTP/2 bidirectional communication between the server and clients, allowing real-time response delivery.

### How it could be?

_No response_

### Other related information

_No response_

YongGoose:

fixes #7543


server/src/main/java/org/apache/seata/server/controller/ClusterController.java
```diff
@@ -117,6 +118,12 @@ public void watch(
             @RequestBody Map<String, Object> groupTerms,
             @RequestParam(defaultValue = "28000") Integer timeout) {
         context.setAsync(true);
+
+        if (Version.isAboveOrEqualVersion250(Version.getCurrent())) {
```
funky-eyes: I think the logic for determining whether the server is version 2.5 should not be on the server side, but should be on the client side. The client should adopt the corresponding connection method to the server based on the server's version, and the server will naturally know whether the client is using HTTP/2 requests.

funky-eyes (commented):
I think we should first upgrade the HTTP client to version 5, so that the client can conveniently use HTTP/2 requests.

YongGoose:
> I think we should first upgrade the HTTP client to version 5, so that the client can conveniently use HTTP/2 requests.

Do I only need to upgrade raft, or should I upgrade common as well?

funky-eyes:
> > I think we should first upgrade the HTTP client to version 5, so that the client can conveniently use HTTP/2 requests.
> 
> Do I only need to upgrade raft, or should I upgrade common as well?

Currently, it appears we only need to upgrade the HTTP client utility class in common, and then this PR will implement real-time response delivery to clients through HTTP/2's bidirectional capabilities. Additionally, we need to implement another PR to integrate the Raft implementation in discovery with the new version of the HTTP client, and upgrade the protocol based on the server version in the metadata (note that the version might be null in version 2.0).

funky-eyes (commented):
This branch has conflicts that must be resolved

common/pom.xml
```diff
@@ -48,6 +48,10 @@
             <artifactId>httpclient</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents.client5</groupId>
```
funky-eyes: Since we've upgraded the HttpClient version to 5, is it still necessary to keep version 4?

common/src/main/java/org/apache/seata/common/util/HttpClientUtil.java
```diff
@@ -110,6 +118,61 @@ public static CloseableHttpResponse doPost(
...
+                    String requestBody = URLEncodedUtils.format(nameValuePairs, StandardCharsets.UTF_8);
+                    request.setBody(requestBody, org.apache.hc.core5.http.ContentType.APPLICATION_FORM_URLENCODED);
+                } else if (ContentType.APPLICATION_JSON.getMimeType().equals(contentType)) {
+                    String requestBody = OBJECT_MAPPER.writeValueAsString(params);
+                    request.setBody(requestBody, org.apache.hc.core5.http.ContentType.APPLICATION_JSON);
+                }
+            }
+
+            CompletableFuture<SimpleHttpResponse> future = new CompletableFuture<>();
+            http2Client.execute(request, new FutureCallback<SimpleHttpResponse>() {
```
funky-eyes: In the Server-side streaming RPC mode, can it ensure that a single request receives multiple responses? Is it necessary to define an AsyncResponseConsumer to handle the streaming responses?

common/src/main/java/org/apache/seata/common/util/HttpClientUtil.java
```diff
@@ -143,6 +206,56 @@ public static CloseableHttpResponse doPost(String url, String body, Map<String,
...
+            }
+
+            if (StringUtils.isNotBlank(contentType)) {
+                if (ContentType.APPLICATION_JSON.getMimeType().equals(contentType)) {
+                    request.setBody(body, org.apache.hc.core5.http.ContentType.APPLICATION_JSON);
+                }
+            }
+
+            CompletableFuture<SimpleHttpResponse> future = new CompletableFuture<>();
+            http2Client.execute(request, new FutureCallback<SimpleHttpResponse>() {
```
funky-eyes: ditto

common/pom.xml
YongGoose: Since we’ve upgraded to HttpClient version 5, I don’t think there’s a need to keep version 4 anymore.

I’m planning to proceed with the changes accordingly, but upgrading the version causes conflicts with other modules that still use HttpClient version 4.

Should we upgrade all of them as well?
If so, I think it would be better to handle the migration through a separate PR.

image

common/src/main/java/org/apache/seata/common/util/HttpClientUtil.java
YongGoose: > In the Server-side streaming RPC mode, can it ensure that a single request receives multiple responses?

As far as I know, it is possible.

> Is it necessary to define an AsyncResponseConsumer to handle the streaming responses?

When using HttpClient, I believe it’s necessary to define an AsyncResponseConsumer.

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7492/commits/dc051d26158a7ca4a626f9a3bc135eef218cb581

For the tests, I plan to implement them after completing all the other tasks.

common/pom.xml
funky-eyes: I think we can start by submitting a utility class fully implemented with HttpClient5 and add support for HTTP/2 in it. Then, we can mark the HttpClient4 version of the utility class as deprecated. Next, we can publish some community tasks, such as detecting whether there is an existing dependency on HttpClient5. If it exists, use version 5; if not, fall back to version 4. In this case, output a warning log to notify users that HttpClient4 will soon be removed.

common/src/main/java/org/apache/seata/common/util/HttpClientUtil.java
```diff
@@ -173,6 +294,46 @@ public static CloseableHttpResponse doGet(
...
+                        .setConnectionRequestTimeout(timeout, TimeUnit.MILLISECONDS)
+                        .build())
+                .build()) {
+            http2Client.start();
+
+            SimpleHttpRequest request = new SimpleHttpRequest("GET", url);
+            if (headers != null) {
+                headers.forEach(request::setHeader);
+            }
+            CompletableFuture<SimpleHttpResponse> future = new CompletableFuture<>();
```
funky-eyes: I believe this change is incorrect. This only uses an asynchronous approach to populate the response result. Instead, you should define an HttpCallback interface (to avoid a strong dependency on HttpClient), which can then be implemented by the external caller and passed in. Whenever HttpClient's FutureCallback receives a callback, it should invoke the caller's callback method.

common/pom.xml
YongGoose: Sounds good to me :)

YongGoose: Done in https://github.com/apache/incubator-seata/pull/7492/commits/e83568f20078453db884be9ec1d8bab359e68ee1 !

common/src/main/java/org/apache/seata/common/util/HttpClientUtil.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7492/commits/a4e36d58d8f12687196345266aa63ae8a334e26b !

common/pom.xml
YongGoose: I've created two related issues. 
If there's anything else that needs to be done, feel free to leave a comment! I'll keep thinking about it as well
- #7560 
- #7561

funky-eyes (commented):
It looks like there are no major issues with this PR. The remaining tasks are:
1. Add test cases for both HTTP/1 and HTTP/2.
2. Verify whether server push is working.

YongGoose:
> It looks like there are no major issues with this PR. The remaining tasks are:
> 
> 1. Add test cases for both HTTP/1 and HTTP/2.
> 2. Verify whether server push is working.

For server push, we need to modify the business logic, so I’m planning to handle it in a separate PR. 
What do you think?

For the tests done in https://github.com/apache/incubator-seata/pull/7492/commits/7190abc76dfcf27907f237d82a19bbf119c00299 !

YongGoose:
@funky-eyes 

The issue only seems to occur on JDK 8.
Do you happen to know anything about this?

funky-eyes:
> @funky-eyes
> 
> The issue only seems to occur on JDK 8. Do you happen to know anything about this?

The CloseableHttpAsyncClient you built is different for JDK 8 and versions above JDK 8. I think this is the reason for the issue. I don't understand why JDK 8 needs to be treated differently. As far as I remember, in OkHttp, both JDK 8 and higher versions support HTTP/2. Is there any difference with Apache HttpClient?

funky-eyes:
> For server push, we need to modify the business logic, so I’m planning to handle it in a separate PR. What do you think?
> 
> For the tests done in [7190abc](https://github.com/apache/incubator-seata/commit/7190abc76dfcf27907f237d82a19bbf119c00299) !

ok

YongGoose:
> The CloseableHttpAsyncClient you built is different for JDK 8 and versions above JDK 8. I think this is the reason for the issue. I don't understand why JDK 8 needs to be treated differently. As far as I remember, in OkHttp, both JDK 8 and higher versions support HTTP/2. Is there any difference with Apache HttpClient?

Apache HttpClient5 requires ALPN for HTTP/2.
However, JDK 8 does not provide ALPN support out of the box.
To enable HTTP/2 on JDK 8, i implemented it using Conscrypt.
- https://stackoverflow.com/questions/53387107/conscrypt-with-jdk8-to-enable-alpn-for-http2

Because of this limitation, even when we didn’t explicitly differentiate JDK 8, the build still failed on JDK 8.
- https://github.com/apache/incubator-seata/actions/runs/16593932387/job/46941952693

YongGoose:
There are two options:

The first is to make the test methods in `HttpClient5UtilTest` work in an JDK 11+ environment first, and then handle it together as mentioned in the issue below.
- https://github.com/apache/incubator-seata/issues/7561

The second is to address it in the current PR.

Which approach do you think is better?

funky-eyes:
> There are two options:
> 
> The first is to make the test methods in `HttpClient5UtilTest` work in an JDK 11+ environment first, and then handle it together as mentioned in the issue below.
> 
> * [Add Dependency Detection and Fallback Logic for HttpClient5 #7561](https://github.com/apache/incubator-seata/issues/7561)
> 
> The second is to address it in the current PR.
> 
> Which approach do you think is better?

I don't think this approach is appropriate, because if JDK 8 does not support HTTP/2, then upgrading to HttpClient 5 does not make much sense. Therefore, I suggest looking into OkHttp to see if it supports HTTP/2 on JDK 8. If it does, we can consider switching to OkHttp instead of HttpClient.

YongGoose:
> I don't think this approach is appropriate, because if JDK 8 does not support HTTP/2, then upgrading to HttpClient 5 does not make much sense. Therefore, I suggest looking into OkHttp to see if it supports HTTP/2 on JDK 8. If it does, we can consider switching to OkHttp instead of HttpClient.

Ok

YongGoose:
@funky-eyes 

Please take another look :)
https://github.com/apache/incubator-seata/pull/7492/commits/132c7de49f1b0bf74be5d0f31a4199b5931a1b1a

funky-eyes (approved):
LGTM

PR #7496, opened 2025-07-03
feature: add oceanbase oracle support

### LINKED ISSUE #7279: [feature]:add oceanbase oracle mode support

xxsc0529:

Oceanbase's MySQL mode is directly driven by MySQL connections. Due to the coding model, it is not convenient to distinguish between the specific patterns of querying the database according to connect, and only the oracle mode is adapted here. The code logic of Oracle is basically reused, and the reason why this part of the code is added is that the Oracle mode of OceanBase can only be connected through the OceanBase protocol. Due to the driver problem, the oracle mode of OceanBase cannot obtain the value of the auto-increment primary key in the oracle mode through getGenerateKeys, so the transaction control of the auto-increment SQL statement by sequence number is not supported.

close #7279

Test code that basically reuses the oracle schema
org.apache.seata.rm.datasource.exec.OceanBaseInsertExecutorTest
org.apache.seata.rm.datasource.sql.struct.TableMetaCacheFactoryTest

use https://github.com/apache/incubator-seata-samples

oceanbase is ant group open source products.
the oracle mode of OceanBase cannot obtain the value of the auto-increment primary key in the oracle mode through getGenerateKeys, so the transaction control of the auto-increment SQL statement by sequence number is not supported.


script/server/db/oceanbase.sql
```diff
@@ -0,0 +1,96 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+-- -------------------------------- The script used when storeMode is 'db' --------------------------------
+-- the table to store GlobalSession data
+CREATE TABLE global_table
```
slievrly: This SQL script is a DB initialization script used to adapt to the seata-server connection. I see that this PR is not compatible with the oceanbase storage of seata-server, so I understand that it is not necessary.

xxsc0529: I've deleted it

slievrly (approved):
LGTM

PR #7501, opened 2025-07-06
test: add unit test case for fury serializer

GoodBoyCoder:
- [x] I have registered the PR [changes](../changes).

### Ⅰ. Describe what this PR did
add unit test case for fury serializer

PR #7503, opened 2025-07-06
feature: support fory serializer and fory undolog parser

### LINKED ISSUE #7500: upgrade to Apache Fory 0.11.1
### Check Ahead

### Why you need it?

Apache Fury renamed to Apache Fory.
https://mvnrepository.com/artifact/org.apache.fory/fory-core

Fory 0.11.2 might be ready soon.

### How it could be?

_No response_

### Other related information

_No response_

diguage:

add fory serializer support

fixes #7500


funky-eyes:
I don't think this is a feature. First, Fory is just Fury after it was renamed. You should modify the current Fury serialization implementation to make it mutually compatible, regardless of whether the serialization identifier is 'fury' or 'fory', because they are actually the same type of serialization.

funky-eyes (changes requested):
I don't think this is a feature. First, Fory is just Fury after it was renamed. You should modify the current Fury serialization implementation to make it mutually compatible, regardless of whether the serialization identifier is 'fury' or 'fory', because they are actually the same type of serialization.

diguage:
> make it mutually compatible

I think it's easier to maintain another package than to maintain compatibility.

1. The old package can be deleted at any time
2. Use the same package and rename it in the future. (If not renamed, the package name is inconsistent with the internal serialization protocol.)

funky-eyes:
> > make it mutually compatible
> 
> I think it's easier to maintain another package than to maintain compatibility.
> 
> 1. The old package can be deleted at any time
> 2. Use the same package and rename it in the future. (If not renamed, the package name is inconsistent with the internal serialization protocol.)

I don't think so, because these two things are the same thing, there's no need to distinguish them, and compatibility is not an issue. Fastjson and fastjson2 are differentiated because early fastjson2 had incompatibility problems with fastjson, so they were distinguished. However, fury and fory don't have this problem at all - it's just like io.seata being changed to org.apache.seata. It should maintain backward compatibility. You can have fury directly reference fory's implementation, so both are directly compatible, because whenever fury serialization is called, it will actually use fory serialization.

diguage:
> I don't think so, because these two things are the same thing

If I follow your idea, I will first load org.apache.fory.Fory, and if successful, run it according to Fory. If the class cannot be found, try to load org.apache.fury.Fury.

No more new modules, still in the original modules `seata-serializer-fury`.

Is this OK?

funky-eyes:
> > I don't think so, because these two things are the same thing
> 
> If I follow your idea, I will first load org.apache.fory.Fory, and if successful, run it according to Fory. If the class cannot be found, try to load org.apache.fury.Fury.
> 
> No more new modules, still in the original modules `seata-serializer-fury`.
> 
> Is this OK?
> 
> 

I don't think this is necessary, because fory is just fury. As long as there's no fory dependency, just prompt the user to upgrade the dependency to fory, and describe the upgrade from fury to fory in the changelog so users can switch dependencies.

diguage:
> I don't think this is necessary, because fory is just fury.

OK.

> describe the upgrade from fury to fory in the changelog so users can switch dependencies.

Will this package always be called seata-serializer-fury?

funky-eyes:
> > I don't think this is necessary, because fory is just fury.
> 
> OK.
> 
> > describe the upgrade from fury to fory in the changelog so users can switch dependencies.
> 
> Will this package always be called seata-serializer-fury?

I think we should keep the seata-serializer-fury module, then have seata-serializer-fury depend on the seata-serializer-fory module, move the code logic from fury to the fory module, and add functionality to the fury module to check whether the fory dependency exists.

diguage:
> I think we should keep the seata-serializer-fury module, then have seata-serializer-fury depend on the seata-serializer-fory module

This is fine for now, but it will be confusing after fury is removed.

funky-eyes:
> > I think we should keep the seata-serializer-fury module, then have seata-serializer-fury depend on the seata-serializer-fory module
> 
> This is fine for now, but it will be confusing after fury is removed.

The Fury serialization feature is only available in recent versions, so there probably aren’t many users. Also, it’s just a matter of changing a dependency, so I don’t think it will have much impact.

slievrly (commented):
Please provide the necessary security test cases for the serialization framework.

diguage:
> the necessary security test cases

I read the test code of other serialization frameworks, and except for Hessian, the tests are basically the same.

Could you give some examples or explain it in detail?

slievrly:
> > the necessary security test cases
> 
> I read the test code of other serialization frameworks, and except for Hessian, the tests are basically the same.
> 
> Could you give some examples or explain it in detail?

https://github.com/apache/incubator-seata/pull/7501/files

The serialization framework should support a whitelist. Most security vulnerabilities are caused by unsafe deserialization. Serialization outside the whitelist should throw exceptions. When adding a whitelist, it is necessary to ensure that the overall transaction function is normal.

diguage:
> https://github.com/apache/incubator-seata/pull/7501/files
> 
> The serialization framework should support a whitelist.

OK, let me study it. And then add the security test cases.

diguage:
**In the seata-serializer-fury module, make it compatible with fory. In addition, add a module: seata-serializer-fory.** 

@funky-eyes  What do you think of this solution?

serializer/seata-serializer-fury/src/main/java/org/apache/seata/serializer/fury/FurySerializerFactory.java
```diff
@@ -16,42 +16,49 @@
...
-import org.apache.fury.ThreadLocalFury;
-import org.apache.fury.ThreadSafeFury;
-import org.apache.fury.config.CompatibleMode;
-import org.apache.fury.config.Language;
-import org.apache.seata.core.serializer.SerializerSecurityRegistry;
+import org.apache.seata.core.serializer.Serializer;
+
+import java.util.function.Function;
 
 public class FurySerializerFactory {
```
diguage: I think this class is redundant and I want to **use the following SerializerDelegate to replace it**. Is it possible?

funky-eyes:
> **In the seata-serializer-fury module, make it compatible with fory. In addition, add a module: seata-serializer-fory.**
> 
> @funky-eyes What do you think of this solution?

No problem. The fury module can depend directly on the fory module — fury is merely a backward‑compatible wrapper that delegates all internal calls to fory's implementation.

diguage:
> No problem. The fury module can depend directly on the fory module — fury is merely a backward‑compatible wrapper that delegates all internal calls to fory's implementation.

Ok, let me finish the code this weekend.

diguage:
> https://github.com/apache/incubator-seata/pull/7501/files
> 
> The serialization framework should support a whitelist.

I wrote similar code and ran the test, but they no longer throw exceptions. So, I do not add the kind of test code.

slievrly:
@diguage SerializerTypeTest.testAllSerializerTypesHaveUniqueCode:94 Serializer types FURY and FORY have the same code ==> expected: not equal but was: <86>

diguage:
> Serializer types FURY and FORY have the same code ==> expected: not equal but was: <86>

I fixed it.

slievrly:
@diguage Error:  Failures: 
Error:    FurySerializerTest.testUnSafeDeserializer:95 Unexpected exception type thrown ==> expected: <org.apache.fury.exception.DeserializationException> but was: <org.apache.fory.exception.DeserializationException>
Error:    FurySerializerTest.testUnSafeSerializer:103 Unexpected exception type thrown ==> expected: <org.apache.fury.exception.InsecureException> but was: <org.apache.fory.exception.InsecureException>

diguage:
> Error: Failures:

OK. Let me fix it.

@slievrly It's done.

funky-eyes (approved):
LGTM

slievrly (approved):
LGTM

PR #7505, opened 2025-07-06
bugfix: prevent Netty I/O thread blocking by async channel release via reconnectExecutor

### LINKED ISSUE #7497: The client I/O thread may be blocked
### Check Ahead

### Ⅰ. Issue Description

Image

Image

Image
在 NettyClientChannelManager.java 中，releaseChannel 方法和 acquireChannel 方法都会对 channelLocks 对象进行加锁。exceptionCaught 方法（位于 AbstractNettyRemotingClient.java 的内部类 ClientHandler 中）会调用 releaseChannel。而exceptionCaught是出于netty的io线程中，如果io线程核心数只有1，则会导致请求和接受响应的线程一并阻塞，导致流量跌零。当然只有当服务端非优雅下线时，并且恰好reconnect任务与执行exceptionCaught方法同时进行才会导致这个现象，因为服务端节点宕机后已经无法连接了，该同步锁会等待建立连接请求超时才会释放该锁，导致这个期间客户端的请求全部hang住
In NettyClientChannelManager.java, both the releaseChannel method and acquireChannel method acquire locks on the channelLocks object. The exceptionCaught method (located in the inner class ClientHandler of AbstractNettyRemotingClient.java) calls releaseChannel. Since exceptionCaught runs in Netty's I/O thread, if there's only 1 I/O thread core, it will cause both request and response handling threads to be blocked simultaneously, resulting in zero traffic throughput. This phenomenon only occurs when the server shuts down ungracefully and the reconnect task happens to run concurrently with the exceptionCaught method execution. Because the server node is down and no longer connectable, the synchronization lock will wait until the connection establishment request times out before releasing the lock, causing all client requests to hang during this period.

另外channelLocks似乎永远不会进行清理，当服务端在k8s中部署，每次重启都更换一个新的ip时，久而久之客户端就会面临oom的风险。
Additionally, channelLocks seems to never be cleaned up. When the server is deployed in Kubernetes and gets a new IP address with each restart, over time the client will face the risk of OOM (Out of Memory).

### Ⅱ. Describe what happened

_No response_

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

Dltmd202:

Fixes a potential Netty I/O thread blocking issue by executing `releaseChannel()` asynchronously via a dedicated `reconnectExecutor` thread pool.

Also ensures proper shutdown of `reconnectExecutor` to avoid thread leaks.

fixes #7497

- Naming conventions (`rpcReconnectExecutor`) aligned with existing merge thread patterns.
- `reconnectExecutor` is now managed with proper init and destroy lifecycle hooks.

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
@@ -681,8 +681,13 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
             if (LOGGER.isInfoEnabled()) {
                 LOGGER.info("channel inactive: {}", ctx.channel());
             }
-            clientChannelManager.releaseChannel(
-                    ctx.channel(), NetUtil.toStringAddress(ctx.channel().remoteAddress()));
+            timerExecutor.execute(() -> {
+                try {
+                    clientChannelManager.releaseChannel(ctx.channel(), getAddressFromChannel(ctx.channel()));
+                } catch (Throwable throwable) {
+                    LOGGER.error("release channel error: {}", throwable.getMessage(), throwable);
+                }
+            });
             super.channelInactive(ctx);
         }
 
@@ -701,7 +706,18 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
                     } catch (Exception exx) {
                         LOGGER.error(exx.getMessage());
                     } finally {
-                        clientChannelManager.releaseChannel(ctx.channel(), getAddressFromContext(ctx));
+                        try {
+                            timerExecutor.execute(() -> {
+                                try {
+                                    clientChannelManager.releaseChannel(
+                                            ctx.channel(), getAddressFromChannel(ctx.channel()));
+                                } catch (Throwable throwable) {
+                                    LOGGER.error("release channel error: {}", throwable.getMessage(), throwable);
+                                }
+                            });
+                        } catch (Exception e) {
+                            LOGGER.error("failed to schedule releaseChannel: {}", e.getMessage(), e);
+                        }
                     }
                 }
                 if (idleStateEvent == IdleStateEvent.WRITER_IDLE_STATE_EVENT) {
@@ -723,7 +739,13 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
                     FrameworkErrorCode.ExceptionCaught.getErrCode(),
                     NetUtil.toStringAddress(ctx.channel().remoteAddress()) + "connect exception. " + cause.getMessage(),
                     cause);
-            clientChannelManager.releaseChannel(ctx.channel(), getAddressFromChannel(ctx.channel()));
+            timerExecutor.execute(() -> {
+                try {
+                    clientChannelManager.releaseChannel(ctx.channel(), getAddressFromChannel(ctx.channel()));
+                } catch (Throwable throwable) {
+                    LOGGER.error("release channel error: {}", throwable.getMessage(), throwable);
+                }
+            });
             if (LOGGER.isInfoEnabled()) {
                 LOGGER.info("remove exception rm channel:{}", ctx.channel());
             }
--- a/core/src/test/java/org/apache/seata/core/rpc/netty/ResourceCleanupTest.java
+++ b/core/src/test/java/org/apache/seata/core/rpc/netty/ResourceCleanupTest.java
@@ -17,7 +17,9 @@
 package org.apache.seata.core.rpc.netty;
 
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelId;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -137,6 +140,13 @@ public void init() {
                     new NamedThreadFactory(getThreadPrefix(), MAX_MERGE_SEND_THREAD));
             mergeSendExecutorService.submit(new MergedSendRunnable());
         }
+        reconnectExecutor = new ThreadPoolExecutor(
```
funky-eyes: Why not use the timerExecutor thread pool directly?

Dltmd202: Hi @funky-eyes
I added a separate thread pool mainly to give it a clear thread name, so it’s easier to trace when reconnect-related issues happen. But if that feels unnecessary here, I’m happy to switch to using the existing `timerExecutor`. Let me know what you think!

YongGoose: IMO, creating a new thread pool is generally preferred when there’s a high-priority task or when real-time processing is critical. Also, how frequently the issue occurs can be an important factor in deciding whether to create a new thread pool.

From a simple traceability standpoint, the current `timeoutExecutor` is used in several places — such as handling reconnections and removing timed-out messages.

So, it might make sense to update the prefix to something more general that fits well across all these usages, and reuse it accordingly.

funky-eyes: > Hi @funky-eyes I added a separate thread pool mainly to give it a clear thread name, so it’s easier to trace when reconnect-related issues happen. But if that feels unnecessary here, I’m happy to switch to using the existing `timerExecutor`. Let me know what you think!

I wrote very clearly in the issue why we should reuse the timeoutExecutor. I suggest you can take a look at the reason.

Dltmd202: Thanks for pointing that out. I revisited the issue and I now see your point more clearly — reusing timeoutExecutor makes sense, especially since it’s already used for similar tasks like handling reconnections and timeouts. Avoiding an extra thread pool also helps keep things lean and easier to manage.

I’ll go ahead and update the code to reuse timeoutExecutor accordingly. Appreciate the feedback!

Dltmd202: I’ve finished the changes. Would you mind taking a look when you have time?
@funky-eyes @YongGoose

funky-eyes (approved):
LGTM

core/src/test/java/org/apache/seata/core/rpc/netty/ResourceCleanupTest.java
```diff
@@ -143,6 +145,26 @@ void testCleanupWithNullChannel() {
...
+        channelManagerField.setAccessible(true);
+        NettyClientChannelManager originalManager = (NettyClientChannelManager) channelManagerField.get(client);
+
+        NettyClientChannelManager spyManager = spy(originalManager);
+        channelManagerField.set(client, spyManager);
+
+        handler.exceptionCaught(mockCtx, new IllegalArgumentException("test"));
+
+        Thread.sleep(500);
+        verify(spyManager).releaseChannel(eq(channel), anyString());
```
YongGoose: Is there a specific reason you used `anyString()`?
It seems like serverAddress will always be `127.0.0.1:8091`, so I’m wondering if matching the exact value would be more appropriate.

PeppaO:
This modification cannot solve the problem. I think you can reproduce the problem first, then analyze the cause, and retest after modification.

PeppaO:
To reproduce, start three tc nodes 8091/8092/8093, start a business-xa application, observe the corresponding tcp connection port, then kill 8091, and observe whether the tcp connection is disconnected and reconnected

PeppaO:
After resolving the disconnection and reconnection, perform a stress test to see if the TPS is 0 during the shutdown period after killing one of the tc nodes.
@funky-eyes @Dltmd202

funky-eyes (commented):
All places in ChannelHandler that call the clientChannelManager.releaseChannel method need to be made asynchronous.

core/src/test/java/org/apache/seata/core/rpc/netty/ResourceCleanupTest.java
Dltmd202: You’re right — that makes sense to me as well. I’ll go ahead and update it to match the exact value instead of using `anyString()`.

funky-eyes:
18.18% of diff hit (target 60.50%)
Can you add some test cases to improve the coverage?

Dltmd202:
Sure, I’ll take care of it!

Dltmd202:
I’ve finished the changes. Would you mind taking a look when you have time?
@funky-eyes

PR #7506, opened 2025-07-07
test: add test case for package spring.rm.fence

YvCeung:

add test case for package spring.rm.fence

No singlet test case was added for the logic of asynchronous log cleaning in SpringFenceHandler because
` FenceLogIdentity logIdentity = LOG_QUEUE.take(); ` It will cause the daemon thread to be blocked and eventually lead to the single test being stuck


YvCeung:
image
`private static final CommonFenceStore COMMON_FENCE_DAO = CommonFenceStoreDataBaseDAO.getInstance();`

This static member variable cannot modify access restrictions through reflection in jdk17 or later versions, making it impossible to mock. Moreover, there is no need to change the source code structure for coverage, so turn off this pr

PR #7548, opened 2025-07-18
Replace `mockito-inline` with `mockito-core`

strangelookingnerd:

Replace mockito-inline with mockito-core

In version 5.x mockito switched their default mockmaker to mockito-inline ([source](https://github.com/mockito/mockito/releases/tag/v5.0.0)).
The corresponding artifact became obsolete and can be replaced with mockito-core

This PR replaces / removes the artifact and switches on mockito-core which in turn allows upgrading to latest 5.18.0.

Enables upgrading mockito.

Does not apply

Build locally without issues

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -135,8 +135,7 @@
         <tomcat-embed.version>9.0.106</tomcat-embed.version>
 
         <!-- # for test -->
-        <mockito.version>4.11.0</mockito.version>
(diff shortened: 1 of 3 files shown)


YongGoose:
If we update the `mockito` dependency to the `5.18.0` version , it will not work with our current Java 8 build environment. The reason is that version 5.18.0 is compiled for Java 11, while our codebase is built using Java 8.

Issue #7523, opened 2025-07-11
When destroying `RmNettyRemotingClient`, all `TableMetaRefreshHolder` should actively close the process


linghengqian:
### Check Ahead

- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

- [ ] I am willing to try to fix this bug myself.

### Ⅰ. Issue Description

- This issue is actually the successor of https://github.com/apache/incubator-seata/issues/7042 , but because https://github.com/apache/incubator-seata/pull/7044 circumvents the problems encountered when using HikariCP in a clever way, I cannot reproduce the problem described in the current issue using only Postgres JDBC Driver and Seata Client in https://github.com/linghengqian/seata-refresh-holder-test/blob/master/src/test/java/com/github/linghengqian/SimpleTest.java .🤣
- To explain what I found in https://github.com/apache/shardingsphere/pull/35427 , I introduced a minimal reproducible unit test in https://github.com/linghengqian/seata-refresh-holder-test/blob/master/src/test/java/com/github/linghengqian/ShardingSphereTest.java . This needs to be explained by introducing the ShardingSphere Proxy into the process.
- When shardingsphere proxy is started in-process and Seata Client is destroyed by `org.apache.shardingsphere.mode.manager.ContextManager#close()`, the following operation of `org.apache.shardingsphere.transaction.base.seata.at.SeataATShardingSphereTransactionManager#close()` is not enough to completely destroy Seata Client.
```java
import org.apache.seata.config.ConfigurationFactory;
import org.apache.seata.core.rpc.netty.RmNettyRemotingClient;
import org.apache.seata.core.rpc.netty.TmNettyRemotingClient;
    public void close() {
        TmNettyRemotingClient.getInstance().destroy();
        RmNettyRemotingClient.getInstance().destroy();
        ConfigurationFactory.reload();
    }
```
- Once I execute `Awaitility.await().timeout(1L, TimeUnit.HOURS).pollDelay(2L, TimeUnit.MINUTES).until(() -> true);` in https://github.com/linghengqian/seata-refresh-holder-test/blob/master/src/test/java/com/github/linghengqian/ShardingSphereTest.java, the internal thread of Seata Client will continue to try to refresh the metadata.
```
[INFO] Running com.github.linghengqian.ShardingSphereTest
[...]

PR #7559, opened 2025-07-27
feature: Introduce Cleanup API for `TableMetaRefreshHolder` Instance

YongGoose:

#7523


rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
```diff
@@ -108,7 +108,19 @@ private static void removeHolderFromMap(String resourceId) {
         LOGGER.info("Removed TableMetaRefreshHolder for resourceId: {}", resourceId);
     }
 
+    /**
+     * Shutdown all TableMetaRefreshHolder threads.
+     */
+    public static void shutdown() {
```
funky-eyes: Is it possible for us to use the publish-subscribe pattern, so that after the rm close message is published, TableMetaCache proceeds to shut down?

YongGoose: > Is it possible for us to use the publish-subscribe pattern, so that after the rm close message is published, TableMetaCache proceeds to shut down?

event

Something like this could work.

However, compared to explicitly calling the `shutdown API` when the `RM` is `destroyed`, I don’t see a significant advantage.

If we go with the approach where `TableMetaCacheFactory` implements a `listener`, we could consider making shutdown a `private API`.

What do you think?

funky-eyes: When registering itself with the Seata-Server (TC), I believe that `DataSourceProxy` should proxy the `close` method. Upon closing, it should remove itself from both `DefaultResourceManager` and `TableMetaCacheFactory`. Currently, `TableMetaCacheFactory#removeHolderFromMap` is executed passively—only when refreshing table metadata encounters an exception due to a closed datasource does the removal occur. This approach, however, feels inelegant.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -47,7 +47,7 @@
  * The type Data source proxy.
  *
  */
-public class DataSourceProxy extends AbstractDataSourceProxy implements Resource {
+public class DataSourceProxy extends AbstractDataSourceProxy implements Resource, AutoCloseable {
```
YongGoose: How about implementing `AutoCloseable` so that it can be automatically closed using a try-with-resources statement?

WDYT?

rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
linghengqian: > When registering itself with the Seata-Server (TC), I believe that `DataSourceProxy` should proxy the `close` method. Upon closing, it should remove itself from both `DefaultResourceManager` and `TableMetaCacheFactory`. Currently, `TableMetaCacheFactory#removeHolderFromMap` is executed passively—only when refreshing table metadata encounters an exception due to a closed datasource does the removal occur. This approach, however, feels inelegant.

- Out of curiosity, I don't seem to see any class that actually implements `org.apache.seata.core.model.ResourceManager#unregisterResource(org.apache.seata.core.model.Resource)`.
```shell
org.apache.seata.common.exception.NotSupportYetException: unregister a resource
	at org.apache.seata.rm.datasource.DataSourceManager.unregisterResource(DataSourceManager.java:96)
	at org.apache.seata.rm.DefaultResourceManager.unregisterResource(DefaultResourceManager.java:125)
```

YongGoose: I think so too

funky-eyes: > > When registering itself with the Seata-Server (TC), I believe that `DataSourceProxy` should proxy the `close` method. Upon closing, it should remove itself from both `DefaultResourceManager` and `TableMetaCacheFactory`. Currently, `TableMetaCacheFactory#removeHolderFromMap` is executed passively—only when refreshing table metadata encounters an exception due to a closed datasource does the removal occur. This approach, however, feels inelegant.
> 
> * Out of curiosity, I don't seem to see any class that actually implements `org.apache.seata.core.model.ResourceManager#unregisterResource(org.apache.seata.core.model.Resource)`.
> 
> ```shell
> org.apache.seata.common.exception.NotSupportYetException: unregister a resource
> 	at org.apache.seata.rm.datasource.DataSourceManager.unregisterResource(DataSourceManager.java:96)
> 	at org.apache.seata.rm.DefaultResourceManager.unregisterResource(DefaultResourceManager.java:125)
> ```

Yes, that’s why we need to actually perform the unregisterResource operation on it.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
funky-eyes: > How about implementing `AutoCloseable` so that it can be automatically closed using a try-with-resources statement?
> 
> WDYT?

I don’t think it’s necessary, because the datasource itself already provides a close interface.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
YongGoose: > Yes, that’s why we need to actually perform the unregisterResource operation on it.

Shall we handle this together in this PR?

funky-eyes: > > Yes, that’s why we need to actually perform the unregisterResource operation on it.
> 
> Shall we handle this together in this PR?

Ok

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7559/commits/f111e690a5fc569e00fb96fc5186710ff41b2c61 !

rm-datasource/src/main/java/org/apache/seata/rm/datasource/sql/struct/TableMetaCacheFactory.java
YongGoose: Could you please let me know if there’s anything I’m missing or need to improve in https://github.com/apache/incubator-seata/commit/eea1fdd6b59e342062f0d39acd03709bb2c54f2c !

core/src/main/java/org/apache/seata/core/protocol/UnregisterRMRequest.java
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+package org.apache.seata.core.protocol;
+
+import java.io.Serializable;
+
+/**
+ * Represents a request to unregister a resource from the Resource Manager (RM).
+ */
+public class UnregisterRMRequest extends AbstractIdentifyRequest implements Serializable {
```
funky-eyes: It seems that I haven’t found anywhere on the server side that handles this request.

funky-eyes:
1. The functionality of UnregisterRMRequest should be separated into an individual PR, with the current PR focusing solely on resolving the issue where tableMetaRefresh does not terminate.
2. In NettyClientChannelManager, cache the rpcContext for a channel based on the response from RegisterRMResponse, and also cache the server version number. You may refer to the design of the RpcContext class for this implementation.
3. When checking the server version, if it is lower than 2.6.0, output a log indicating that UnregisterRMRequest is not supported by older server versions, and therefore, resources cannot be unregistered.
4. RMClient should provide both destroy and destroyAll interfaces. The destroy interface should accept resource information, construct an UnregisterRMRequest, and send it to the server to unregister the resource. The destroyAll interface should proactively notify the server that the client node is going offline.
5. These points can be split into multiple PRs to improve review efficiency and ensure that each PR is clear and focused.

Hi @slievrly, The above is a summary of our previous meeting. Do you have any additional comments or suggestions?

YongGoose:
> 5\. These points can be split into multiple PRs to improve review efficiency and ensure that each PR is clear and focused.

Agree

YongGoose:
> 1. The functionality of UnregisterRMRequest should be separated into an individual PR, with the current PR focusing solely on resolving the issue where tableMetaRefresh does not terminate.

Please take another look!
Reverted the commit to https://github.com/apache/incubator-seata/pull/7559/commits/f111e690a5fc569e00fb96fc5186710ff41b2c61!

YongGoose:
> * In NettyClientChannelManager, cache the rpcContext for a channel based on the response from RegisterRMResponse, and also cache the server version number. You may refer to the design of the RpcContext class for this implementation.
> * When checking the server version, if it is lower than 2.6.0, output a log indicating that UnregisterRMRequest is not supported by older server versions, and therefore, resources cannot be unregistered.
> * RMClient should provide both destroy and destroyAll interfaces. The destroy interface should accept resource information, construct an UnregisterRMRequest, and send it to the server to unregister the resource. The destroyAll interface should proactively notify the server that the client node is going offline.
> * These points can be split into multiple PRs to improve review efficiency and ensure that each PR is clear and focused.

For this tasks, I'll open a issue and get to work :)
- https://github.com/apache/incubator-seata/issues/7593
- https://github.com/apache/incubator-seata/issues/7594

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -452,4 +452,9 @@ private void validMySQLVersion(Connection connection) {
             LOGGER.error("check mysql version fail error: {}", e.getMessage());
         }
     }
+
+    public void close() throws Exception {
```
funky-eyes, line 456:
```suggestion
    @Override
    public void close() throws Exception {
```

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
```diff
@@ -452,4 +452,9 @@ private void validMySQLVersion(Connection connection) {
...
+        DefaultResourceManager.get().unregisterResource(this);
```
funky-eyes: The unregisterResource method of DataSourceManager throws a NotSupportYetException with the message "unregister a resource".

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
YongGoose: Since DataSourceProxy doesn’t extend AutoCloseable, I don’t think it can be overridden.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
YongGoose: That’s correct.
But wasn’t the implementation for that part planned to be done in a separate PR?

funky-eyes: > That’s correct. But wasn’t the implementation for that part planned to be done in a separate PR?

So I think this line of code should be added in a separate PR, not in this one.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
funky-eyes: > Since DataSourceProxy doesn’t extend AutoCloseable, I don’t think it can be overridden.

You're right — I was mistaken.

rm-datasource/src/main/java/org/apache/seata/rm/datasource/DataSourceProxy.java
YongGoose: Done in https://github.com/apache/incubator-seata/pull/7559/commits/189b5c600edab191380c7ab2293e1a62de46e8d9 !

funky-eyes: > Done in [189b5c6](https://github.com/apache/incubator-seata/commit/189b5c600edab191380c7ab2293e1a62de46e8d9) !

I think there may have been a misunderstanding. I meant that the line
`DefaultResourceManager.get().unregisterResource(this);`
should be removed from this PR and only reintroduced once unregisterResource has been properly implemented.

YongGoose: OK 👍🏻

funky-eyes (commented):
I think we should rewrite the PR title and then register the pull request.

funky-eyes (commented):
One more question: should the TABLE_META_CACHE in AbstractTableMetaCache be cleared when a data source is closed? I also noticed TABLE_META_CACHE is shared globally — under multiple data sources, if different databases contain tables with the same name, could that cause metadata conflicts? In addition, when calling tableMetaCache.refresh the connection used is tied to a single database, and that connection cannot retrieve metadata for tables in other databases. @slievrly, could you please confirm whether the issues I described actually exist?

slievrly:
> 5\. These points can be split into multiple PRs to improve review efficiency and ensure that each PR is clear and focused.

OK, go ahead.

slievrly (approved):
LGTM

YongGoose:
@slievrly 

> One more question: should the TABLE_META_CACHE in AbstractTableMetaCache be cleared when a data source is closed? I also noticed TABLE_META_CACHE is shared globally — under multiple data sources, if different databases contain tables with the same name, could that cause metadata conflicts? In addition, when calling tableMetaCache.refresh the connection used is tied to a single database, and that connection cannot retrieve metadata for tables in other databases.

Please take a look at this question :)

funky-eyes (approved):
LGTM

TODO:

> One more question: should the TABLE_META_CACHE in AbstractTableMetaCache be cleared when a data source is closed? I also noticed TABLE_META_CACHE is shared globally — under multiple data sources, if different databases contain tables with the same name, could that cause metadata conflicts? In addition, when calling tableMetaCache.refresh the connection used is tied to a single database, and that connection cannot retrieve metadata for tables in other databases. @slievrly, could you please confirm whether the issues I described actually exist?

PR #7563, opened 2025-07-29
bugfix: fix NPE when server-side filter is disabled and filter chain is null.

YvCeung:

the NullPointerException caused by the filter chain being null when the server-side filter is disabled.
image


funky-eyes (approved):
LGTM

PR #7570, opened 2025-08-05
bugfix: fix order() behavior in GlobalTransactionalInterceptorHandler to ensure correct sorting of invocation handlers

### LINKED ISSUE #7568: The order in AbstractProxyInvocationHandler serves no purpose
### Check Ahead

### Ⅰ. Issue Description

https://github.com/apache/incubator-seata/blob/2.x/integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
https://github.com/apache/incubator-seata/blob/2.x/integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/parser/DefaultInterfaceParser.java#L84
`Collections.sort(invocationHandlerList, Comparator.comparingInt(ProxyInvocationHandler::order));`

The method order is the one actually used for sorting, not getOrder. Therefore, the order in AbstractProxyInvocationHandler is entirely unnecessary. I suggest setting the value of order in AbstractProxyInvocationHandler to 0 and implementing the order method in the parent class ProxyInvocationHandler. This approach ensures backward compatibility and prevents issues with aspect order inconsistencies when directly implementing the order method in the parent class.

### Ⅱ. Describe what happened

_No response_

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

keepConcentration:

This PR fixes the incorrect sorting behavior of `GlobalTransactionalInterceptorHandler` in `DefaultInterfaceParser.parserInterfaceToProxy()`.
Although the `order` field was being set using `getOrder()` and `setOrder()`, it was not being reflected in the sorting logic due to the lack of an overridden `order()` method.
The fix ensures that all `ProxyInvocationHandler` implementations, including `GlobalTransactionalInterceptorHandler`, are sorted based on the intended `order` value.

fixes #7568 

This fixes the handler ordering inconsistency described in the related issue.


integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
YongGoose, line 1: Is there a specific reason why `order` method wasn’t implemented in `AbstractProxyInvocationHandler`?
IMO, Implementing `order` in `AbstractProxyInvocationHandler` would allow us to take advantage of inheritance.

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
```diff
@@ -59,7 +59,7 @@ public void setOrder(int order) {
 
     @Override
     public int getOrder() {
-        return this.order;
+        return this.order();
```
YongGoose: Is there a reason why the `order` function needs to return?

YongGoose (commented):
Nice work!
Left some comments.

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
keepConcentration: I observed that both `SagaAnnotationActionInterceptorHandler` and `TccActionInterceptorHandler`, which inherit from `AbstractProxyInvocationHandler`, explicitly override the `order()` method.

To maintain consistency and clarity across all subclasses, I felt it would be more appropriate for `GlobalTransactionalInterceptorHandler` to do the same, rather than relying on a shared implementation in the abstract class.

Of course, if you believe it would be better to introduce a shared `order()` implementation in the abstract class, I'm more than happy to follow that direction.
I'm open to adjusting the PR according to your preference or the maintainers' conventions.

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
keepConcentration: You're right — the line should return the `order` field directly instead of calling `this.order()`. I will update the code accordingly to return the `order` field as you suggested !

keepConcentration:
Thank you for the review! Let me summarize the feedback as I understand it:
1. Implement the `order()` method in `AbstractProxyInvocationHandler`:
```java
    @Override
    public int order() {
        return this.order;
    }
```

2. Implement the `order()` method in `GlobalTransactionalInterceptorHandler`:
```java
    @Override
    public int order() {
        return this.order;
    }
```
→ With this, all classes implementing `ProxyInvocationHandler` will now explicitly implement the `order()` method.

3. The `getOrder()` method in `AbstractProxyInvocationHandler` returns `this.order`:
```java
    @Override
    public int getOrder() {
        return this.order;
    }
```
 
One thing I noticed is that both the `order()` and `getOrder()` methods in `AbstractProxyInvocationHandler` now have the same implementation, which results in a bit of duplication.

Please let me know if this approach looks good to you. I’ll proceed accordingly once I hear back from you.

YongGoose:
> One thing I noticed is that both the `order()` and `getOrder()` methods in `AbstractProxyInvocationHandler` now have the same implementation, which results in a bit of duplication.
> 
> Please let me know if this approach looks good to you. I’ll proceed accordingly once I hear back from you.

Since the original issue was that `getOrder` wasn’t working correctly, this approach looks fine to me.
@funky-eyes  WDYT?

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
YongGoose: I believe adding it to the abstract class would be a better approach, as this way the subclasses aren’t forced to implement it explicitly through inheritance.

YongGoose:
@keepConcentration 

Additionally, for the tests, it would be good to declare multiple classes inheriting from `AbstractProxyInvocationHandler`, assign different `order` values to each, and verify that they are sorted as intended.

YongGoose:
@keepConcentration 

When work is done, Don't hesitate to ping me :)

keepConcentration:
@YongGoose

Thank you again for the detailed feedback!
I've updated the implementation based on your suggestions, including:

- Implemented the `order()` method in `AbstractProxyInvocationHandler`
- Added a test to verify sorting behavior in `AbstractProxyInvocationHandler`

Please feel free to take another look when you have time.
Let me know if there’s anything else I should adjust 🙏

YongGoose (changes requested):
Please resolve the conflicts.

keepConcentration:
@YongGoose

Ok, I resolved!

keepConcentration:
@YongGoose 

Thank you for the earlier feedback 🙏
I have made the following additional changes:

- Added a unit test (`testSetOrderAndSorting`) to verify sorting behavior:
  - Creates multiple `ProxyInvocationHandler` instances with different order values
  - Uses `setOrder()` to change the `order` dynamically
  - Asserts both pre-sort and post-sort ordering using the `order()` method

This should ensure that the ordering logic in `AbstractProxyInvocationHandler` works as intended.
Please review when you have time — I believe this is now ready for review ✅

integration-tx-api/src/test/java/org/apache/seata/integration/tx/api/interceptor/parser/DefaultInterfaceParserTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
```
YongGoose: Please avoid using `*`

YongGoose (approved):
LGTM 🚀
Also, I'd appreciate it if you could revise the parts mentioned in the comments :)

integration-tx-api/src/test/java/org/apache/seata/integration/tx/api/interceptor/parser/DefaultInterfaceParserTest.java
keepConcentration: Got it! I’ll replace the wildcard imports with explicit ones.

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/AbstractProxyInvocationHandler.java
```diff
@@ -28,7 +28,7 @@ public abstract class AbstractProxyInvocationHandler implements ProxyInvocationH
 
     protected abstract Object doInvoke(InvocationWrapper invocation) throws Throwable;
 
-    protected int order = Integer.MAX_VALUE;
+    protected int order = 0;
```
funky-eyes:
```suggestion
    protected int order = Integer.MAX_VALUE;
```
I think we should continue to use Integer.MAX_VALUE for this value, and then return 0 in the order method of GlobalTransactionalInterceptorHandler, as the order for GlobalTransactionalInterceptorHandler.

YongGoose: Considering the existing handler order, this seems like a good approach.

YongGoose:
@funky-eyes 

The TCC transaction propagation test is failing.
Could this be related to [this comment](https://github.com/apache/incubator-seata/issues/7568#issuecomment-3170466342)?

integration-tx-api/src/main/java/org/apache/seata/integration/tx/api/interceptor/handler/GlobalTransactionalInterceptorHandler.java
```diff
@@ -442,4 +442,9 @@ public SeataInterceptorPosition getPosition() {
     public String type() {
         return InvocationHandlerType.GlobalTransactional.name();
     }
+
+    @Override
+    public int order() {
+        return this.order;
```
funky-eyes:
```suggestion
        return 0;
```

funky-eyes:
> @funky-eyes
> 
> The TCC transaction propagation test is failing. Could this be related to [this comment](https://github.com/apache/incubator-seata/issues/7568#issuecomment-3170466342)?

I think the order of the test cases has changed again, which caused this issue. You can see that the exception started appearing right after this PR was merged: https://github.com/apache/incubator-seata/actions/runs/16847550947/job/47731890204

funky-eyes (approved):
LGTM

funky-eyes:
If you're using the DingTalk app, please send your DingTalk account to my email at `jianbin@apache.org`, and I will invite you to join the Seata developer group

PR #7576, opened 2025-08-06
optimize: Add empty push protection for Configuration

### LINKED ISSUE #7564: Add empty push protection for NacosConfiguration
### Check Ahead

### Ⅰ. Issue Description

https://github.com/apache/incubator-seata/blob/2.x/config/seata-config-nacos/src/main/java/org/apache/seata/config/nacos/NacosConfiguration.java#L445
When Nacos pushes an empty configuration, it will overwrite the current cached configuration. If a user accidentally deletes the configuration content, or if there are compatibility issues between different Nacos versions that result in lost configuration content, Seata may fail to retrieve information such as the transaction group, which can lead to system failures.

### Ⅱ. Describe what happened

_No response_

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

YvCeung:

The configuration centers that Seata interfaces with, such as Nacos, will update the local configuration upon receiving a request for configuration changes. However, Nacos may mistakenly push null values, causing the local configuration to be cleared and thereby affecting business operations. Protective measures need to be taken for this situation

fix #7564 

```shell
mvn clean test
```


YongGoose (changes requested):
It would be good to add a test to verify that the logs are being output correctly.

YvCeung:
@funky-eyes PTAL

funky-eyes (approved):
LGTM
@YongGoose PTAL

config/seata-config-consul/src/test/java/org/apache/seata/config/consul/ConsulConfigurationTest.java
```diff
@@ -131,6 +133,31 @@ void testInitSeataConfig() throws Exception {
...
+        when(mockConsulClient.getKVValue(eq(dataId), isNull(), any(QueryParams.class)))
+                .thenReturn(blankResponse);
+
+        // Run onChangeEvent in a separate thread since it loops indefinitely
+        Thread thread = new Thread(() -> listener.onChangeEvent(new ConfigurationChangeEvent()));
+        thread.start();
+        Thread.sleep(100);
+        // Interrupt to break the loop
+        thread.interrupt();
+        thread.join(500);
```
YongGoose, line 155: What is the purpose of this code?

YvCeung: image
Since this change event of Consul is executed in a continuous loop, if the current thread is not interrupted, the single test will never end. Our goal is to verify the business logic, so we give it 100ms to execute and then interrupt it

YongGoose: But if it's `blank`, doesn't the `return` statement make it exit the while loop?

YvCeung: > But if it's `blank`, doesn't the `return` statement make it exit the while loop?

Yes, you're right. However, here "return" should be changed to "continue". In fact, long polling should not be exited

YongGoose: I also see `return` statements in other configuration code.
And please don’t close the conversation while it’s still ongoing.

YvCeung: > I also see `return` statements in other configuration code. And please don’t close the conversation while it’s still ongoing.

That's because other configuration types passively receive change push notifications. If an empty configuration is received, just return it directly. consul, on the other hand, requires the client to continuously undergo long-term training (active acquisition) at regular intervals.

YongGoose: Ok, I understand
Thank you for your explanation :)

YongGoose (approved):
LGTM

PR #7577, opened 2025-08-07
optimize: remove the 4MB size limit when decompressing with zstd

funky-eyes:

Other compression methods do not have a maximum compression limit, so the compression limit is removed to maintain consistency.
```
Invalid decompressed size: 4917969, the value of size ranges from 0 to 4194304
	at org.apache.seata.compressor.zstd.ZstdUtil.decompress(ZstdUtil.java:44)
```


xingfudeshi (approved):
LGTM

compressor/seata-compressor-zstd/src/main/java/org/apache/seata/compressor/zstd/ZstdUtil.java
```diff
@@ -24,26 +24,18 @@
  */
 public class ZstdUtil {
 
-    public static final int MAX_COMPRESSED_SIZE = 4 * 1024 * 1024;
```
xjlgod, line 27: Why we have this limit before?

PR #7580, opened 2025-08-11
test: fix the exception caused by the disorder of test case execution order

funky-eyes:

image


server/src/test/java/org/apache/seata/server/DynamicPortTestConfig.java
YongGoose, line 1: Is DynamicPortTestConfig working properly by doing this?

changes/zh-cn/2.x.md
```diff
@@ -49,7 +49,8 @@
 - [[#7541](https://github.com/seata/seata/pull/7541)] 修复 jakarta 依赖在 jdk17+ 单测失败问题
 - [[#7540](https://github.com/seata/seata/pull/7540)] 修复mock server端口冲突问题
 - [[#7578](https://github.com/seata/seata/pull/7578)]  zstd解压由jni改为ZstdInputStream
-
+- [[#7580](https://github.com/seata/seata/pull/7580)]  RmNettyRemotingClient#destory时将transactionServiceGroup置为null
+- [[#7580](https://github.com/seata/seata/pull/7580)]  修复测试用例顺序错乱导致的异常
```
YongGoose: This needs to be fixed

server/src/test/java/org/apache/seata/server/spring/listener/ClearServerServicePortInitializer.java
YongGoose, line 1: What is the purpose of this class?

server/src/test/java/org/apache/seata/server/DynamicPortTestConfig.java
funky-eyes: > Is DynamicPortTestConfig working properly by doing this?

When I run tests locally, the logic in DynamicPortTestConfig does not get executed, and the port is still messed up in GitHub CI. Therefore, I used `@TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})` to resolve this issue.

server/src/test/java/org/apache/seata/server/spring/listener/ClearServerServicePortInitializer.java
funky-eyes: > What is the purpose of this class?

Clear the port configuration.

changes/zh-cn/2.x.md
```diff
@@ -49,7 +49,8 @@
 - [[#7541](https://github.com/seata/seata/pull/7541)] 修复 jakarta 依赖在 jdk17+ 单测失败问题
 - [[#7540](https://github.com/seata/seata/pull/7540)] 修复mock server端口冲突问题
 - [[#7578](https://github.com/seata/seata/pull/7578)]  zstd解压由jni改为ZstdInputStream
-
+- [[#7580](https://github.com/seata/seata/pull/7580)]  RmNettyRemotingClient#destory时将transactionServiceGroup置为null
```
```suggestion
```

changes/zh-cn/2.x.md
funky-eyes: done

server/src/test/java/org/apache/seata/server/DynamicPortTestConfig.java
YongGoose: Do we need to apply `@TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})` to all places where `DynamicPortTestConfig` is used?

core/src/main/java/org/apache/seata/core/rpc/netty/RmNettyRemotingClient.java
```diff
@@ -320,6 +320,7 @@ public void destroy() {
         super.destroy();
         initialized.getAndSet(false);
         instance = null;
+        transactionServiceGroup = null;
```
YongGoose, line 323: By the way, is it unnecessary to set `applicationId` to null in the destroy method?

funky-eyes: > By the way, is it unnecessary to set `applicationId` to null in the destroy method?

Since this instance has already been destroyed, I don't think it is necessary to retain the value of transactionServiceGroup. Otherwise, when other test cases are executed, requests may still be sent to the server because transactionServiceGroup still has a value.

server/src/test/java/org/apache/seata/server/DynamicPortTestConfig.java
funky-eyes: > Do we need to apply `@TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})` to all places where `DynamicPortTestConfig` is used?

If DynamicPortTestConfig can be made effective, I don’t think this approach is necessary. My current change is only temporary, intended to prevent CI failures from blocking code review. Therefore, so I do not plan to replace `DynamicPortTestConfig` with `@TestPropertySource(properties = {"server.port=${random.int[10000,20000}"})`.

core/src/main/java/org/apache/seata/core/rpc/netty/RmNettyRemotingClient.java
YongGoose: Yes, I agree with your opinion regarding transactionServiceGroup.
That’s why I was wondering if there’s any reason to keep applicationId :)

server/src/test/java/org/apache/seata/server/DynamicPortTestConfig.java
YongGoose: Or how about creating and applying a custom annotation like the one below?'
Then, We can just add single annotation


```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})
public @interface RandomPortTest {
}
```

funky-eyes: > Or how about creating and applying a custom annotation like the one below?' Then, We can just add single annotation
> 
> ```java
> @Target(ElementType.TYPE)
> @Retention(RetentionPolicy.RUNTIME)
> @TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})
> public @interface RandomPortTest {
> }
> ```

This is also one way to do it.

funky-eyes: > Or how about creating and applying a custom annotation like the one below?' Then, We can just add single annotation
> 
> ```java
> @Target(ElementType.TYPE)
> @Retention(RetentionPolicy.RUNTIME)
> @TestPropertySource(properties = {"server.port=${random.int[10000,20000]}"})
> public @interface RandomPortTest {
> }
> ```

I tried this approach, but `TestPropertySource` simply doesn’t take effect unless the annotation is placed directly on the actual test class. Otherwise, it won’t work if it’s applied through the `RandomPortTest` annotation.

funky-eyes: It was my mistake—the issue was caused by the server.port setting in `application.properties`. After I removed it, everything worked fine. I’ll use `RandomPortTest` instead of `DynamicPortTestConfig`.

funky-eyes: Although the @RandomPortTest annotation does generate a random port, it doesn’t clear the value before and after the test runs, which causes the previous random number to be reused. To address this, I’ve added a BaseSpringBootTest base class as a replacement.

funky-eyes: @YongGoose It's still not working, because all the Spring Boot contexts in the test classes end up being reused, even though the ports are randomized. So I ultimately used `@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)` to ensure that each test case has an isolated context. Although this increases the overall test duration, it seems to be effective.

server/src/test/java/org/apache/seata/server/coordinator/DefaultCoordinatorTest.java
```diff
@@ -110,6 +108,11 @@ public void tearUp() throws IOException {
         deleteAndCreateDataFile();
     }
 
+    @AfterEach
+    public void down() {
+        System.clearProperty(SERVER_SERVICE_PORT_CAMEL);
+    }
```
YongGoose: Wasn't it executed in BaseSpringBootTest?

funky-eyes: > Wasn't it executed in BaseSpringBootTest?

done

YongGoose (approved):
LGTM 🚀

Issue #7561, opened 2025-07-28
Add Dependency Detection and Fallback Logic for OkHttp


YongGoose:
### Check Ahead

- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

- [ ] I am willing to try to implement this feature myself.

### Why you need it?

To make the transition from `HttpClient4` to `OkHttp` smoother, we need to implement dependency detection and fallback logic
- If OkHttp is available → use OkHttp implementation
- If OkHttp is not available → fallback to HttpClient4
- When falling back, log a warning message:

### How it could be?

- Detect whether OkHttp is available on the classpath
- Use OkHttp implementation if available
- If not available, fallback to OkHttp and log a warning message

### Other related information

You can start this task once the PR below has been merged.
- https://github.com/apache/incubator-seata/pull/7492

YvCeung:
PLS assign it to me

YongGoose:
@YvCeung 

Since the dependency has been changed from `HttpClient5` to `OkHttp`, this issue should be updated accordingly. 
I plan to update the issue later this week.

In the meantime, it would be great to take a look at the merged PR and start thinking about how to approach the implementation.
- https://github.com/apache/incubator-seata/pull/7492

YongGoose:
@YvCeung 

As the first step for this PR, I recommend identifying the scenarios where `OkHttp` cannot be used.
It would be helpful to investigate various conditions, such as `JDK` versions and supported `HTTP` versions.

After that, consider whether `HttpClient4` can be used properly in those cases.

YongGoose:
If this is going to take some time, I’d recommend working on #7560 first.

YvCeung:
> [@YvCeung](https://github.com/YvCeung)
> 
> As the first step for this PR, I recommend identifying the scenarios where `OkHttp` cannot be used. It would be helpful to investigate various conditions, such as `JDK` versions and supported `HTTP` versions.
> 
> After that, consider whether `HttpClient4` can be used properly in those cases.

Should I wait until issue is #7544 completed before starting my task?

YongGoose:
> Should I wait until issue is [#7544](https://github.com/apache/incubator-seata/issues/7544) completed before starting my task?

Nope, You don't have to wait :)

YvCeung:
#7904  resolve it

PR #7586, opened 2025-08-15
feature : Integrate Raft and Protocol Switching Based on Server Version

### LINKED ISSUE #7544: Integrate Raft and Protocol Switching Based on Server Version
### Check Ahead

### Why you need it?

Integrate the updated HTTP client into the discovery module’s Raft implementation. The system should detect the server version from metadata and dynamically switch between HTTP/1.1 and HTTP/2 accordingly. Handle cases where the version may be null (e.g., in v2.0).

### How it could be?

_No response_

### Other related information

_No response_

YongGoose:

fixes #7544


YongGoose:
@funky-eyes 

I first select the Node in order to determine whether to use HTTP/2 or HTTP/1, and then extract the address accordingly.

There’s one decision point we need to address.
With HTTP/1, I can process the result after receiving the response, whereas with HTTP/2, the result is handled via a callback.
I’m currently considering how to make the behavior consistent between the two.

If you have any suggestions for a good approach, Don't hesitate to tell me :)

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
```diff
@@ -221,46 +245,61 @@ protected static void startQueryMetadata() {
         }
     }
 
-    private static String queryHttpAddress(String clusterName, String group) {
+    // 그러면 여기에서 string을 반환하는 것이 아니라 node를 반환해야 된다. (node를 반환해서 node의 버전에 따라 결정)
+    // 또는 2.0이 가능한 node를 먼저 선택하게 하는 거는 어떨까?
+    private static Node selectNodeForHttpAddress(String clusterName, String group) {
```
YongGoose, line 226: This method is designed to select the node.

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
```diff
@@ -221,46 +245,61 @@ protected static void startQueryMetadata() {
...
+            List<InetSocketAddress> initAddresses = INIT_ADDRESSES.get(clusterName);
+            if (CollectionUtils.isNotEmpty(initAddresses)) {
+
+                return null;
+            }
+        }
+        return null;
+    }
+
+    private static String queryHttpAddress(String clusterName, Node selectedNode) {
```
YongGoose, line 260: This method extracts the HTTP address based on the selected node and the cluster name.

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
```diff
@@ -418,33 +457,48 @@ private static boolean watch() throws RetryableException {
...
+        return false;
+    }
+
+    private static boolean executeWatchRequest(
+            Node selectedNode, String tcAddress, Map<String, String> param, Map<String, String> header)
+            throws RetryableException {
+        if (selectedNode.isHttp2Supported()) {
+            Http2ClientUtil.doPost("http://" + tcAddress + "/metadata/v1/watch", param, header, CALLBACK);
+            return true; // TODO : Handle the response properly
+        }
```
YongGoose: So that HTTP/2 is used based on the node’s version.

funky-eyes:
> @funky-eyes
> 
> I first select the Node in order to determine whether to use HTTP/2 or HTTP/1, and then extract the address accordingly.
> 
> There’s one decision point we need to address. With HTTP/1, I can process the result after receiving the response, whereas with HTTP/2, the result is handled via a callback. I’m currently considering how to make the behavior consistent between the two.
> 
> If you have any suggestions for a good approach, Don't hesitate to tell me :)

I agree, because during initialization, the server-side API is called to obtain metadata about the members in the Raft group. At this point, it’s possible to retrieve the version number of each member, which can then be used for version switching as needed.

common/src/main/java/org/apache/seata/common/metadata/Node.java
```diff
@@ -99,6 +100,32 @@ public void setVersion(String version) {
         this.version = version;
     }
 
+    @JsonIgnore
+    public boolean isHttp2Supported() {
+        String baseVersion = "2.6.0";
```
funky-eyes, line 105: Could we use the Version class to compare versions instead?

common/src/main/java/org/apache/seata/common/util/Http2ClientUtil.java
```diff
@@ -35,9 +35,9 @@
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-public class Http5ClientUtil {
+public class Http2ClientUtil {
```
funky-eyes, line 38: Could we name it HttpClientUtil? It handles more than just HTTP/2 requests.

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
```diff
@@ -418,36 +433,109 @@ private static boolean watch() throws RetryableException {
...
+                processor.process(null, new RetryableException("Null response"));
+            }
+        } catch (Exception e) {
+            processor.process(null, new RetryableException(e.getMessage(), e));
+        }
+    }
+
+    private static void executeHttp2WatchRequest(
+            String tcAddress, Map<String, String> param, Map<String, String> header, ResponseProcessor processor) {
+        Http2ClientUtil.doPost(
```
funky-eyes, line 495: Since it was changed to HTTP/2 server push, will the watch thread started in startQueryMetadata that runs a while(true) loop become an infinite loop?

common/src/main/java/org/apache/seata/common/util/Http2ClientUtil.java
YvCeung: I think a factory class can be provided, for example, called `HttpClientFactory`, to obtain the specific client instance through the `getClient` method: For instance, if it is detected that http2 is available in the business system, use Http2ClientUtil; otherwise, use the original HttpClientUtil. However, this requires unifying the current apis of these two tool classes

YvCeung: However, recently I have been thinking about one point: whether the http2 function is supported or not not only depends on the business system's reliance on jar packages, but also on whether the `TC` end supports the http2 protocol. Currently, I'm considering whether the judgment logic for the latter can be directly placed in the `getClient` method of `HttpClientFactory ` for judgment. If so, such an implementation would be more elegant. This is what I now solve this issue (https://github.com/apache/incubator-seata/issues/7561)

YongGoose: At first, what I considered was abstraction through an `interface`.
I thought it would be good to dynamically select the class at runtime via abstraction, but since the method return values of the http2 and http utility classes are different, that turned out to be impossible.

So for now, I also think a factory class is the best option. However, unifying the APIs of the two utility classes seems difficult at this point due to differences in their internal logic.

It might be a good idea to revisit this after you finish the work on #7561.
In the meantime, I’ll try to work on this PR over the weekend and wrap it up as quickly as possible.

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
YongGoose: Could I ask you to provide some more explanation? 
I'm not yet sure what exactly is causing concern.

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
```diff
@@ -418,36 +433,109 @@ private static boolean watch() throws RetryableException {
...
+                    }
+
+                    @Override
+                    public void onFailure(Throwable e) {
+                        processor.process(null, e);
+                    }
+
+                    @Override
+                    public void onCancelled() {
+                        processor.process(null, new RetryableException("Request cancelled"));
```
YvCeung, line 514: `RetryableException` is triggered in the `okhttp`  thread. The `watch` method does not throw this exception, which may result in it not being caught by the caller ?

discovery/seata-discovery-raft/src/main/java/org/apache/seata/discovery/registry/raft/RaftRegistryServiceImpl.java
funky-eyes: > Could I ask you to provide some more explanation? I'm not yet sure what exactly is causing concern.

With HTTP/2, configuration changes are delivered by the server via asynchronous pushes. Our current while(true) logic is based on HTTP/1.1 long-polling: the loop blocks synchronously waiting for the server response. Because HTTP/2 uses asynchronous server push, that while(true) no longer provides the blocking behavior and will instead continuously issue watch requests and metadata-fetch requests.

PR #7587, opened 2025-08-17
bugfix: clamp getOrder to avoid tie/overflow

### LINKED ISSUE #7579: UT failed in AdapterSpringSeataInterceptorTest
### Check Ahead

### Ⅰ. Issue Description

https://github.com/apache/incubator-seata/actions/runs/16847550947/job/47731890204

### Ⅱ. Describe what happened

_No response_

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

jihun4452:
Fixes #7579

- [x] I have read the CONTRIBUTING.md guidelines.
- [x] I have registered the PR changes.

- Normalize `AdapterSpringSeataInterceptor#getOrder()`:
  - Prefer the explicit order set via `setOrder(...)` to avoid overflow artifacts from the proxy.
  - Clamp values `>= Ordered.LOWEST_PRECEDENCE` to `Integer.MAX_VALUE - 1` to avoid a tie with Spring’s `TransactionInterceptor`.
- No public API change; only edge-case ordering is adjusted.

Yes.

Added:
- `order_should_be_less_than_lowest_precedence`
- `order_should_passthrough_when_not_lowest`

Run: `./mvnw -pl spring -am -Dtest=AdapterSpringSeataInterceptorTest test -DfailIfNoTests=false`  
Env: macOS, JDK 17, Maven 3.9.x → BUILD SUCCESS locally.

Some environments return an overflowed (negative) order from `ProxyInvocationHandler#getOrder()`, which broke the UT.  
Preferring the explicit order and clamping the max value stabilizes execution precedence and avoids a tie with `TransactionInterceptor`.

### DIFF
--- a/spring/src/main/java/org/apache/seata/spring/annotation/AdapterSpringSeataInterceptor.java
+++ b/spring/src/main/java/org/apache/seata/spring/annotation/AdapterSpringSeataInterceptor.java
@@ -29,7 +29,10 @@ import javax.annotation.Nullable;
 
 public class AdapterSpringSeataInterceptor implements MethodInterceptor, SeataInterceptor, Ordered {
 
-    private ProxyInvocationHandler proxyInvocationHandler;
+    @Nullable
+    private Integer explicitOrder;
+
+    private final ProxyInvocationHandler proxyInvocationHandler;
 
     public AdapterSpringSeataInterceptor(ProxyInvocationHandler proxyInvocationHandler) {
         Assert.notNull(proxyInvocationHandler, "proxyInvocationHandler must not be null");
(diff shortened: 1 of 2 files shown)


YongGoose:
In the code below, the execution order of the `Seata interceptor` has already been adjusted to be one lower than Spring’s `TransactionInterceptor`.
- https://github.com/apache/incubator-seata/blob/b0566047717a485ba0b62e16d8724d50ffe77e6c/spring/src/main/java/org/apache/seata/spring/annotation/GlobalTransactionScanner.java#L449

That’s why I don’t think this change is necessary.

Personally, I also prefer keeping the code clear. 
If we end up having two different `order` values, it might make future maintenance more complicated.

WDYT?

jihun4452:
Thanks for pointing it out. I didn’t realize it was already covered. I’ll close this PR.

PR #7608, opened 2025-08-28
optimize: modify the parameter name in refreshToken method

YvCeung:

I found that when integrating with Seata's built-in NamingServer, the refreshToken method in the register process incorrectly uses the TC address as its parameter.  In fact, the token should be refreshed using the NamingServer address itself.  I fixed this by updating the method signature and calls to refreshToken(String namingServerAddress) to correctly reflect the actual parameter meaning.

对接自身的namingServer时，`refreshToken`方法的形参字面意思上应该是注册中心本身的地址 而不是TC 地址
```java
private static void refreshToken(String tcAddress) throws RetryableException
```


funky-eyes (approved):
LGTM

PR #7635, opened 2025-09-13
test: fix JUnit test method access modifiers and annotations

contrueCT:

This PR corrects JUnit test method access modifiers and annotation usage:

- Fix `@BeforeEach`/`@AfterEach` method visibility in JUnit 5 tests (private → public)
- Fix `@Rule` field access modifier in JUnit 4 tests (private → public)  
- Remove inappropriate `@Test` annotation from utility methods

These changes ensure proper JUnit framework functionality and test execution.


YongGoose (changes requested):
Are there similar cases elsewhere? 
If so, it would be good to address all of them in this PR.

contrueCT:
> Are there similar cases elsewhere? If so, it would be good to address all of them in this PR.

Got it, I will look for similar cases and update the PR accordingly.

contrueCT:
> Are there similar cases elsewhere? If so, it would be good to address all of them in this PR.

Thanks for your suggestion! I’ve checked and fixed all similar cases in the test classes. Could you please take another look when you get a chance?

YongGoose:
> Thanks for your suggestion! I’ve checked and fixed all similar cases in the test classes. Could you please take another look when you get a chance?

Please use JUnit5 instead of JUnit4.

contrueCT:
Done. Replaced all JUnit4 references with JUnit5. Please take another look when you get a chance.

contrueCT:
@YongGoose I've updated the PR based on your feedback. Could you please take another look when you have time? Thanks!

contrueCT:
The CI failure in the seata-config-zk module only occurs in the Java 8 build and is unrelated to my changes. The same tests pass successfully in Java 17 and Java 21, indicating a compatibility issue between ZooKeeper client and Java 8 in the CI environment—not a code problem.

discovery/seata-discovery-etcd3/src/test/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImplTest.java
YoWuwuuuw, line 1: This test class is an integration test that relies on a real etcd service. Because it is marked with `@Disable`, CI will not be triggered. I think it is helpful to include a screenshot of the active triggering test in the description for reference.

contrueCT:
Hi @YongGoose ,  
I've addressed all the feedback you left in your review [https://github.com/apache/incubator-seata/pull/7635#issuecomment-3291173245](url)
Could you please take another look when you have a moment?  
Thanks a lot! 😊

slievrly:
@YongGoose Pls take a look.

discovery/seata-discovery-etcd3/src/test/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImplTest.java
```diff
@@ -118,8 +141,8 @@ public void testSubscribe() throws Exception {
         registryService.register(inetSocketAddress);
         // 2.subscribe
         EtcdListener etcdListener = new EtcdListener();
-        registryService.subscribe(CLUSTER_NAME, etcdListener);
-        // 3.delete instance,see if the listener can be notified
+        registryService.subscribe(DEFAULT_TX_GROUP, etcdListener);
```
YongGoose: Just curious, was there any particular reason for using `DEFAULT_TX_GROUP` instead of CLUSTER_NAME?

YongGoose (approved):
Sorry for the late reply.
LGTM

YongGoose (changes requested):
Please apply spotless :)

Once you’ve fixed the Spotless issues, please ping me and I’ll give my approval.

discovery/seata-discovery-etcd3/src/test/java/org/apache/seata/discovery/registry/etcd3/EtcdRegistryServiceImplTest.java
YongGoose, line 144: Just curious, was there any particular reason for using DEFAULT_TX_GROUP instead of CLUSTER_NAME?

contrueCT: Yes! The subscribe() and lookup() methods expect a transaction service group, not an etcd cluster name.
The original code had mismatched key paths:

- Registration: registry-seata-<DEFAULT_TX_GROUP>/...
- Subscription: registry-seata-<CLUSTER_NAME>/...

Even though both values were "default", using DEFAULT_TX_GROUP is semantically correct and matches how Seata actually works - clients look up services by transaction group.

contrueCT:
I've fixed the Spotless issues, plaese take a look : ) @YongGoose

YongGoose (approved):
LGTM

YongGoose:
@funky-eyes 

I think this pr can be merged now :)

slievrly (approved):
LGTM

Issue #7585, opened 2025-08-14
Error:  io.seata.MockTest.testRollback


funky-eyes:
### Check Ahead

- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

- [ ] I am willing to try to fix this bug myself.

### Ⅰ. Issue Description

```
2025-08-14 06:40:22,842 [SEATA] [main] INFO  io.seata.MockTest - branch register(0.6.1) ok, branchId=9690077789048841
2025-08-14 06:40:22,843 [SEATA] [ServerHandlerThread_1_27_500] INFO  org.apache.seata.mockserver.processor.MockRemotingProcessor - process message : SeataMergeMessage GlobalCommitRequest{xid='172.18.0.1:8077:9690077789048840', extraData='null'}

2025-08-14 06:40:22,845 [SEATA] [rpcDispatch_RMROLE_7_8] INFO  io.seata.core.rpc.netty.RmMessageListener - onMessage:xid=172.18.0.1:8077:9690077789048840,branchId=9690077789048841,branchType=TCC,resourceId=mock-action-061,applicationData={"mock":"mock"}
2025-08-14 06:40:22,845 [SEATA] [rpcDispatch_RMROLE_7_8] INFO  io.seata.rm.AbstractRMHandler - Branch committing: 172.18.0.1:8077:9690077789048840 9690077789048841 mock-action-061 {"mock":"mock"}
commitTcc:172.18.0.1:8077:9690077789048840,null
2025-08-14 06:40:22,845 [SEATA] [rpcDispatch_RMROLE_7_8] INFO  io.seata.rm.AbstractResourceManager - TCC resource commit result :true, xid:172.18.0.1:8077:9690077789048840, branchId:9690077789048841, resourceId:mock-action-061
2025-08-14 06:40:22,845 [SEATA] [rpcDispatch_RMROLE_7_8] INFO  io.seata.rm.AbstractRMHandler - Branch commit result: PhaseTwo_Committed
2025-08-14 06:40:22,845 [SEATA] [rpcDispatch_RMROLE_7_8] INFO  io.seata.core.rpc.netty.RmRpcClient - RmRpcClient sendResponse xid=172.18.0.1:8077:9690077789048840,branchId=9690077789048841,branchStatus=PhaseTwo_Committed,result code =Success,getMsg =null
[...]

PR #7662, opened 2025-09-25
bugfix: ensure visibility of rm and The methods in MockTest are executed in order

### LINKED ISSUE #7558: Test Failure: io.seata.MockTest.testRollback Fails with RPC Timeout
### Check Ahead

### Ⅰ. Issue Description

During the GitHub Actions CI run ([link to job](https://github.com/apache/incubator-seata/actions/runs/16547019230/job/46796366972?pr=7492)), the test io.seata.MockTest.testRollback fails consistently due to an RPC Timeout. The failure log indicates the following error:

```
io.seata.core.exception.TransactionException: RPC Timeout
    at io.seata.MockTest.testRollback ...
```

### Ⅱ. Describe what happened

- Multiple configuration properties under transport.* and server.* are missing in the logs, leading to the use of default values.
- There are repeated errors loading configuration providers (e.g., Nacos provider not found).
- The mock server on 127.0.0.1:8077 may not be properly set up or reachable during the test, leading to Netty channel timeouts and closures.
- The failure always occurs after a 30-second wait, suggesting a network or setup problem rather than a code logic issue.

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

maple525866:

#7585 
With the help of @funky-eyes's analysis, I think it is not a network or port problem, but a visibility problem.
Because the rm member variable in RmClientTest is not declared as volatile, this may cause the rm variable to be modified multiple times during concurrent testing, thus causing this problem。

Here is the fix I made for the problem:
- The volatile keyword ensures that writes to the rm variable are immediately visible to all threads, preventing instruction reordering. 
- Test ordering: The @TestMethodOrder and @Order annotations ensure that tests are executed sequentially, avoiding concurrent access issues. 
- Double-checked locking safety: volatile + synchronized ensures thread-safe lazy loading.

fixes #7558 

mvn clean test

### DIFF
--- a/test-old-version/src/test/java/io/seata/MockTest.java
+++ b/test-old-version/src/test/java/io/seata/MockTest.java
@@ -29,16 +29,14 @@ import io.seata.core.rpc.netty.TmRpcClient;
 import io.seata.rm.DefaultResourceManager;
 import org.apache.seata.mockserver.MockCoordinator;
 import org.apache.seata.mockserver.MockServer;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * the type MockServerTest
  */
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public class MockTest {
 
     static String RESOURCE_ID = "mock-action-061";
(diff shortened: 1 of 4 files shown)


test-old-version/src/test/java/io/seata/MockTest.java
```diff
@@ -29,16 +29,14 @@
 import io.seata.rm.DefaultResourceManager;
 import org.apache.seata.mockserver.MockCoordinator;
 import org.apache.seata.mockserver.MockServer;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
```
funky-eyes: Please don't use wildcard imports

funky-eyes (approved):
LGTM

PR #7674, opened 2025-09-29
feature: implement early rollback of global transactions when TM disconnects

### LINKED ISSUE #4422: rollback of global transactions ahead of time
## Why you need it?
当tm端的channel与tc端断开时，很大可能是因为tm宕机了，如果tm宕机说明tm无法决议事务，但是调用链可能已经到了最后一步，假设这个tm调用了10个微服务，那么在AT下已经锁定了非常多的数据，而默认的timeoutrollback的时间是60秒，tm的宕机会导致数据被锁定60秒不可用，这是比较恶劣的
解决方案参考：
1.通过绑定channelInactive的事件处理，并需要增加开关，因为有可能由于网络问题误判，导致回滚了事务，所以默认开关应该是关闭的，不提前回滚全局事务
2.拥有提前rollback权限的是begin的那个tc，其余tc即便感知到tm断线，也不应该进行rollback，避免出现同时进行rollback或者其中一台tc与tm断开连接导致的误判，所以粒度可以控制在begin的tc上

## How it could be?

kssumin:
- [x] I have registered the PR [changes](https://github.com/apache/incubator-seata/tree/2.x/changes) in `changes/en-us/2.x.md` and `changes/zh-cn/2.x.md`.

## Summary

When a Transaction Manager (TM) client disconnects unexpectedly in a microservice environment, global transactions remain orphaned in "Begin" status until timeout (typically 30+ seconds). This causes unnecessary resource locks, blocking other transactions and degrading system performance during failure scenarios.

This PR introduces fail-fast transaction cleanup by detecting TM disconnections and immediately rolling back orphaned transactions, reducing resource blocking from 30+ seconds to under 1 second.

## Problem

In production microservice deployments, TM clients can disconnect due to:
- Network partitions or instability
- Container restarts in Kubernetes environments
- Process crashes or OOM kills
- Service scaling operations

When this occurs, the current behavior is problematic:

**Current State**: Global transactions started by the disconnected TM remain in "Begin" status, holding:
- Database row locks and table locks
- Connection pool resources
- Memory on TC server for session state

**User Impact**: A 5-second network partition results in 30+ seconds of resource blocking, multiplied across all in-flight transactions. In high-throughput systems processing hundreds of transactions per second, this amplifies temporary failures into prolonged availability issues. Downstream services experience cascading performance degradation as they wait for locks to be released.

**Root Cause**: Seata's transaction lifecycle is designed to react to explicit TM commands (commit/rollback). There is no proactive cleanup mechanism that responds to TM connection state—the only cleanup path is passive timeout waiting.

## Solution

Introduced **connection-aware transaction lifecycle management** that treats TM disconnection as a transaction lifecycle event, not just a network event.

The approach enables fail-fast cleanup through:

**1. Event-Driven Detection**
Hook into existing Netty channel lifecycle to detect TM disconnections in real-time, rather than discovering orphaned transactions only during timeout sweeps.

**2. Conservative Matching Strategy**
Two-level identification prevents false positives:
- **Primary**: Transaction Service Group (VGroup) matching—Seata's standard logical grouping that maps services to TC clusters
- **Secondary**: Application ID validation when available provides additional confidence

Only transactions in "Begin" status are candidates—transactions already in terminal or transitional states (Committing, Rollbacking, etc.) are never touched.

**3. Opt-In Safety Model**
Disabled by default (`server.enableRollbackWhenDisconnect=false`) to ensure zero impact on existing deployments. Users must explicitly enable after validating their transaction patterns.

**4. Standard Rollback Semantics**
Reuses the existing timeout rollback path, transitioning transactions to "TimeoutRollbacking" status before invoking standard cleanup logic. This ensures consistency with normal failure handling and preserves audit trails.

### What Changes for Users

**No Breaking Changes**
- Existing behavior is preserved when disabled (default state)
- No client-side changes required
- No protocol or wire format changes
- Can be safely deployed without configuration changes

**New Capability When Enabled**
Users gain proactive failure recovery:
- **Resource Reclamation**: Locks released in seconds instead of waiting for 30+ second timeouts
- **Fault Isolation**: TM failures don't cascade through lock contention to other services
- **Operational Visibility**: Explicit logging of early rollbacks enables monitoring and alerting on TM health

**Configuration**
```properties
# Server-side only, no client changes needed
server.enableRollbackWhenDisconnect=false  # Default: disabled
```

Closes #4422


YongGoose:
@kssumin 

Where did you handle the `channelInactive` event?

kssumin:
> @kssumin
> 
> Where did you handle the `channelInactive` event?


@YongGoose 
The `channelInactive` event is handled in the **existing** `handleDisconnect()` method. [Location](https://github.com/apache/incubator-seata/blob/5a89e80734ec1a39d6d5215c3d691e6cff02769f/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingServer.java#L194-L226)

**Flow:**
- Netty fires `channelInactive` → calls existing `handleDisconnect()` → we added TM check → calls our handler

YongGoose (approved):
LGTM

PR #7705, opened 2025-10-17
bugfix: ensure environment is fully loaded before ConfigurationFactory initialization

### LINKED ISSUE #7641: seata2.5配置文件加载顺序有问题，导致SpringBootConfigurationProvider#getConfigFromEnvironment无法正常加载到配置
### Check Ahead

### Ⅰ. Issue Description

2.3的是先加载application.properties文件再去加载seata.config.type配置的，正常

Image

2.5的没有加载到就去加载seata.config.type配置了，导致无法正常加载
Image

### Ⅱ. Describe what happened

_No response_

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

WangzJi:
fixes: #7641 


seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/loader/SeataPropertiesLoader.java
```diff
@@ -60,10 +60,7 @@ public class SeataPropertiesLoader implements ApplicationContextInitializer<Conf
     @Override
     public void initialize(ConfigurableApplicationContext applicationContext) {
         ConfigurableEnvironment environment = applicationContext.getEnvironment();
-        if (ObjectHolder.INSTANCE.getObject(OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT) == null) {
-            ObjectHolder.INSTANCE.setObject(
-                    OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT, applicationContext.getEnvironment());
-        }
+        ObjectHolder.INSTANCE.setObject(OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT, environment);
```
funky-eyes, line 63: Can you explain why this change was made?
What's the difference between setting the environment only when it's empty and setting it unconditionally (without checking for emptiness)?

WangzJi: @funky-eyes Thanks for the review!

  The key point: **Line 64 triggers `ConfigurationFactory` static initialization**, which needs a complete environment.

  With the old null check:
  - `EnvironmentPostProcessor` sets environment (without application.properties in Spring Cloud)
  - `ApplicationContextInitializer` skips updating (due to null check)
  - `ConfigurationFactory` gets incomplete environment → can't read `seata.config.type` from application.properties

  Unconditional update ensures:
  - Always use `ApplicationContextInitializer`'s environment (which has all property sources loaded)
  - Safe because it's always more complete than `EnvironmentPostProcessor`'s

  This explains why bootstrap.yml workaround works - it's loaded early enough.

funky-eyes: So, in Spring Cloud there can be multiple Environment instances, which is why the environment may be incomplete. Otherwise, as I understand it, the same instance shouldn't have this problem, right?

psxjoy:
This branch has conflicts that must be resolved, and maybe use `spotless:apply` before merged is a good choice.

seata-spring-autoconfigure/seata-spring-autoconfigure-core/src/main/java/org/apache/seata/spring/boot/autoconfigure/loader/SeataPropertiesLoader.java
WangzJi: In Spring Cloud, it's the same Environment instance throughout, but its PropertySources content evolves over time.
  Timeline:
  1. SeataClientEnvironmentPostProcessor runs (Order: HIGHEST_PRECEDENCE)
     - Sets environment in ObjectHolder
     - At this point: PropertySources = [bootstrap, system]
     - application.yml NOT loaded yet

  2. ConfigDataEnvironmentPostProcessor runs (Order: HIGHEST_PRECEDENCE + 10)
     - Loads application.yml into the same Environment instance
     - Now: PropertySources = [application, bootstrap, system]

PR #7722, opened 2025-10-22
optimize: serializer type meaning

jsbxyyx:


xingfudeshi:
Please register this PR in change logs.

core/src/main/java/org/apache/seata/core/serializer/SerializerType.java
funky-eyes, line 1: If we add a new serializer, what conventions should we follow when assigning its single‑byte identifier?

jsbxyyx: Non-repeating and within byte range

funky-eyes: > Non-repeating and within byte range

I suggest adding a comment documenting the rules for serializer identifiers so that future contributors know how to add new serializers.

jsbxyyx: see: L21-L23

funky-eyes (approved):
LGTM

PR #7752, opened 2025-10-30
optimize: support HTTP/2 push for the Watch API in Server Raft mode

YvCeung:


server/src/main/java/org/apache/seata/server/cluster/manager/ClusterWatcherManager.java
```diff
@@ -107,21 +111,43 @@ private void sendWatcherResponse(Watcher<HttpContext> watcher, HttpResponseStatu
...
+                ctx.writeAndFlush(response);
+            }
+        } else {
+            try {
+
+                Http2Headers headers = new DefaultHttp2Headers().status(nettyStatus.codeAsText());
+                headers.set(HttpHeaderNames.CONTENT_LENGTH, "0");
+
+                ctx.write(new DefaultHttp2HeadersFrame(headers));
+                ctx.write(new DefaultHttp2DataFrame(Unpooled.EMPTY_BUFFER, true));
```
funky-eyes: I want to keep this connection open and reuse it to deliver updates whenever the leader changes, rather than requiring the client to open a new HTTP/2 watch request after every response.

YvCeung: done in [45f362e](https://github.com/apache/incubator-seata/pull/7752/commits/45f362e515cdd6c1598515214654163e1729a120)

PR #7753, opened 2025-10-31
optimize: metadata discovery support for zookeeper

YoWuwuuuw:

Enhance the Zookeeper registry center with metadata registration and discovery capabilities based on branch 'gsoc-2025-meta-registry'.

1. **Register and write metadata:** In `register(ServiceInstance)`, the metadata is serialized into a string and written to the ZK temporary node `data`. The node path remains `.../registry/zk/{cluster}/{ip:port}`.

2. **Lookup/listen and backfill metadata:** In `lookup` and child node change callbacks, `getData()` reads the node `data`, parses it into a `Map<String,String>`, constructs it using `ServiceInstance.fromStringMap(address, metadata)`, and sends the metadata back to the calling side.

3. **Prevent metadata loss during reconnection `recover()`:** A new `REGISTERED_PATH_DATA_MAP` cache is added for `path→data`. When `recover()` rebuilds the temporary node, it uses the cached `data` to write back, ensuring metadata consistency.

1. **Compatibility:** If getData() fails or the node has no data, it will fall back to the address instance `InetSocketAddress` only.
2. **Serialization:** Uses lightweight line-based key=value\n text, with zero dependencies;


xingfudeshi (approved):
LGTM

PR #7756, opened 2025-10-31
optimize: metadata discovery support for redis

YoWuwuuuw:

Enhance the Redis registry center with metadata registration and discovery capabilities based on branch 'gsoc-2025-meta-registry'.

**1. Storage Design**

Survival Key: `registry.redis.{cluster}_{ip:port}` (String, `SETEX` for keep-alive, heartbeat renewal)

Metadata Key: `registry.redis.{cluster}.meta_{ip:port}` (Hash, `HSET` for key -> value, `EXPIRE` has the same TTL as the survival key)

> Compatibility: Write two keys for instances with metadata, write one key for instances without metadata, ensuring normal operation even with client and server version mismatches.

**2. Registration Process**

Write the survival key first (without publishing) → Ensure its existence

Then write the metadata hash and set the TTL (set the TTL even without metadata)

Finally, publish the `REGISTER` event to ensure consumers can read the "with metadata" instance on the first read.

> Heartbeat: During renewal, only refresh the TTL of the two keys, do not repeatedly write metadata.

**3. Lookup/Subscribe:**

- Initial Load: `SCAN registry.redis.{cluster}_*`, directly parse `{ip:port}` from the key name, read the corresponding meta hash, and build... ServiceInstance (with metadata preferred)

- Subscribe to updates: Upon receiving the REGISTER, first delete the old instance at the same address, then add the new instance (with metadata) to remove duplicates and avoid "coexistence of instances with/without metadata at the same address".

### DIFF
--- a/discovery/seata-discovery-redis/src/main/java/org/apache/seata/discovery/registry/redis/RedisRegistryServiceImpl.java
+++ b/discovery/seata-discovery-redis/src/main/java/org/apache/seata/discovery/registry/redis/RedisRegistryServiceImpl.java
@@ -43,13 +43,13 @@ import java.lang.management.ManagementFactory;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


xingfudeshi (approved):
LGTM

PR #7760, opened 2025-11-02
feature: unify Jackson/fastjson serialization. 

### LINKED ISSUE #7554: feature: unify Jackson/fastjson serialization.
### Check Ahead

### Why you need it?

Currently, saga, tcc, at, and seata-server use Jackson/fastjson for serialization. We need to unify it. https://github.com/apache/incubator-seata/issues/6937

### How it could be?

_No response_

### Other related information

_No response_

LegendPei:

unify Jackson/fastjson serialization

Modules involved:
integration-tx-api
seata-saga-statelang
tcc

API interface：
```mermaid
classDiagram
    class JsonSerializer {
        <<interface>>
        +String toJSONString(Object object)
        +T parseObject(String text, Class~T~ clazz)
        +T parseObject(String text, Type type)
    }
    
    class JsonUtil {
        -JsonSerializer serializer
        +JsonUtil(JsonSerializer serializer)
        +String toJSONString(Object object)
        +T parseObject(String text, Class~T~ clazz)
        +T parseObject(String text, Type type)
        +JsonSerializer getSerializer()
    }
    
    class JsonSerializerFactory {
        <<static>>
        +JsonSerializer getSerializer(String name)
    }
    
    class FastjsonJsonSerializer {
        +String toJSONString(Object object)
        +T parseObject(String text, Class~T~ clazz)
        +T parseObject(String text, Type type)
        +String toJSONString(Object object, boolean prettyPrint)
        +String toJSONString(Object object, boolean ignoreAutoType, boolean prettyPrint)
        +T parseObject(String text, Class~T~ type, boolean ignoreAutoType)
        +boolean useAutoType(String json)
    }
    
    class JacksonJsonSerializer {
        +String toJSONString(Object object)
        +T parseObject(String text, Class~T~ clazz)
        +T parseObject(String text, Type type)
        +String toJSONString(Object object, boolean prettyPrint)
        +String toJSONString(Object object, boolean ignoreAutoType, boolean prettyPrint)
        +T parseObject(String text, Class~T~ type, boolean ignoreAutoType)
        +boolean useAutoType(String json)
    }
    
    class GsonJsonSerializer {
        +String toJSONString(Object object)
        +T parseObject(String text, Class~T~ clazz)
        +T parseObject(String text, Type type)
        +String toJSONString(Object object, boolean prettyPrint)
        +String toJSONString(Object object, boolean ignoreAutoType, boolean prettyPrint)
        +T parseObject(String text, Class~T~ type, boolean ignoreAutoType)
        +boolean useAutoType(String json)
    }
    
    JsonUtil --> JsonSerializer : uses
    JsonSerializerFactory --> JsonSerializer : creates
    FastjsonJsonSerializer ..|> JsonSerializer : implements
    JacksonJsonSerializer ..|> JsonSerializer : implements  
    GsonJsonSerializer ..|> JsonSerializer : implements
```

Use flowcharts：
mermaid-diagram-2025-11-02-201407

fixes #7554


maple525866:
Are you developing based on an old branch? It would be better to use a new branch next time; otherwise, there will be many commits that don't belong to this pull request, which is detrimental to CR

LegendPei:
> Are you developing based on an old branch? It would be better to use a new branch next time; otherwise, there will be many commits that don't belong to this pull request, which is detrimental to CR您是否基于旧分支进行开发？下次最好使用一个新的分支；否则，将有许多提交不属于这个pull请求，这对CR是有害的

Yes, sorry, I forgot to reset when I created the new branch earlier, so there are some unnecessary commits : (

LegendPei:
I’d really appreciate any feedback or suggestions from the community. orz
I’m currently working on improving test coverage for this patch, and if all goes well, I plan to gradually replace the existing JsonUtil usages in other modules as well :)

common/src/main/java/org/apache/seata/common/json/JsonUtil.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private final JsonSerializer serializer;
+
+    /**
+     * Create a JsonUtil instance with the specified JsonSerializer implementation
+     *
+     * @param serializer the JsonSerializer implementation to use
+     * @throws IllegalArgumentException if serializer is null
+     */
+    public JsonUtil(JsonSerializer serializer) {
```
YvCeung: It might be better to make the constructor of `JsonUtils` private to prevent multiple instances from being created, since the class is intended to be used as a static utility.

LegendPei: Okay, thank you. I'll start modifying the code now

LegendPei: I'm sorry, but I don't quite understand why this modification is necessary. Isn't it possible to use different JSON parsers (fastjson or jackson) in different modules based on their specific requirements? Why not keep JsonUtil as an instantiable class, allowing each module to create its own instance using the appropriate JsonSerializer implementation?Or am I misunderstanding the intention of this unified serializer?

YvCeung: `JsonUtil` 内部确实维护了单例的 JsonSerializer， 但 `JsonUtil` 本身却是可以多次 new 的。 所以从设计角度看，它并不比直接通过工厂类 `JsonSerializerFactory.getSerializer(xxx)` 获得实例更简洁。 与其让大家都 new 一个“工具类”，不如直接让业务层通过工厂方法获取单例实例。❤

JsonUtil internally maintains a singleton JsonSerializer,
but JsonUtil itself can still be instantiated multiple times.
From a design perspective, this makes it no simpler than directly obtaining an instance via JsonSerializerFactory.getSerializer().
Instead of having everyone create new instances of this utility class, it would be cleaner and more aligned with the idea of a unified serialization entry point to let the business layer directly acquire the singleton instance through the factory method.

LegendPei: 哦，我明白了，是不是这个JsonUtil应该作为一个全局统一的Json解析器，比如统一所有的json解析的底层都为fastjson实现；如果某个模块需要使用到其他的json解析如jackson，那就应该直接使用 JsonSerializer 再通过JsonSerializerFactory.getSerializer 来获取实例，这样的话就不用多次 new JsonUtil 🤔 感谢你的指点 orz

Oh, I understand. Shouldn't this JsonUtil be used as a globally unified JSON parser, for example, unifying all the underlying JSON parsing implementations to fastjson? If a certain module needs to use other JSON parsers such as Jackson, then it should directly use JsonSerializer and obtain an instance through JsonSerializerFactory.getSerializer. In this way, there is no need to repeatedly new JsonUtil 🤔Thanks for your guidance orz

common/src/main/java/org/apache/seata/common/json/JsonUtil.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Get the underlying JsonSerializer instance
+     *
+     * @return the JsonSerializer instance
+     */
+    public JsonSerializer getSerializer() {
+        return serializer;
+    }
```
maple525866: 工具类应该专职于工具类的职责，而不应该能返回实例，这侵占了工厂类JsonSerializerFactory的职责，你觉得呢？同时也不符合良好的设计模式

I feel there's a problem here. Utility classes should focus on their utility responsibilities and shouldn't be able to return instances. This encroaches on the responsibilities of the factory class, JsonSerializerFactory. What do you think?It also doesn't conform to good design patterns.

LegendPei: 好的谢谢你，我会把它删掉的❤

Okay, thank you. I will delete it❤

json-common/src/main/java/org/apache/seata/common/json/JsonUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Serialize the given object to JSON string
+     *
+     * @param object the object to serialize
+     * @return the JSON string representation
+     * @throws JsonParseException if serialization fails
+     */
+    public static String toJSONString(Object object) {
+        return JsonSerializerFactory.getSerializer(CONFIG_JSON_PARSER_NAME).toJSONString(object);
```
funky-eyes: I suggest defining a static variable to avoid the extra performance overhead of repeatedly calling JsonSerializerFactory.getSerializer(CONFIG_JSON_PARSER_NAME) to obtain an instance.

json-common/src/main/java/org/apache/seata/common/json/impl/FastjsonJsonSerializer.java
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    private static final SerializerFeature[] FEATURES_PRETTY = new SerializerFeature[] {
+        SerializerFeature.DisableCircularReferenceDetect,
+        SerializerFeature.WriteDateUseDateFormat,
+        SerializerFeature.PrettyFormat
+    };
+
+    private static final Feature[] READER_FEATURES_SUPPORT_AUTO_TYPE =
+            new Feature[] {Feature.SupportAutoType, Feature.OrderedField};
```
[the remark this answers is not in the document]
LegendPei: It's best not to make such modifications to maintain maximum compatibility with the original JSON tool

json-common/src/test/java/org/apache/seata/common/json/GsonJsonSerializerTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Test
+    public void testUseAutoType_withoutType() {
+        String json = "{\"name\":\"test\"}";
+        boolean hasAutoType = ((GsonJsonSerializer) jsonSerializer).useAutoType(json);
+        assertThat(hasAutoType).isFalse();
+    }
+
+    @Test
+    public void testToJSONString_prettyPrint() {
```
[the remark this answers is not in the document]
LegendPei: 'testToJSONString_prettyPrint' is correct because it tests the toJSONString method

funky-eyes (commented):
Additionally, I think we need a global configuration option for trusted classes whose values are patterns like java.*, cn.xxx.xx, com.xxx.* — in other words, a whitelist of allowed classes to prevent deserialization vulnerabilities. I’m not sure whether this should be implemented in this PR, but at minimum the whitelist feature must be merged immediately after this PR is accepted; otherwise it would constitute a security risk. @slievrly

json-common/src/main/java/org/apache/seata/common/json/JsonUtil.java
LegendPei: 先缓存默认解析器实例怎么样，这样的话如果使用的是默认的解析器，就直接调用，不然就通过工厂生成，并且工厂内部已使用 ConcurrentHashMap 缓存实例，工厂内的 INSTANCES 会确保同一个 name 只会创建一次 JsonSerializer，后续调用直接返回缓存实例

How about caching the default parser instance first? This way, if the default parser is used, it can be directly invoked. Otherwise, it will be generated through a factory, and the factory internally uses a ConcurrentHashMap to cache instances. The INSTANCES within the factory will ensure that only one JsonSerializer will be created for the same name, and subsequent calls will directly return the cached instance

funky-eyes: > 先缓存默认解析器实例怎么样，这样的话如果使用的是默认的解析器，就直接调用，不然就通过工厂生成，并且工厂内部已使用 ConcurrentHashMap 缓存实例，工厂内的 INSTANCES 会确保同一个 name 只会创建一次 JsonSerializer，后续调用直接返回缓存实例
> 
> How about caching the default parser instance first? This way, if the default parser is used, it can be directly invoked. Otherwise, it will be generated through a factory, and the factory internally uses a ConcurrentHashMap to cache instances. The INSTANCES within the factory will ensure that only one JsonSerializer will be created for the same name, and subsequent calls will directly return the cached instance

However, there is overhead from map.get(), so you can simply assign JsonSerializerFactory.getSerializer(CONFIG_JSON_PARSER_NAME) to a member variable.

saga/pom.xml
```diff
@@ -48,6 +48,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
funky-eyes, line 51: Since the json-common module already depends on fastjson, can we remove the separately declared fastjson dependency from the saga module?

funky-eyes (commented):
There are also some places that instantiate ObjectMapper directly; these should be replaced with JsonSerializer for JSON serialization and deserialization.

saga/pom.xml
LegendPei: 应该可以，我先再次检查saga模块是否还有用到fastjson的地方吧，然后再把那些直接使用objectmapper的地方替换掉先

It should be possible. I'll first double-check if there are any places in the saga module where fastjson is still being used, and then replace those direct uses of objectmapper

json-common/src/main/java/org/apache/seata/common/json/JsonUtil.java
LegendPei: 哦我懂您的意思了，应该在静态初始化时就获取并缓存默认解析器实例，谢谢您的指点，我现在就改一下

Oh, I understand what you mean. We should obtain and cache the default parser instance during static initialization. Thank you for your guidance. I'll make the change now

saga/pom.xml
LegendPei: 然后就是有些ObjectMapper是涉及到undolog和一些协议上的序列化，这些应该不用换吧；而且我不是很知道有哪些ObjectMapper需要替换的，因为我看它这个ObjectMapper很多都是配合着一些ObjectNode啊这些jackson里面的类用的

Then, some ObjectMapper involves undolog and serialization on some protocols, which should not be replaced; moreover, I don't quite know which ObjectMapper needs to be replaced, because I see that many of them are used in conjunction with ObjectNode and other classes in Jackson

funky-eyes: > 然后就是有些ObjectMapper是涉及到undolog和一些协议上的序列化，这些应该不用换吧；而且我不是很知道有哪些ObjectMapper需要替换的，因为我看它这个ObjectMapper很多都是配合着一些ObjectNode啊这些jackson里面的类用的
> 
> Then, some ObjectMapper involves undolog and serialization on some protocols, which should not be replaced; moreover, I don't quite know which ObjectMapper needs to be replaced, because I see that many of them are used in conjunction with ObjectNode and other classes in Jackson

The serialization package and anything under the serialization module do not need to be replaced; all other code should be updated to use the unified JSON serialization entry point.

LegendPei: 我需要一点时间来检查替换objectmapper，因为涉及到objectmapper的代码比较多；然后就是saga模块的fastjson依赖是可以删除的，因为只有`FastjsonParser`这个类使用了fastjson，把saga原先的json解析器删除后才能移除这个依赖

I need some time to check and replace the ObjectMapper, as there is a significant amount of code involving ObjectMapper; furthermore, the dependency on fastjson in the saga module can be removed, as only the class `FastjsonParser` uses fastjson. This dependency can only be removed after deleting the original json parser in saga

json-common/src/main/java/org/apache/seata/common/json/JsonSerializerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.seata.common.loader.EnhancedServiceLoader;
+import org.apache.seata.common.util.CollectionUtils;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class JsonSerializerFactory {
+
+    private static final String DEFAULT_SERIALIZER = "fastjson";
```
funky-eyes: I recommend using Jackson as the default, as it is generally safer.

json-common/src/main/java/org/apache/seata/common/json/JsonSerializerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class JsonSerializerFactory {
+
+    private static final String DEFAULT_SERIALIZER = "jackson";
+
+    private static final Map<String, org.apache.seata.common.json.JsonSerializer> INSTANCES = new ConcurrentHashMap<>();
```
maple525866: Remove "org.apache.seata.common.json." and replace it with "import".

json-common/src/main/java/org/apache/seata/common/json/JsonSerializerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Get JsonSerializer instance by name.
+     *
+     * @param name the serializer name (e.g., "fastjson", "jackson", "gson")
+     * @return the JsonSerializer instance
+     */
+    public static org.apache.seata.common.json.JsonSerializer getSerializer(String name) {
+        final String serializerName = Optional.ofNullable(name).orElse(DEFAULT_SERIALIZER);
+        return CollectionUtils.computeIfAbsent(
```
maple525866, line 42: ditto, and check other changed files for any other similar issues.

maple525866 (approved):
LGTM

funky-eyes (approved):
LGTM

changes/zh-cn/2.x.md
```diff
@@ -20,6 +20,8 @@
 
 ### feature:
 
+- [[#7760](https://github.com/apache/incubator-seata/pull/7760)] 统一了Jackson/fastjson序列化器
```
```suggestion
- [[#7760](https://github.com/apache/incubator-seata/pull/7760)] 统一Jackson/fastjson序列化器
```

PR #7765, opened 2025-11-04
optimize: Optimize RM/TM client reconnect logic

### LINKED ISSUE #5338: Is it necessary that both the RM client and the TM client start a timer for reconnecting to seata server ？

### Ⅰ. Issue Description
Is it necessary that both rm client and tm client start a timer for reconnecting to seata server？

The log of 2 timeoutCheckers is as follows：
2023-02-13 10:12:51.547  INFO 4079 --- [eoutChecker_1_1] i.s.d.r.k8s.K8sRegistryServiceImpl
2023-02-13 10:12:51.917  INFO 4079 --- [eoutChecker_2_1] i.s.d.r.k8s.K8sRegistryServiceImpl 
2023-02-13 10:13:01.547  INFO 4079 --- [eoutChecker_1_1] i.s.d.r.k8s.K8sRegistryServiceImpl 
2023-02-13 10:13:01.916  INFO 4079 --- [eoutChecker_2_1] i.s.d.r.k8s.K8sRegistryServiceImpl 

```
io.seata.core.rpc.netty.AbstractNettyRemotingClient#init
public void init() {
        timerExecutor.scheduleAtFixedRate(new Runnable() {
            @Override
            public void run() {
                clientChannelManager.reconnect(getTransactionServiceGroup());
            }
        }, SCHEDULE_DELAY_MILLS, SCHEDULE_INTERVAL_MILLS, TimeUnit.MILLISECONDS);
```

lokidundun:

- Optimize RM/TM client reconnect logic
- Replaces individual reconnect timers with a shared global timer for all RM/TM clients
- Maintains separate channel management for RM and TM while optimizing timer resources

fixes https://github.com/apache/incubator-seata/issues/5338


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -155,6 +157,28 @@ public AbstractNettyRemotingClient(
         clientBootstrap.setChannelHandlers(new ClientHandler(), new ChannelEventHandler(this));
         clientChannelManager = new NettyClientChannelManager(
                 new NettyPoolableFactory(this, clientBootstrap), getPoolKeyFunction(), nettyClientConfig);
+
+        Configuration configuration = ConfigurationFactory.getInstance();
+        this.enableReconnect =
+                configuration.getBoolean("client.reconnect.enable", DefaultValues.DEFAULT_ENABLE_CLIENT_RECONNECT);
+
```
YvCeung: Although there is currently a passive reconnection mechanism (on-demand reconnection), it inevitably increases request latency, as the connection is only detected and re-established when a request is actually sent.  I would like to understand: if the scheduled reconnection switch is disabled, are there any other negative impacts beyond the increased request latency?  Additionally, is it necessary to give users the decision-making authority over this switch?

lokidundun: I haven't fully figured this out yet. Let me think about it more. Thank you for your suggestion, and I will value it seriously.

lokidundun: I agree that exposing this switch to users might introduce unnecessary complexity，I will remove it,Thank you for your insightful feedback—it really helped refine the design.

lokidundun:
@YvCeung  PTAL❤️

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
maple525866, line 321: Is there a risk of resource leaks here? Currently, `destroy()` only removes the instance from `CLIENT_INSTANCES`, but does not stop `GLOBAL_RECONNECT_TIMER` when there are no instances. Even if all clients are destroyed, the scheduler remains alive and occupies threads.

maple525866:
At the same time, you should also increase testing to verify the correctness of the changes. @lokidundun

lokidundun:
> At the same time, you should also increase testing to verify the correctness of the changes. @lokidundun

Thanks for the hint! I've added shutdownGlobalTimerIfNoClients() in destroy(), wired up the global timer/worker lifecycle and added tests covering the path. I'll further harden shutdown/wait logic and concurrency handling — please take another look when you have time.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -773,4 +815,109 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+        if (CLIENT_INSTANCES.isEmpty() && GLOBAL_TIMER_STARTED.compareAndSet(true, false)) {
+            ScheduledExecutorService timer = GLOBAL_RECONNECT_TIMER_REF.getAndSet(null);
+            if (timer != null) {
+                try {
+                    timer.shutdown();
+                    timer.awaitTermination(1, TimeUnit.SECONDS);
+                } catch (Exception ex) {
+                    LOGGER.warn("Shutdown global reconnect timer failed: {}", ex.getMessage(), ex);
+                }
+            }
```
maple525866: Currently, `shutdownGlobalTimerIfNoClients()` appears to have a race condition. There's a window of opportunity: if another thread immediately creates a client and calls `add(this)` after `isEmpty()` returns true, the timer and worker will be stopped, but there will still be an active client. As a result, the new client won't be reconnected by the timer.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -773,4 +815,109 @@ public void close(ChannelHandlerContext ctx, ChannelPromise future) throws Excep
...
+            ExecutorService worker = GLOBAL_RECONNECT_WORKER_REF.getAndSet(null);
+            if (worker != null) {
+                try {
+                    worker.shutdown();
+                    worker.awaitTermination(1, TimeUnit.SECONDS);
```
maple525866: Logically speaking, the `worker` also has this problem.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lokidundun: Thank you for identifying the race condition. I have tried to resolve it by introducing a global lock and a double-check mechanism to ensure atomicity when checking client instances and shutting down resources.
Key Fixes:
- Added private static final ReentrantLock GLOBAL_SHUTDOWN_LOCK to synchronize the entire operation of checking CLIENT_INSTANCES.isEmpty() and shutting down the timer/worker.

Please take another look when you have time❤️.

lokidundun:
maybe this pr is turning into comlexity,I will close it and open a new one

PR #7783, opened 2025-11-14
optimize: Optimize RM/TM client reconnect 

### LINKED ISSUE #5338: Is it necessary that both the RM client and the TM client start a timer for reconnecting to seata server ？

lokidundun:

Optimized RM/TM client reconnect logic to prevent duplicate timer creation .

fixes https://github.com/apache/incubator-seata/issues/5338


lokidundun:
I've addressed the resource leak issue by removing the static modifier from the reconnect timer and related resources.Am I on the right track?Let me know if further adjustments are needed!If you are convenient,please take a look❤️. @maple525866

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -271,6 +287,11 @@ public void destroyChannel(String serverAddress, Channel channel) {
 
     @Override
     public void destroy() {
+        if (reconnectTimer != null && timerStarted.get()) {
+            reconnectTimer.shutdown();
```
maple525866: When reconnectTimer runs shutdown(), it does not interrupt ongoing tasks, which means that reconnection may still occur after destruction.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -96,6 +99,10 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting
     protected final Condition mergeCondition = mergeLock.newCondition();
     protected volatile boolean isSending = false;
 
+    private final Runnable reconnectTask;
+    private ScheduledExecutorService reconnectTimer;
```
maple525866: It's probably better to have a globally unique one here, otherwise there would be multiple reconnectTimers.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
maple525866: Aside from the reconnection issue, logically speaking, there is also a race condition problem here.

lokidundun: > When reconnectTimer runs shutdown(), it does not interrupt ongoing tasks, which means that reconnection may still occur after destruction.

It means that I should use the `reconnectTimer.shutdownNow()` not the `reconnectTimer.shutdown()` to stop all active tasks,right? 

> Aside from the reconnection issue, logically speaking, there is also a race condition problem here.

It means that I ought to use ` protected final ReentrantLock mergeLock = new ReentrantLock()` to promise the Atomicity,like this?
```
public void destroy() {
    mergeLock.lock();
    try {
        if (reconnectTimer != null && timerStarted.get()) {
            reconnectTimer.shutdown();
            timerStarted.set(false);
            LOGGER.info("Instance reconnect timer stopped (role: {})", transactionRole.name());
        }
        clientBootstrap.shutdown();
        if (mergeSendExecutorService != null) {
            mergeSendExecutorService.shutdown();
        }
        super.destroy();
    } finally {
        mergeLock.unlock();
    }
}
```

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lokidundun: > It's probably better to have a globally unique one here, otherwise there would be multiple reconnectTimers.

I haven't come up with a perfect solution for now. Could you please share your thoughts?❤️

lokidundun: > It's probably better to have a globally unique one here, otherwise there would be multiple reconnectTimers.

Or I think we can tackle like this? What do you think?
```
private volatile ScheduledExecutorService reconnectTimer;
    @Override
    public void init() {
        if (timerStarted.compareAndSet(false, true)) {
            mergeLock.lock();
            try {
                this.reconnectTimer = new ScheduledThreadPoolExecutor(
                        1, new NamedThreadFactory("Reconnect-Timer-" + transactionRole.name(), 1));
                this.reconnectTimer.scheduleAtFixedRate(
                        this.reconnectTask, SCHEDULE_DELAY_MILLS, SCHEDULE_INTERVAL_MILLS, TimeUnit.MILLISECONDS);
                LOGGER.info("Instance reconnect timer started (role: {})", transactionRole.name());
            }finally {
                mergeLock.unlock();
            }
        }
        if (this.isEnableClientBatchSendRequest()) {
            mergeSendExecutorService = new ThreadPoolExecutor(
                    MAX_MERGE_SEND_THREAD,
                    MAX_MERGE_SEND_THREAD,
                    KEEP_ALIVE_TIME,
                    TimeUnit.MILLISECONDS,
                    new LinkedBlockingQueue<>(),
                    new NamedThreadFactory(getThreadPrefix(), MAX_MERGE_SEND_THREAD));
            mergeSendExecutorService.submit(new MergedSendRunnable());
        }
        super.init();
        clientBootstrap.start();
    }
```

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -120,17 +127,18 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting
...
-                SCHEDULE_DELAY_MILLS,
-                SCHEDULE_INTERVAL_MILLS,
-                TimeUnit.MILLISECONDS);
+        if (timerStarted.compareAndSet(false, true)) {
+            mergeLock.lock();
+            try {
+                this.reconnectTimer = new ScheduledThreadPoolExecutor(
+                        1, new NamedThreadFactory("Reconnect-Timer-" + transactionRole.name(), 1));
+                this.reconnectTimer.scheduleAtFixedRate(
+                        this.reconnectTask, SCHEDULE_DELAY_MILLS, SCHEDULE_INTERVAL_MILLS, TimeUnit.MILLISECONDS);
```
funky-eyes: reconnectTask is still null at this point; you should initialize reconnectTask before creating the thread pool. Also, why not reuse timerExecutor instead of creating a new scheduled thread pool?

lokidundun: Thanks for your detailed feedback! Here are the optimizations I've received and made in response to your comments:
I've initialized `reconnectTask `in the constructor in advance and adjusted the logic to reuse the existing `timerExecutor`
```
    @Override
    public void init() {
        mergeLock.lock();
        try {
            if (timerStarted.compareAndSet(false, true)) {
                timerExecutor.scheduleAtFixedRate(
                        reconnectTask, SCHEDULE_DELAY_MILLS, SCHEDULE_INTERVAL_MILLS, TimeUnit.MILLISECONDS);
                LOGGER.info("Reconnect timer started (role: {})", transactionRole.name());
            }

            if (this.isEnableClientBatchSendRequest()) {
                mergeSendExecutorService = new ThreadPoolExecutor(
                        MAX_MERGE_SEND_THREAD,
                        MAX_MERGE_SEND_THREAD,
                        KEEP_ALIVE_TIME,
                        TimeUnit.MILLISECONDS,
                        new LinkedBlockingQueue<>(),
                        new NamedThreadFactory(getThreadPrefix(), MAX_MERGE_SEND_THREAD));
                mergeSendExecutorService.submit(new MergedSendRunnable());
            }

            super.init();
            clientBootstrap.start();
        } finally {
            mergeLock.unlock();
        }
    }

    public AbstractNettyRemotingClient(
            NettyClientConfig nettyClientConfig,
            ThreadPoolExecutor messageExecutor,
            NettyPoolKey.TransactionRole transactionRole) {
        super(messageExecutor);
        this.transactionRole = transactionRole;
        clientBootstrap = new NettyClientBootstrap(nettyClientConfig, transactionRole);
        clientBootstrap.setChannelHandlers(new ClientHandler(), new ChannelEventHandler(this));
        clientChannelManager = new NettyClientChannelManager(
                new NettyPoolableFactory(this, clientBootstrap), getPoolKeyFunction(), nettyClientConfig);
        this.reconnectTask = () -> {
            if (!timerStarted.get()) {
                return;
            }
            try {
                String serviceGroup = getTransactionServiceGroup();
                if (StringUtils.isNotBlank(serviceGroup)) {
                    clientChannelManager.reconnect(serviceGroup);
                }
            } catch (Throwable t) {
                LOGGER.error("Reconnect task failed for role: {}", transactionRole.name(), t);
            }
        };
    }
```
Am I on the right way.I’d really appreciate any feedback and let me know if you have any further questions!

funky-eyes: This looks much better. Good job.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -120,29 +124,30 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting
...
-                    MAX_MERGE_SEND_THREAD,
-                    KEEP_ALIVE_TIME,
-                    TimeUnit.MILLISECONDS,
-                    new LinkedBlockingQueue<>(),
-                    new NamedThreadFactory(getThreadPrefix(), MAX_MERGE_SEND_THREAD));
-            mergeSendExecutorService.submit(new MergedSendRunnable());
+        mergeLock.lock();
+        try {
+            if (timerStarted.compareAndSet(false, true)) {
+                timerExecutor.scheduleAtFixedRate(
```
maple525866: I don't think `mergeLock` should be used here, because it's originally designed for merging and sending data; not using it can ensure single responsibility.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -271,11 +289,19 @@ public void destroyChannel(String serverAddress, Channel channel) {
 
     @Override
     public void destroy() {
-        clientBootstrap.shutdown();
-        if (mergeSendExecutorService != null) {
-            mergeSendExecutorService.shutdown();
+        mergeLock.lock();
+        try {
+            if (timerStarted.compareAndSet(true, false)) {
+                LOGGER.info("Reconnect timer stopped (role: {})", transactionRole.name());
```
maple525866: ditto

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
lokidundun: That means I need to use another like `private final ReentrantLock reconnectLock = new ReentrantLock();` instead of `mergeLock` .Am I on the right way.I’d really appreciate any feedback and let me know if you have any further questions!

maple525866: Yes, it needs to be replaced with another lock instead of using the original `mergeLock`.

lokidundun:
Could you please take another look ❤️ @funky-eyes

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -96,6 +98,11 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting
     protected final Condition mergeCondition = mergeLock.newCondition();
     protected volatile boolean isSending = false;
 
+    private final Runnable reconnectTask;
+    private AtomicBoolean timerStarted = new AtomicBoolean(false);
```
funky-eyes: I'm not quite clear on the purpose of this PR. Currently, `timerStarted` is an instance variable and not shared, which means both RM and TM will each create their own `reconnectTask`. The `reconnectTask` itself is also an instance variable with no sharing.

For the old logic, what is the point of adding `timerStarted`? Wouldn't it be better to unify the reconnect thread pool for both RM and TM into a single one, so that there is only one task per entity (one for RM and one for TM) instead of starting two separate thread pools?

PR #7794, opened 2025-11-19
support @GlobalTransactional and @GlobalLock in parent class and interface

### LINKED ISSUE #7793: support @GlobalTransactional and @GlobalLock in parent class and interface
### Check Ahead

### Why you need it?

for Interface-Oriented Programming, we want only add `@GlobalTransactional` and `@GlobalLock` on interface, not on the actual class

### How it could be?

_No response_

### Other related information

_No response_

whaon:

support `@GlobalTransactional` and `@GlobalLock` in parent class and interface

fixes #7793


common/src/main/java/org/apache/seata/common/util/ReflectionUtil.java
```diff
@@ -705,6 +705,45 @@ public static Map<String, Object> getAnnotationValues(Annotation annotation) thr
...
+    /**
+     *
+     * @param targetClass the target class
+     * @param methodName the method name
+     * @param paramTypes the param types
+     * @param annotationClass the annotation class
+     * @return the annotation
+     * @param <A>
+     */
+    public static <A extends Annotation> A findAnnotationInHierarchy(
```
funky-eyes, line 743: I believe method-level annotation data should take precedence. Only if a method does not have that annotation should we check the corresponding method in the superclass. This ensures that if the same annotation appears with different configured values, the value on the method is the one that ultimately applies.

whaon: you are right! the invoker will check the actual class first just as you wish
code is here:
```java
    private <A extends Annotation> A findAnnotation(Method method, Class<?> targetClass, Class<A> annotationClass) {
        A anno = getAnnotation(method, targetClass, annotationClass);
        if (anno == null) {
            anno = ReflectionUtil.findAnnotationInHierarchy(
                    targetClass, method.getName(), method.getParameterTypes(), annotationClass);
        }
        return anno;
    }
```

common/src/main/java/org/apache/seata/common/util/ReflectionUtil.java
```diff
@@ -705,6 +705,45 @@ public static Map<String, Object> getAnnotationValues(Annotation annotation) thr
...
+            Class<?> targetClass, String methodName, Class<?>[] paramTypes, Class<A> annotationClass) {
+        Class<?> superClass = targetClass.getSuperclass();
+        while (superClass != null && superClass != Object.class) {
+            try {
+                Method superMethod = superClass.getDeclaredMethod(methodName, paramTypes);
```
funky-eyes, line 796: Should we attempt to read the annotation from the superclass instead of only checking the method itself?

whaon: update, pls have a check

PR #7815, opened 2025-11-28
test: fix combine test to avoid failure due to ordering

cperkkk:

This pull request addresses several flaky test, in the `apache/incubator-seata` project. The test was failing 
intermittently due to assumptions about the ordering of the test. I've noticed that when running the test in specific ordering will produce a test failure (maven surefire plugin unfortunately does not guarantee test ordering), such as:

```
mvn test -Dtest=org.apache.seata.config.AbstractConfigurationTest#AtestGetConfigWithNull+BtestGetConfigWithDefaultValue
```

The tests will fail with output:

```
ERROR] Failures:
[ERROR]   AbstractConfigurationTest.AtestGetConfigWithNull:152 expected: not <null>
```

This is due to the necessity of the cache being set by `testGetConfigWithDefaultValue` first, before the eventual `testGetConfigWithNull` running for it to pass. So there's an assumed ordering but the test fails to explicitly guarantee the ordering (different version of maven surefire will run the order randomly and if the test are refactored / renamed, the current test order will change and it will fail), combining the test will help with removing this flakiness.

It does not fixes a particular github issue.

I've only changed test and not code under test

We test the combined version

```
mvn test -Dtest=org.apache.seata.config.AbstractConfigurationTest#testGetConfigWithDefaultValue
```

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

- The fix ensures the test is robust to non-deterministic behaviors of ordering, making it compatible with future updates  to it or related dependencies.

Let me know if further improvements or clarifications are needed!


slievrly (approved):
LGTM

PR #7819, opened 2025-11-29
test: fix brittle RPC status test

cperkkk:

This pull request addresses several flaky test, in the `apache/incubator-seata` project. The test was failing 
intermittently due to assumptions about the ordering of the test. I've noticed that when running the test in specific ordering will produce a test failure (maven surefire plugin unfortunately does not guarantee test ordering), such as:

```shell
mvn test -Dtest=org.apache.seata.common.rpc.RpcStatusTest#CbeginCount,org.apache.seata.common.rpc.RpcStatusTest#AendCount,org.apache.seata.common.rpc.RpcStatusTest#BremoveStatus
```

The tests will fail with output:
```shell
org.opentest4j.AssertionFailedError: expected: <-1> but was: <0>
	at org.apache.seata.common.rpc.RpcStatusTest.AendCount(RpcStatusTest.java:54)
```

This is due to the necessity of the count being set by beginCount first before the eventual endCount runs, but the test fails to explicitly mention ordering (such as using import org.junit.jupiter.api.Order; etc), making this a flaky and brittle test depending on the version of surefire plugin used / if there's a method rename.

It does not fixes a particular github issue.

I've only changed test and not code under test

We can permute the run order, after fixing the test, any run order by surefire will pass the test

```shell
mvn test -Dsurefire.runOrder=alphabetical -Dtest=org.apache.seata.common.rpc.RpcStatusTest#beginCount,org.apache.seata.common.rpc.RpcStatusTest#removeStatus,org.apache.seata.common.rpc.RpcStatusTest#endCount
```

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

- The fix ensures the test is robust to non-deterministic behaviors of ordering, making it compatible with future updates  to it or related dependencies.

Let me know if further improvements or clarifications are needed!


common/src/test/java/org/apache/seata/common/rpc/RpcStatusTest.java
```diff
@@ -50,6 +50,10 @@ public void beginCount() {
 
     @Test
     public void endCount() {
+        // ensure a correct state before testing
```
slievrly, line 53: How about placing it BeforeEach?

cperkkk: Hi @slievrly thank you for the reply:
the only problematic function is the endCount and not the other one, and asserts to 0 only works if there's a beginCount and a beginCount that doesn't run before removed service, so endCount evaluates to +1-1==0, if I use beforeEach (for example adding removeStatus on beforeEach), I can do that but I have to change assertions for the endCount to be -1, instead of 0, do you prefer that to be used instead of the current one? 

Thank you for your guidance.

lokidundun:
Please apply `mvn spotless:apply`

cperkkk:
> Please apply `mvn spotless:apply`

Okay, spotless:apply applied, thanks

xingfudeshi (approved):
LGTM

PR #7827, opened 2025-12-01
test: fix non-deterministic in TableMetaTest

XiaoyangCai360:

This PR resolves flakiness in the `testGetPrimaryKeyOnlyName` unit test in `rm-datasource/src/test/java/org/apache/seata/rm/datasource/sql/struct/TableMetaTest.java`.

The test was failing intermittently when multiple primary keys were defined (`id`, `userCode`). This was caused by the test relying on a fixed, hardcoded order for the list of primary key names, while the method's underlying collection (`TableMeta.getAllIndexes()`) does not guarantee a fixed iteration order.

The fix ensures a stable, order-independent comparison by sorting both the actual and expected lists of primary key names before assertion.

**No production code changes were made; this is a test-only change to improve CI stability.**

This PR specifically addresses the flakiness of existing unit tests

To confirm the flakiness is resolved, run the affected tests repeatedly using a non-deterministic test runner such as [nondex](https://github.com/TestingResearchIllinois/NonDex) to verify stability. Successful runs across multiple attempts verify the fix.

For example, run the following command:
```bash
mvn -pl rm-datasource \
    -Dtest=org.apache.seata.rm.datasource.sql.struct.TableMetaTest#testGetPrimaryKeyOnlyName \
    -DnondexRuns=10 \
    -DfailIfNoTests=false
```

- Test-only change in the `rm-datasource` module.


slievrly (approved):
LGTM

PR #7851, opened 2025-12-11
feature: add support for managing transaction groups

funky-eyes:

image

image

image

image


server/pom.xml
```diff
@@ -301,7 +301,6 @@
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
```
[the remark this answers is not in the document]
funky-eyes: That's incorrect. The common module's HttpClient has been migrated to OkHttp, so the dependency should not be declared with test scope here—remove the test scope, otherwise you'll get a ClassNotFoundException at runtime.

namingserver/src/test/java/org/apache/seata/namingserver/NamingControllerV2Test.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import java.util.Map;
+import java.util.UUID;
+
+import static org.apache.seata.common.NamingServerConstants.CONSTANT_GROUP;
+import static org.junit.jupiter.api.Assertions.*;
```
xingfudeshi: Don't use *

xingfudeshi (approved):
LGTM

namingserver/src/test/java/org/apache/seata/namingserver/NamingControllerV2Test.java
funky-eyes: done

PR #7858, opened 2025-12-14
test: Fix non-deterministic in HttpTest caused by HashMap iteration order

XiaoyangCai360:

This PR fixes test flakiness in  
`org.apache.seata.integration.http.HttpTest.convertParamOfJsonStringTest`.

The test was intermittently failing because it compared the `toString()` output of a `Map`, which depends on the iteration order of the underlying `HashMap`. Since `HashMap` does not guarantee a deterministic key order, different JVM or
NonDex execution orders could produce logically equivalent but differently ordered string representations (e.g. `{age=15, name=zhangsan}` vs `{name=zhangsan, age=15}`), causing false negatives.

The fix updates the test to perform **order-independent map comparisons** by asserting map equality directly, instead of relying on `toString()` output.

**This change is test-only and does not modify any production code.**

This PR focuses on correcting an existing flaky unit test by making its assertions deterministic and order-independent. No new functionality is introduced.

The fix can be verified by repeatedly running the affected test under a non-deterministic execution environment such as [NonDex](https://github.com/TestingResearchIllinois/NonDex).

Example command:
```bash
mvn -pl integration/http \
    -Dtest=org.apache.seata.integration.http.HttpTest#convertParamOfJsonStringTest \
    -DnondexRuns=10 \
    -DfailIfNoTests=false
```

- Test-only change in the `seata-http` module. 
- Improves CI stability under non-deterministic execution orders.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -137,6 +137,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#7808](https://github.com/apache/incubator-seata/pull/7808)] Deflake multiple Insert Executor tests by fixing order-dependent primary key (PK) value comparison
 - [[#7815](https://github.com/apache/incubator-seata/pull/7815)] test: fix combine test to avoid failure due to ordering
 - [[#7827](https://github.com/apache/incubator-seata/pull/7827)] Fix non-deteriministic in TableMetaTest#testGetPrimaryKeyOnlyName
+- [[#7858](https://github.com/apache/incubator-seata/pull/7858)] Fix flakiness in HttpTest.convertParamOfJsonStringTest caused by non-deterministic Map iteration order
 
 
 ### refactor:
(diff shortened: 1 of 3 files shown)


YongGoose:
Has there been a CI record where this test code failed?

XiaoyangCai360:
Thanks for the question.

There is no public CI failure record for this test. The flakiness was identified using NonDex, which executes tests under non-deterministic JVM behaviors (for example, randomized iteration order). Under multiple NonDex runs, this test fails intermittently because it relies on the Map.toString() output, which depends on the iteration order of HashMap.

Although this issue may not consistently surface in CI, it represents a potential source of flakiness that can be exposed under different JVM implementations or execution orders. This change removes the order-dependent assertion and makes the test deterministic.

xingfudeshi (approved):
LGTM

PR #7859, opened 2025-12-14
test: fix flaky testIsRaftMode and testToString in MetadataTest

XiaoyangCai360:

This PR resolves test flakiness in `org.apache.seata.common.metadata.MetadataTest`, specifically:
- `MetadataTest.testToString`
- `MetadataTest.testIsRaftMode`

The flakiness was caused by:
1) the test class sharing a single static `Metadata` instance across tests, making assertions dependent on test execution order (e.g., `storeMode` mutated by other tests); and
2) `testToString` relying on an exact `toString()` output, which is brittle because field/map ordering may vary across executions and under NonDex.

The fix makes the tests deterministic by setting up required state explicitly and using order-independent assertions (instead of strict string equality).

**This PR is test-only and does not modify production code.**

This PR focuses on stabilizing existing unit tests by removing order-dependent assumptions and brittle `toString()` comparisons.

Run the affected tests repeatedly under a non-deterministic execution environment such as [NonDex](https://github.com/TestingResearchIllinois/NonDex) to confirm stability.

Example command:
```bash
mvn -pl common \
    -Dtest=org.apache.seata.common.metadata.MetadataTest \
    -DnondexRuns=20 \
    -DfailIfNoTests=false
```

- Test-only change in the `common` module. 
- Improves CI stability under non-deterministic execution orders.


YongGoose (commented):
Has there been a CI record where this test code failed? 
Also, I think using beforeEach could easily fix the issue.

WDYT?

XiaoyangCai360:
Thanks for the questions and the suggestion!

1. There is no public CI failure record for these tests. The flakiness was identified using NonDex, which exposes non-deterministic behaviors such as shared static state and order-dependent assertions. Although this issue may not consistently surface in CI today, it represents a latent source of non-deterministic failures that can cause flaky CI behavior under different JVM implementations, execution orders, or future refactoring.
2. I agree that using BeforeEach would also address the shared-state issue. However, in this PR, I took a minimal approach to make the affected tests self-contained by explicitly setting the required state (storeMode) and avoiding brittle, order-dependent toString() equality checks.

YongGoose:
> Thanks for the questions and the suggestion!
> 
> 1. There is no public CI failure record for these tests. The flakiness was identified using NonDex, which exposes non-deterministic behaviors such as shared static state and order-dependent assertions. Although this issue may not consistently surface in CI today, it represents a latent source of non-deterministic failures that can cause flaky CI behavior under different JVM implementations, execution orders, or future refactoring.
> 2. I agree that using BeforeEach would also address the shared-state issue. However, in this PR, I took a minimal approach to make the affected tests self-contained by explicitly setting the required state (storeMode) and avoiding brittle, order-dependent toString() equality checks.

Ok. I understand
Nice catch :)

YongGoose:
Please resolve the conflicts

xingfudeshi (approved):
LGTM

PR #7870, opened 2025-12-17
feature: upgrade the namingserver and console modules to JDK 25 and SpringBoot 3.5, and add the spring-ai dependency to the console module.

Xb2555:

- Set the JDK version of the console and namingserver modules to 25 to support Springboot 3.5 and Spring-AI dependencies, preparing for the launch of the MCP function in the future.
- As the Springboot version has been upgraded from 2.7 to 3.5, most classes have been migrated from the javax package to the Jakarta package, with minor code changes.
- In the ConsoleRemotingFilter of the namingserver, since AsyncRestTemplate has been deprecated, virtual threads in JDK 21 and above and RestTemplate are used to implement the original asynchronous forwarding + lightweight blocking function, which is more lightweight and usable.
- The WebSecurityConfig configuration class in the Console module has also been changed to the configuration method of the new version of SpringSecurity.


namingserver/src/main/resources/application.yml
```diff
@@ -18,6 +18,9 @@
 server:
```
funky-eyes, line 18: We need to enable virtual threads.

namingserver/src/main/java/org/apache/seata/namingserver/filter/ConsoleRemotingFilter.java
```diff
@@ -57,15 +54,15 @@ public class ConsoleRemotingFilter implements Filter {
 
     private final NamingManager namingManager;
 
-    private final AsyncRestTemplate asyncRestTemplate;
+    private final RestTemplate restTemplate;
```
funky-eyes, line 56: Since we've switched to RestTemplate, we need to enable virtual threads to prevent blocking the thread pool when forwarding HTTP requests.

namingserver/src/main/java/org/apache/seata/namingserver/filter/ConsoleRemotingFilter.java
```diff
@@ -113,34 +110,27 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
-                                        response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-                                    } finally {
-                                        asyncContext.complete();
-                                    }
-                                }
-
-                                @Override
-                                public void onSuccess(ResponseEntity<byte[]> responseEntity) {
-                                    // Copy response headers and status code
+                            Thread.startVirtualThread(() -> {
```
funky-eyes: I believe there's no need to switch from Tomcat's default business thread pool (with a maximum of 200 threads) to virtual threads. Instead, we can directly configure Tomcat to use a virtual thread pool as its executor for handling business threads.

Moreover, in this scenario, we can simply wait synchronously for the response from the downstream server and then return it to the client—no asynchronous processing is required. Adopting a synchronous request model with virtual threads is perfectly sufficient and straightforward.

console/src/main/java/org/apache/seata/console/config/WebSecurityConfig.java
```diff
@@ -79,60 +84,60 @@ public class WebSecurityConfig extends WebSecurityConfigurerAdapter {
...
+    public PasswordEncoder passwordEncoder() {
+        return new BCryptPasswordEncoder();
     }
 
-    @Override
-    protected void configure(AuthenticationManagerBuilder auth) throws Exception {
-        auth.userDetailsService(userDetailsService).passwordEncoder(passwordEncoder());
+    @Bean
+    public AuthenticationManager authenticationManager() {
+        DaoAuthenticationProvider authenticationProvider = new DaoAuthenticationProvider(userDetailsService);
```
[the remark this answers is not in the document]
Xb2555: DaoAuthenticationProvider constructor is to accept UserDetailsService parameters and setUserDetailsService method has been deprecated
image
image

funky-eyes (approved):
LGTM

PR #7885, opened 2025-12-20
optimize: replace fury with fory

slievrly:

optimize fury to fory


core/src/main/java/org/apache/seata/core/serializer/SerializerServiceLoader.java
```diff
@@ -53,8 +54,14 @@ public final class SerializerServiceLoader {
 
     private static final Map<String, Serializer> SERIALIZER_MAP = new HashMap<>();
 
+    private static final Map<String, String> SERIALIZER_ALIAS_MAP = new HashMap<>();
+
     private static final String SPLIT_CHAR = ",";
 
+    static {
+        SERIALIZER_ALIAS_MAP.put("fury", "fory");
```
funky-eyes, line 62: Should we add aliases to the @LoadLevel annotation to better support compatibility when third-party components are renamed?

slievrly: These are two ways of implementation. One SPI implementation corresponding to multiple names is not a universal scenario.Because don't want to increase the complexity of the SPI. So the current implementation handles this logic in the total entry of each SPI type.

funky-eyes: ok

funky-eyes (approved):
LGTM

PR #7891, opened 2025-12-22
bugfix: raft split-brain causes incorrect cluster information

funky-eyes:

When the old leader is re-elected, it sends a heartbeat to the Naming Server—targeting either itself or the (former) leader. Although the term has already changed, the heartbeat thread and the re-election thread run concurrently. As a result, the leader status hasn't been updated yet when the heartbeat is sent. Therefore, the node needs to be corrected during the next heartbeat cycle. To achieve this, we should flag the role; if the role differs, it indicates that the node has undergone a change.

### DIFF
--- a/server/src/test/java/org/apache/seata/server/instance/RaftServerInstanceStrategyTest.java
+++ b/server/src/test/java/org/apache/seata/server/instance/RaftServerInstanceStrategyTest.java
@@ -51,8 +51,12 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 
 import static org.apache.seata.common.Constants.OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT;
-import static org.junit.jupiter.api.Assertions.*;
-import static org.mockito.Mockito.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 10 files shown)


server/src/test/java/org/apache/seata/server/instance/RaftServerInstanceStrategyTest.java
```diff
@@ -0,0 +1,184 @@
...
+import org.springframework.core.env.MutablePropertySources;
+import org.springframework.core.env.StandardEnvironment;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+import static org.apache.seata.common.Constants.OBJECT_KEY_SPRING_CONFIGURABLE_ENVIRONMENT;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
```
xingfudeshi: Do not import *

xingfudeshi (approved):
LGTM

server/src/test/java/org/apache/seata/server/instance/RaftServerInstanceStrategyTest.java
funky-eyes: done

PR #7893, opened 2025-12-23
feature: add global and branch session and lock management MCP tools

Xb2555:

1. tool for obtaining namespace information of registered TC nodes has been implemented

2. Management tools for global and branch session have been implemented

3. The function of remotely invoking the TC interface has been realized

4. A global lock management tool has been implemented


console/pom.xml
```diff
@@ -149,6 +149,11 @@
             <artifactId>spring-ai-starter-mcp-server-webmvc</artifactId>
             <version>${spring-ai.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.seata</groupId>
+            <artifactId>seata-core</artifactId>
```
funky-eyes: Why do we need to introduce this dependency?

console/src/main/java/org/apache/seata/mcp/service/impl/ConsoleRemoteServiceImpl.java
```diff
@@ -0,0 +1,283 @@
...
+
+    private final JwtTokenUtils jwtTokenUtils;
+
+    public MCPRPCServiceImpl(JwtTokenUtils jwtTokenUtils) {
+        this.jwtTokenUtils = jwtTokenUtils;
+    }
+
+    private final RestTemplate restTemplate = new RestTemplate();
+
+    private final String NAMING_SPACE_URL = "http://127.0.0.1:%s";
```
funky-eyes: Since the NamingServer and Console are integrated into a single component, why are we routing calls through network loopback? Wouldn't it be better to abstract some service interfaces on the Console side, with the actual implementations provided directly in the NamingServer?

console/src/main/java/org/apache/seata/mcp/service/impl/ConsoleRemoteServiceImpl.java
```diff
@@ -0,0 +1,243 @@
...
+
+    private final ObjectMapper objectMapper;
+
+    public ConsoleRemoteServiceImpl(JwtTokenUtils jwtTokenUtils, RestTemplate restTemplate, ObjectMapper objectMapper) {
+        this.jwtTokenUtils = jwtTokenUtils;
+        this.restTemplate = restTemplate;
+        this.objectMapper = objectMapper;
+    }
+
+    @Value("${seata.console.naming-space-url:http://127.0.0.1:%s}")
```
funky-eyes: Why is it named "naming-space-url"? And why is the address separated from the port? If the Console is deployed independently, it would make more sense to have a single property like "namingserver-addr" that contains one or more NamingServer addresses (comma-separated), allowing for either a single address or multiple ones for high availability.

console/src/main/java/org/apache/seata/mcp/service/impl/ConsoleRemoteServiceImpl.java
```diff
@@ -0,0 +1,240 @@
...
+@Service
+public class ConsoleRemoteServiceImpl implements ConsoleApiService {
+
+    private final JwtTokenUtils jwtTokenUtils;
+
+    private final RestTemplate restTemplate;
+
+    private final ObjectMapper objectMapper;
+
+    @Value("${seata.console.namingserver-addr:http://127.0.0.1:8081}")
```
funky-eyes: The address list should not include any HTTP-related information; instead, it should simply be a comma-separated list of addresses.

funky-eyes (approved):
LGTM

PR #7925, opened 2026-01-08
Fix #7869: Rename vgroup_table.cluster to cluster_name

### LINKED ISSUE #7869: A bug that causes SQL execution to fail in Oracle DB due to the cluster column in the vgroup_table table.
### Check Ahead

### Ⅰ. Issue Description

In oracle db, cluster is identified as a reserved word.

```
// script/server/db/oracle.sql
CREATE TABLE vgroup_table
(
    vGroup    VARCHAR2(255) PRIMARY KEY,
    namespace VARCHAR2(255),
    cluster   VARCHAR2(255)
);
```

So when I run the oracle.sql script in the oracle environment, it fails.

```java
// server/src/main/java/org/apache/seata/server/storage/db/store/VGroupMappingDataBaseDAO.java
public boolean insertMappingDO(MappingDO mappingDO) {
        String sql = "INSERT INTO " + vMapping + " (vgroup,namespace, cluster) VALUES (?, ?, ?)";
        Connection conn = null;
        PreparedStatement ps = null;
        try {
            int index = 1;
            conn = vGroupMappingDataSource.getConnection();
            conn.setAutoCommit(true);
            ps = conn.prepareStatement(sql);
            ps.setString(index++, mappingDO.getVGroup());
            ps.setString(index++, mappingDO.getNamespace());
            ps.setString(index++, mappingDO.getCluster());

            return ps.executeUpdate() > 0;
        } catch (SQLException e) {
            throw new SeataRuntimeException(ErrorCode.ERR_CONFIG, e);
        } finally {
            IOUtil.close(ps, conn);
        }
    }
...
```

### Ⅱ. Describe what happened

Therefore, it is expected that the insertMappingDO method of VGroupMappingDataBaseDAO will fail to execute the query when executed in an Oracle environment.

deleteMappingDOByVGroup, queryMappingDO also appear similar.

I think it would be better to change the cluster column to a keyword other than a reserved word. What do you think?

I find it odd that these queries only cause problems with Oracle, while they work fine with other database vendors.

However, I believe that using double quotes to avoid reserved word issues could potentially affect other database vendors, even if it doesn't cause problems with Oracle.

Please give me your opinion. Thank you.

You can reproduce it by setting up an Oracle DB environment and testing it.

NA

### Ⅵ. Environment

NA

Gitkbc:
What this PR does:
- Fixes Oracle SQL failures caused by reserved keyword `cluster`
- Renames `vgroup_table.cluster` to `cluster_name` across all supported databases
- Updates MappingDO, DAO queries, and result mappings
- Adds database upgrade scripts to preserve existing data

Why this approach:
- Avoids vendor-specific quoting
- Prevents Oracle case-sensitivity issues
- Keeps schema consistent across database vendors

Backward compatibility:
- Existing installations are supported via upgrade scripts

Fixes #7869


Gitkbc:
Hi maintainers,

I’ve addressed the SQL rename across all supported databases, added upgrade scripts, and included tests to improve coverage.

License and Spotless checks are now passing locally.
Please let me know if any additional adjustments or test coverage are required.

Thanks for your review.

script/server/db/mysql.sql
```diff
@@ -94,7 +94,7 @@ CREATE TABLE IF NOT EXISTS `vgroup_table`
 (
     `vGroup`    VARCHAR(255),
     `namespace` VARCHAR(255),
-    `cluster`   VARCHAR(255),
-  UNIQUE KEY `idx_vgroup_namespace_cluster` (`vGroup`,`namespace`,`cluster`)
+    `cluster_name`   VARCHAR(255),
```
funky-eyes, line 97: If we rename the column cluster to cluster_name,
what should we do with the existing vgroup data that has already been migrated to NamingServer for transaction group management?
How should the migration process to the new table structure be handled?

PR #7997, opened 2026-02-14
bugfix:Branch logic is prepare Failed at phase one, need to notify TC of global rollback

### LINKED ISSUE #7975: something about: modify XA mode pre commit transaction from commit phase to before close phase
### Check Ahead

### Ⅰ. Issue Description

this is some problem when I test in XA mode use PG，use “springboot-feign-seata-xa” project。
update Seata to 2.5.0，The Timeout Exception is ingore。

### Ⅱ. Describe what happened

1.Timeout Exception is throw

Image

2.Check the stack, Timeout Exception will Ingore By this code

Image

3.The console indicates that the transaction has been committed, but in fact, the transaction was rolled back.

Please confirm if it is a problem. If it is a problem, please fix it.

_No response_

_No response_

### Ⅵ. Environment

jdk1.8.0_172  Seata2.5.0

xiaoxiangyeyu0:
… of global rollback

When an error occurs during the close phase of XA mode, Spring will swallow the exception and TM can initiate global transaction commit normally. TC can also submit global transactions normally, which is unreasonable. Therefore, a prepare Failed status is added to report to TC and roll back globally

fixes https://github.com/apache/incubator-seata/issues/7975


common/src/main/java/org/apache/seata/core/model/GlobalStatus.java
```diff
@@ -235,4 +235,17 @@ public static boolean isTwoPhaseHeuristic(GlobalStatus status) {
...
     }
+
+    /**
+     * Is one phase prepare failed boolean.
+     *
+     * @param status the status
+     * @return the boolean
+     */
+    public static boolean isOnePhasePrepareFailed(GlobalStatus status) {
+        if (status == GlobalStatus.RollbackRetrying) {
```
funky-eyes, line 246: Why is a phase‑one failure determined by the global status rather than the branch status, and why is the check limited to whether the status is "RollbackRetrying"?

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
```diff
@@ -348,9 +348,13 @@ public void close() throws SQLException {
...
                 }
+            } catch (SQLException xe) {
+                // Branch Report to TC: Exception
+                reportStatusToTC(BranchStatus.PhaseOne_PrepareFailed);
+                throw xe;
             } catch (XAException xe) {
-                // Branch Report to TC: Failed
-                reportStatusToTC(BranchStatus.PhaseOne_Failed);
+                // Branch Report to TC: Exception
+                reportStatusToTC(BranchStatus.PhaseOne_PrepareFailed);
```
funky-eyes: When the database is MariaDB or MySQL 8.0.29 (or earlier), if an END is executed and then an exception occurs during PREPARE, with the changes introduced by this PR: after the TM makes the two-phase commit decision, the server will directly remove this branch. However, the client may continue to hold onto that connection. In this case, will the resources held by the client connection be properly released?

server/src/main/java/org/apache/seata/server/coordinator/DefaultCore.java
```diff
@@ -299,7 +299,7 @@ public boolean doGlobalCommit(GlobalSession globalSession, boolean retrying) thr
                         }
 
                         BranchStatus currentStatus = branchSession.getStatus();
-                        if (currentStatus == BranchStatus.PhaseOne_Failed) {
+                        if (currentStatus == BranchStatus.PhaseOne_Failed || currentStatus == BranchStatus.PhaseOne_PrepareFailed) {
```
funky-eyes: ditto

server/src/main/java/org/apache/seata/server/coordinator/DefaultCore.java
```diff
@@ -433,7 +433,7 @@ public boolean doGlobalRollback(GlobalSession globalSession, boolean retrying) t
                     branchSessions,
                     branchSession -> {
                         BranchStatus currentBranchStatus = branchSession.getStatus();
-                        if (currentBranchStatus == BranchStatus.PhaseOne_Failed) {
+                        if (currentBranchStatus == BranchStatus.PhaseOne_Failed || currentBranchStatus == BranchStatus.PhaseOne_PrepareFailed) {
```
funky-eyes: ditto

common/src/main/java/org/apache/seata/core/model/GlobalStatus.java
xiaoxiangyeyu0: This is the global transaction status returned by TC when TM submits a global transaction. The newly added branch status will update the global transaction status to RollbackRetraining
这是TM提交全局事务时，TC返回的全局事务状态，新加的分支状态会把全局事务状态更新为RollbackRetrying

rm-datasource/src/main/java/org/apache/seata/rm/datasource/xa/ConnectionProxyXA.java
xiaoxiangyeyu0: After the modification of the pre submission fails, call the rollback method and close the original connection
修改预提交失败后，调用回滚方法并关闭原始连接

funky-eyes (approved):
LGTM

Issue #6724, opened 2024-08-02
Prevent virtual thread pinned


funky-eyes:
## Why you need it?
适配jdk19开始的虚拟线程特性,防止synchronized导致虚拟线程pinned后线程阻塞
你可以在jdk19及以上版本增加 -Djdk.tracePinnedThreads=full 参数,使用seata-samples和seata-server.将其使用的线程池改为虚拟线程池进行测试,并且找出所有会将虚拟线程pinned的代码进行修正.当然你也可以直接阅读代码进行判断该问题是否存在,这个任务应该将server和client侧一并进行修正
注: 理论上所有单例模式,双检锁等行为存在的synchronized 对虚拟线程不会有严重影响,因为在其完成首次调用后,不会再进入synchronized所在的代码块
Adapt to the virtual thread feature introduced in JDK 19, preventing thread blocking caused by synchronized leading to virtual thread pinning
You can add the -Djdk.tracePinnedThreads=full parameter in JDK 19 and above versions. Use seata-samples and seata-server to switch their thread pools to virtual thread pools for testing purposes, and identify and correct all code segments that may cause virtual thread pinning. You may also directly review the code to assess whether this issue exists. This task should include corrections on both the server and client sides.

Note: In theory, behaviors like singleton patterns and double-checked locking involving synchronized do not significantly impact virtual threads because once their initial invocation completes, they do not re-enter the synchronized code block.
## How it could be?
A clear and concise description of what you want to happen. You can explain more about input of the feature, and output of it.

linghengqian:
- Please note that Loom will eventually address pinning of `synchronized` – see this Java 23 EA announcement https://inside.java/2024/06/22/quality-heads-up/ .  I kind of doubt that this would be a waste of time.
- Downstream developers are free to use `SDKMAN!` to switch to OpenJDK 23 EA and OpenJDK 24 EA for testing. The loom variant is located on the Project Loom Early-Access Builds page at https://jdk.java.net/loom/ .
```shell
sdk install java 23.ea.29-open
sdk use java 23.ea.29-open

sdk install java 24.ea.8-open
sdk use java 24.ea.8-open
```

funky-eyes:
> * Please note that Loom will eventually address pinning of `synchronized` – see this Java 23 EA announcement https://inside.java/2024/06/22/quality-heads-up/ .  I kind of doubt that this would be a waste of time.
[...]

PR #8023, opened 2026-03-21
refactor: supports global switching of virtual threads

funky-eyes:

除 config 模块外, 所有线程池统一收口创建路径.
增加配置项来开启虚拟线程, 或自动根据 jdk版本自动选择. jdk25 虚拟线程优先级最高. jdk21 则需要手动开启.jdk8 默认平台线程.
Unified the creation path for all thread pools (except the config module).
Added a configuration option to enable virtual threads, with automatic selection based on JDK version: virtual threads take priority on JDK 25+, must be manually enabled on JDK 21, and platform threads are used by default on JDK 8.


pom.xml
```diff
@@ -130,6 +131,16 @@
...
 
     <profiles>
+        <!-- profile: onlyBuildOnJDK21+ -->
+        <profile>
+            <id>JDK21Plus</id>
+            <activation>
+                <jdk>[21,)</jdk>
+            </activation>
+            <modules>
+                <module>threadpool-loom</module>
```
WangzJi, line 141: loom module is only wired into the reactor build and server/pom.xml here. all/pom.xml never depends on seata-threadpool-loom, so seata-spring-boot-starter on JDK 21+ wont see the virtual provider on the classpath — loadOptional returns null and we silently fall back to platform threads. Probably want a JDK21+ profile in all/pom.xml mirroring server/pom.xml:452-464.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualThreadPoolExecutor.java
```diff
@@ -0,0 +1,47 @@
...
+            boolean daemon,
+            RejectedExecutionHandler rejectedHandler) {
+        super(
+                corePoolSize,
+                maximumPoolSize,
+                keepAliveTime,
+                unit,
+                workQueue,
+                VirtualThreadFactoryHelper.newThreadFactory(threadPrefix, daemon),
+                rejectedHandler);
```
WangzJi, line 45: This doesn't match the PR description, which says virtual mode ignores maximumPoolSize/keepAliveTime and uses Integer.MAX_VALUE + SynchronousQueue. Current code passes the caller's bounds straight through, and testVirtualThreadPoolExecutorKeepsConfiguredBounds locks that in. So either the impl needs to follow the description, or the description (and the test assertion) should be updated to say virtual mode still respects caller bounds.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualScheduledThreadPoolExecutor.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+
+/**
+ * Virtual-thread-backed scheduled thread pool implementation.
+ */
+public class VirtualScheduledThreadPoolExecutor extends ScheduledThreadPoolExecutor {
+
+    public VirtualScheduledThreadPoolExecutor(
+            String threadPrefix, int corePoolSize, boolean daemon, RejectedExecutionHandler rejectedHandler) {
+        super(corePoolSize, VirtualThreadFactoryHelper.newThreadFactory(threadPrefix, daemon), rejectedHandler);
```
WangzJi, line 29: ScheduledThreadPoolExecutor parks its workers in DelayedWorkQueue.take() forever, which pins the carrier when those workers are virtual threads — so we get zero loom benefit and bring back the pinning issue from #6724. ~12 migrated call sites go through this (RegistryHeartBeats, DefaultCoordinator.retryRollbacking / asyncCommitting / xaTwoPhaseTimeoutChecker, CommonFenceConfig, ...). Maybe have the loom provider delegate newScheduledThreadPoolExecutor to the platform impl?

threadpool/src/main/java/org/apache/seata/common/thread/ThreadPoolRuntimeEnvironment.java
```diff
@@ -0,0 +1,98 @@
...
+    private ThreadPoolRuntimeEnvironment() {}
+
+    static ThreadPoolType resolveThreadPoolType() {
+        ThreadPoolType configuredType = ThreadPoolType.from(threadPoolTypeSupplier.get());
+        if (configuredType == ThreadPoolType.PLATFORM) {
+            return ThreadPoolType.PLATFORM;
+        }
+        int jdkFeature = jdkFeatureSupplier.getAsInt();
+        if (configuredType == ThreadPoolType.VIRTUAL) {
+            return jdkFeature >= 21 ? ThreadPoolType.VIRTUAL : ThreadPoolType.PLATFORM;
```
WangzJi, line 58: No logger anywhere in this class. If someone sets transport.threadpool=virtual on JDK <21 we silently downgrade here, and if loom isnt on the classpath ThreadPoolExecutorFactory.loadOptional (line 174) silently returns null. From the operators side theres no way to tell which mode is actually in effect. A one-shot WARN on each downgrade path would help a lot.

seata-spring-boot-starter/src/main/java/org/apache/seata/spring/boot/autoconfigure/SeataSagaAutoConfiguration.java
```diff
@@ -118,20 +118,13 @@ public ThreadPoolExecutor sagaAsyncThreadPoolExecutor(
...
-            return new ThreadPoolExecutor(
+            return ThreadPoolExecutorFactory.newThreadPoolExecutor(
+                    "sagaAsyncExecute",
                     properties.getCorePoolSize(),
                     properties.getMaxPoolSize(),
                     properties.getKeepAliveTime(),
                     TimeUnit.SECONDS,
                     new LinkedBlockingQueue<>(),
-                    threadFactory,
                     rejectedExecutionHandler);
```
WangzJi, line 128: Springs ThreadPoolExecutorFactoryBean produces non-daemon threads by default, but this overload of ThreadPoolExecutorFactory.newThreadPoolExecutor defaults to daemon=true. After the switch, in-flight saga branches wont keep the JVM alive on shutdown. Probably want the explicit-daemon overload with daemon=false here.

WangzJi (commented):
Review focused on the 5 most critical issues. Each inline comment is anchored at the relevant line. See individual comments for details.

seata-spring-boot-starter/src/main/java/org/apache/seata/spring/boot/autoconfigure/SeataSagaAutoConfiguration.java
funky-eyes: > Springs ThreadPoolExecutorFactoryBean produces non-daemon threads by default, but this overload of ThreadPoolExecutorFactory.newThreadPoolExecutor defaults to daemon=true. After the switch, in-flight saga branches wont keep the JVM alive on shutdown. Probably want the explicit-daemon overload with daemon=false here.

Virtual threads cannot set daemon=true, and it is only reasonable for saga threads to use daemon=true.

pom.xml
funky-eyes: No issue at all. Seata’s current releases must be compiled with JDK 25 or above, because both the Naming Server and Console require JDK 25 to compile successfully.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualThreadPoolExecutor.java
funky-eyes: PR description has been updated

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualScheduledThreadPoolExecutor.java
funky-eyes: > ScheduledThreadPoolExecutor parks its workers in DelayedWorkQueue.take() forever, which pins the carrier when those workers are virtual threads — so we get zero loom benefit and bring back the pinning issue from #6724. ~12 migrated call sites go through this (RegistryHeartBeats, DefaultCoordinator.retryRollbacking / asyncCommitting / xaTwoPhaseTimeoutChecker, CommonFenceConfig, ...). Maybe have the loom provider delegate newScheduledThreadPoolExecutor to the platform impl?

This issue doesn't occur with JDK 25.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualThreadPoolExecutor.java
WangzJi: Just double-checked — the description still has the same wording (Integer.MAX_VALUE + SynchronousQueue + ignores maximumPoolSize/keepAliveTime), and the example code block still shows the old shape. VirtualThreadPoolExecutor on the latest head also still passes the caller bounds through. Could be a stale browser tab on my side though, mind sharing the diff or pointing me to the new wording?

pom.xml
WangzJi: Compile-time JDK and runtime classpath are different things though — even if we publish starter jars built on JDK 25, the published seata-spring-boot-starter pom doesn't declare seata-threadpool-loom (because all/pom.xml doesn't), so when a user pulls the starter into their JDK 21+ app, Maven won't fetch the loom artifact and the SPI lookup returns null. End users running on whatever JDK they want still need the dependency wired in. Or am I missing something about how the release artifacts are assembled?

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualScheduledThreadPoolExecutor.java
WangzJi: Fair point on JEP 491 — agreed my pinning argument was off, ReentrantLock + Condition.await in DelayedWorkQueue.take() goes through LockSupport.park and unmounts cleanly on JDK 21+ already, so this isn't a pinning bug. Withdrawing that part. The remaining question is whether wrapping ScheduledThreadPoolExecutor with a virtual ThreadFactory actually buys us anything — concurrency is still capped at corePoolSize because that's how STPE schedules tasks, so the unbounded-virtual-threads property doesn't really apply. Not a blocker, just curious whether you measured a difference vs the platform path.

seata-spring-boot-starter/src/main/java/org/apache/seata/spring/boot/autoconfigure/SeataSagaAutoConfiguration.java
WangzJi: Right, Thread.ofVirtual() always produces daemon threads — so once saga uses the virtual provider, daemon=true is forced regardless. My concern is the change for the platform path: Spring's ThreadPoolExecutorFactoryBean defaulted to non-daemon, so saga state-machine async tasks used to keep the JVM alive on shutdown until they finished. With this PR they don't. If that's an intentional behavior change for saga, fine — just want to make sure it's a deliberate call rather than a side effect of the migration.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualScheduledThreadPoolExecutor.java
funky-eyes: > Fair point on JEP 491 — agreed my pinning argument was off, ReentrantLock + Condition.await in DelayedWorkQueue.take() goes through LockSupport.park and unmounts cleanly on JDK 21+ already, so this isn't a pinning bug. Withdrawing that part. The remaining question is whether wrapping ScheduledThreadPoolExecutor with a virtual ThreadFactory actually buys us anything — concurrency is still capped at corePoolSize because that's how STPE schedules tasks, so the unbounded-virtual-threads property doesn't really apply. Not a blocker, just curious whether you measured a difference vs the platform path.

Threads related to scheduled tasks will consume additional system resources when using platform threads.

threadpool-loom/src/main/java/org/apache/seata/common/thread/VirtualThreadPoolExecutor.java
funky-eyes: done

pom.xml
funky-eyes: > Compile-time JDK and runtime classpath are different things though — even if we publish starter jars built on JDK 25, the published seata-spring-boot-starter pom doesn't declare seata-threadpool-loom (because all/pom.xml doesn't), so when a user pulls the starter into their JDK 21+ app, Maven won't fetch the loom artifact and the SPI lookup returns null. End users running on whatever JDK they want still need the dependency wired in. Or am I missing something about how the release artifacts are assembled?

Because we need to support JDK 8 users, seata-threadpool-loom is not bundled into seata-all. Users who require virtual thread support should explicitly add the seata-threadpool-loom dependency themselves.
The seata-threadpool-loom module will be published to Maven Central when the project is deployed using JDK 25.

threadpool/src/main/java/org/apache/seata/common/thread/ThreadPoolRuntimeEnvironment.java
funky-eyes: > No logger anywhere in this class. If someone sets transport.threadpool=virtual on JDK <21 we silently downgrade here, and if loom isnt on the classpath ThreadPoolExecutorFactory.loadOptional (line 174) silently returns null. From the operators side theres no way to tell which mode is actually in effect. A one-shot WARN on each downgrade path would help a lot.

done

lokidundun (approved):
LGTM

john1337:
@funky-eyes 请问线程池使用虚拟线程池后有验证性能么

funky-eyes:
> @funky-eyes 请问线程池使用虚拟线程池后有验证性能么

Currently no, but we expect to test it before the 2.7 release.

This PR is only a refactor of the virtual thread pool. The overall codebase still pools virtual threads and enforces a maximum thread count; it just reduces the total number of threads used.

Of course, after virtual threads are enabled, we can also increase the minimum and maximum thread counts. That would help avoid unbounded virtual thread creation under heavy traffic and ensure the system still has a backpressure mechanism.

PR #8051, opened 2026-04-14
test: Isolate compatibility tests from transport protocol leakage

funky-eyes:
`ServerV1ToClientV2Test` could time out waiting for responses when earlier tests left `transport.protocol=grpc` in shared configuration state. In that condition, the V2 client in the compatibility suite could initialize with the wrong transport and never receive the expected V1 server response.

- **Problem**
  - The multi-version compatibility tests assume the Seata transport, but they were reading process-wide configuration that could be mutated by earlier gRPC-oriented tests.
  - This made `doSendRegister()` fail intermittently with `Should receive response within timeout`.

- **Change**
  - Force `transport.protocol` to `seata` in `MultiVersionCompatibilityTest` setup before creating the V2 client/server test fixtures.
  - Capture the pre-existing `transport.protocol` value and restore it in teardown so the test does not leak configuration changes into other suites.

- **Effect**
  - `ServerV1ToClientV2Test` now runs against the intended transport regardless of prior test ordering.
  - The compatibility suite becomes self-contained instead of depending on global configuration state left behind by other tests.

- **Implementation sketch**
  ```java
  @BeforeEach
  public void setUp() {
      originalTransportProtocol =
              ConfigurationFactory.getInstance().getConfig(ConfigurationKeys.TRANSPORT_PROTOCOL);
      ConfigurationTestHelper.putConfig(ConfigurationKeys.TRANSPORT_PROTOCOL, Protocol.SEATA.value);
      // initialize test fixtures...
  }

  @AfterEach
  public void tearDown() throws InterruptedException {
      // shutdown fixtures...
      if (StringUtils.isBlank(originalTransportProtocol)) {
          ConfigurationTestHelper.removeConfig(ConfigurationKeys.TRANSPORT_PROTOCOL);
      } else {
          ConfigurationTestHelper.putConfig(ConfigurationKeys.TRANSPORT_PROTOCOL, originalTransportProtocol);
      }
  }
  ```

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -64,6 +64,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#7962](https://github.com/apache/incubator-seata/pull/7962)] add unit tests for NacosRegistryProvider and NacosRegistryServiceImpl
 - [[#8003](https://github.com/apache/incubator-seata/pull/8003)] Enhance NacosRegistryServiceImplTest with additional mocks for service name group and cluster
 - [[#7915](https://github.com/apache/incubator-seata/pull/7915)] add unit tests for saga-engine low coverage components
+- [[#8051](https://github.com/apache/incubator-seata/pull/8051)] isolate compatibility tests from transport protocol leakage
 
 ### refactor:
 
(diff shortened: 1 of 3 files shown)


WangzJi (approved):
LGTM.

PR #8052, opened 2026-04-15
feature: add SAGA_ANNOTATION mode to benchmark CLI

### LINKED ISSUE #8040: [Benchmark] Add annotation-based Saga (SAGA_ANNOTATION) mode to benchmark CLI
Check Ahead
- [x] I have searched the issues of this repository and believe that this is not a duplicate.

### Why you need it?
Seata already supports annotation-based Saga through the `saga/seata-saga-annotation` module. This path is based on `@SagaTransactional`, `@CompensationBusinessAction`, and `BranchType.SAGA_ANNOTATION`.

However, `seata-benchmark-cli` currently only supports `AT`, `TCC`, and `SAGA`. That means benchmark coverage is still incomplete from a Saga perspective: we can benchmark the traditional state-machine Saga path, but we cannot benchmark the annotation-based Saga path.

If the goal of the benchmark module is to gradually cover all major Seata transaction modes, then `SAGA_ANNOTATION` is an obvious missing part. Without it, users and contributors cannot compare the annotation-based Saga execution path with other modes in a unified benchmark tool.

### How it could be?
A possible direction is to add a dedicated benchmark mode for annotation-based Saga, for example:

```bash
--mode SAGA_ANNOTATION
```

or another CLI-friendly alias if maintainers prefer one.

Possible implementation scope:

- extend `BenchmarkConfig` mode validation to accept the new mode
- extend `BenchmarkRunner` dispatch logic and add a dedicated executor for annotation-based Saga benchmarking
- provide a minimal benchmark scenario based on `@SagaTransactional` and `@CompensationBusinessAction`
- bootstrap a minimal Spring context or another suitable integration path so annotation scanning, proxying, and Saga resource registration work correctly
- define clear semantics for `--branches` and rollback percentage in this mode
- update README with runnable examples and explain the difference between `SAGA` and `SAGA_ANNOTATION`

Expected outcome:

- benchmark CLI can run annotation-based Saga end-to-end against a TC
- rollback / compensation behavior can be exercised and reflected in metrics
- benchmark users can compare traditional Saga and annotation-based Saga separately and explicitly

Other related information
Current implementation evidence:

- `test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/BenchmarkRunner.java`
  - currently only creates executors for `AT`, `TCC`, and `SAGA`
- `test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java`
  - currently validates only `AT`, `TCC`, and `SAGA`
- `core/src/main/java/org/apache/seata/core/model/BranchType.java`
  - already contains `SAGA_ANNOTATION`
- `saga/seata-saga-annotation/src/main/java/org/apache/seata/saga/rm/api/SagaTransactional.java`
- `saga/seata-saga-annotation/src/main/java/org/apache/seata/saga/rm/api/CompensationBusinessAction.java`

Design notes that may need discussion during implementation:

- whether the CLI mode name should be `SAGA_ANNOTATION`, `SAGA-ANNOTATION`, or both
- whether a small Spring container is the best benchmark bootstrap approach
- how `--branches` should map to annotation-based Saga steps in a way that is intuitive and comparable with other modes

Non-goal for this issue:

- improving the existing traditional/state-machine Saga benchmark (`BranchType.SAGA`) should be tracked separately.

zihenzzz:

fix https://github.com/apache/incubator-seata/issues/8040

  ```bash
  java -jar seata-benchmark-cli.jar --server 127.0.0.1:8091 --mode SAGA_ANNOTATION --tps 100 --duration 30 --threads 1
  ```

  ```bash
  java -jar seata-benchmark-cli.jar --server 127.0.0.1:8091 --mode SAGA_ANNOTATION --tps 100 --duration 30 --threads 1
  --branches 3 --rollback-percentage 30
  ```

### DIFF
--- a/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
+++ b/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
@@ -134,8 +134,10 @@ public class BenchmarkConfig {
     }
 
     private void validateMode() {
-        if (mode != BranchType.AT && mode != BranchType.TCC
-                && mode != BranchType.SAGA && mode != BranchType.SAGA_ANNOTATION) {
+        if (mode != BranchType.AT
+                && mode != BranchType.TCC
+                && mode != BranchType.SAGA
+                && mode != BranchType.SAGA_ANNOTATION) {
             throw new IllegalArgumentException(
                     "Unsupported mode: " + mode + ". Only AT, TCC, SAGA, and SAGA_ANNOTATION are supported.");
         }
--- a/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
+++ b/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
@@ -134,8 +134,10 @@ public class BenchmarkConfig {
     }
 
     private void validateMode() {
-        if (mode != BranchType.AT && mode != BranchType.TCC && mode != BranchType.SAGA) {
-            throw new IllegalArgumentException("Unsupported mode: " + mode + ". Only AT, TCC, and SAGA are supported.");
+        if (mode != BranchType.AT && mode != BranchType.TCC
+                && mode != BranchType.SAGA && mode != BranchType.SAGA_ANNOTATION) {
+            throw new IllegalArgumentException(
+                    "Unsupported mode: " + mode + ". Only AT, TCC, SAGA, and SAGA_ANNOTATION are supported.");
         }
     }
 
--- a/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
+++ b/test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
@@ -19,6 +19,7 @@ package org.apache.seata.benchmark.config;
 import org.apache.seata.benchmark.constant.BenchmarkConstants;
 import org.apache.seata.core.model.BranchType;
 
+import java.util.EnumSet;
 import java.util.Locale;
 
 /**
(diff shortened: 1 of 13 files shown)


test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/config/BenchmarkConfig.java
```diff
@@ -195,8 +195,12 @@ public void validate() {
     }
 
     private void validateMode() {
-        if (mode != BranchType.AT && mode != BranchType.TCC && mode != BranchType.SAGA) {
-            throw new IllegalArgumentException("Unsupported mode: " + mode + ". Only AT, TCC, and SAGA are supported.");
+        if (mode != BranchType.AT
```
WangzJi: How about using an EnumSet instead?

test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/saga/annotation/BenchmarkCompensatableService.java
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.seata.benchmark.saga.annotation;
+
+import org.apache.seata.rm.tcc.api.BusinessActionContext;
+import org.apache.seata.saga.rm.api.CompensationBusinessAction;
+import org.apache.seata.saga.rm.api.SagaTransactional;
+
+/**
+ * Benchmark service interface for SAGA_ANNOTATION mode.
+ *
+ * @author zihenzzz
```
WangzJi: The @author tag in class Javadoc is not needed.  Could you remove it?

test-suite/seata-benchmark-cli/src/main/java/org/apache/seata/benchmark/saga/annotation/BenchmarkCompensatableServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.apache.seata.benchmark.saga.annotation;
+
+import org.apache.seata.rm.tcc.api.BusinessActionContext;
+
+/**
+ * Benchmark implementation of {@link BenchmarkCompensatableService}.
+ *
+ * @author zihenzzz
```
WangzJi: Ditto.

WangzJi (approved):
LGTM.

PR #8125, opened 2026-06-03
optimize: renew table meta cache entries during refresh

wuwen5:

This PR fixes the table metadata cache refresh behavior in `AbstractTableMetaCache`.

Previously, `TABLE_META_CACHE.put(...)` was only executed when the latest fetched `TableMeta` was different from the cached value. Since the cache uses `expireAfterWrite`, unchanged table metadata would not renew the write timestamp during periodic refresh, causing active cache entries to expire after 900 seconds.

This PR removes that restriction and always writes the refreshed `TableMeta` back to `TABLE_META_CACHE` during `refresh(...)`. The change detection logic is kept only for logging, so the cache write time is continuously renewed even when the schema has not changed.

No.

This change is a small behavior adjustment in the cache refresh path and does not introduce new public behavior or API changes.

Also, verifying this issue through automated tests would require time-based cache expiration behavior and scheduled refresh coordination, which makes the test relatively heavy and less stable. The logic change itself is straightforward: always renew the cache entry during refresh, while preserving the existing schema change log behavior.

1. Configure a datasource and ensure table metadata can be loaded into `TABLE_META_CACHE`.
2. Trigger periodic `refresh(...)` calls with unchanged table schema.
3. Confirm that `TABLE_META_CACHE.put(...)` is executed on every refresh.
4. Observe that the cache entry write timestamp is continuously renewed.
5. Verify that the cache entry does not expire after 900 seconds while refresh is running normally.
6. Optionally modify the table schema and confirm that the existing info log is still printed when metadata changes.

This PR intentionally keeps the existing log behavior unchanged:
- cache entries are always rewritten during refresh
- the info log is still printed only when the fetched metadata differs from the previous cached value

The goal is only to ensure `expireAfterWrite` entries remain alive during normal periodic refresh.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -87,7 +87,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8109](https://github.com/apache/incubator-seata/pull/8109)] reduce npmjs dependencies in saga module
 - [[#8112](https://github.com/apache/incubator-seata/pull/8112)] optimize seata-server test performance
 - [[#8116](https://github.com/apache/incubator-seata/pull/8116)] fix incompatible dependencies
-- [[#8125](https://github.com/apache/incubator-seata/pull/8125)] Fix cache expiration caused by missing metadata cache renewal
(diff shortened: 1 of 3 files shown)


funky-eyes (approved):
LGTM

PR #8132, opened 2026-06-07
refactor: merge threadpool and json-common modules

slievrly:
Eliminate two top-level modules (threadpool, json-common) by distributing their code into existing modules based on dependency constraints, and create dedicated jdk17/jdk21 modules for version-specific extensions.

threadpool refactoring:
- Move SPI interface + platform impl (5 files) → common module
- Move Factory + RuntimeEnvironment (2 files) → core module
- Rename threadpool-loom → jdk21 (artifactId: seata-jdk21)
- discovery-core/consul/etcd3/redis: use PlatformThreadPoolProvider directly to avoid core→discovery-core→core circular dependency

json-common refactoring:
- Move JsonSerializer SPI + 4 impl classes + AllowlistManager → common
- Move JsonUtil (depends on ConfigurationFactory) → core module
- Rename json-common-jackson3 → jdk17 (artifactId: seata-jdk17)
- Remove json-common-core dependency from 5 downstream modules (available transitively through core)


funky-eyes (commented):
The overall module consolidation direction looks reasonable, but I think two issues should be addressed before merging:

The complete json-common-core test suite was deleted instead of being moved with the production code. This removes coverage for serializer SPI loading, Jackson 3 fallback, configuration resolution, and, more importantly, JSON deserialization allowlist security behavior. Please move serializer/factory/allowlist tests to seata-common, and move JsonUtilTest plus its configuration resources to seata-core.
The discovery implementations now instantiate PlatformThreadPoolProvider directly. This bypasses ThreadPoolExecutorFactory and therefore ignores the configured transport.threadpool mode for heartbeat, Consul notifier, Etcd watcher, and Redis subscription/update executors. Please introduce a shared provider resolver that does not create a core/discovery dependency cycle, or explicitly document and test that these executors are intentionally platform-thread-only.

Additional issues:

The deprecated integration-tx-api JsonUtil Javadoc says the replacement is in seata-common, while the class is now provided by seata-core.
The PR description refers to seata-jdk17 and seata-jdk21, but the actual artifacts are seata-spi-jdk17 and seata-spi-jdk21.
Please verify whether the removed Maven coordinates have ever been published. If so, compatibility bridge artifacts or an explicit migration strategy are needed.
Please add the corresponding change entries and run dependency:analyze for modules whose direct json-common-core dependency was removed.

PR #8139, opened 2026-06-09
bugfix: upgrade Groovy to 4.0.32 to support JDK 25 

slievrly:
Groovy 4.0.27 is the first 4.0.x release whose bundled ASM supports JDK 25 (class file major version 69). When CI runs saga script task tests on JDK 25 with Spring Boot 3.0-3.3, the Spring Boot BOM downgrades Groovy from Seata's declared 4.0.31 to as low as 4.0.15 (which only supports up to JDK 21), causing GroovyBugError: "Unsupported class file major version 69".

- Upgrade groovy.version from 4.0.31 to 4.0.32 (latest stable)
- Exclude org.apache.groovy from spring-boot-dependencies BOM import to prevent Spring Boot from downgrading Groovy below the version Seata requires

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -84,7 +84,7 @@
         <kryo-serializers.version>0.45</kryo-serializers.version>
         <hessian.version>4.0.63</hessian.version>
         <fastjson2.version>2.0.52</fastjson2.version>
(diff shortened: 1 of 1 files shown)


funky-eyes (approved):
LGTM

PR #8145, opened 2026-06-17
bugfix: fix global lock batch acquire false-failure on Dameng(DM)

### LINKED ISSUE #8144: Global lock batch acquire always fails on Dameng(DM): LockStoreDataBaseDAO.doAcquireLocks relies on executeBatch().length
# [Bug] Global lock batch acquire always fails on Dameng (DM) — `LockStoreDataBaseDAO.doAcquireLocks` relies on `executeBatch().length`

## Ⅰ. Describe the bug

When the Seata Server uses **DB store mode** with a **Dameng (DM)** database, any branch transaction that needs to acquire **two or more row locks** in a single `BranchRegister` **always fails** to acquire the global lock, even though the `lock_table` is empty and there is no real conflict.

The client side gets:

```
io.seata.rm.datasource.exec.LockConflictException: get global lock fail, xid:..., lockKeys:XXX:1;YYY:2
```

The server side logs:

```
LockStoreDataBaseDAO : Global lock batch acquire failed, xid ... branchId ... pks [1, 2]
AbstractExceptionHandler : this request cannot acquire global lock ...
```

A branch that locks only **one** row works fine. So in practice **any business transaction that modifies 2+ rows (master/detail, entity + i18n text, etc.) can never commit**, which makes AT mode effectively unusable on DM for real workloads.

## Ⅱ. Root cause

`org.apache.seata.server.storage.db.lock.LockStoreDataBaseDAO#doAcquireLocks`:

```java
protected boolean doAcquireLocks(Connection conn, List<LockDO> lockDOs) throws SQLException {
    ...
    for (LockDO lockDO : lockDOs) {
        ... ps.addBatch();
    }
    return ps.executeBatch().length == lockDOs.size();   // <-- here
}
```

It judges batch success by `executeBatch().length == lockDOs.size()`.

The DM JDBC driver (`DmdbPreparedStatement`) implements batch execution via `executeBatchByRow()` and **aggregates** the per-statement results through `ExecuteRetInfo.union(...)`, so `executeBatch()` returns an `int[]` whose **length does not equal the number of batched statements**.

As a result:
- The lock rows **are actually inserted** into `lock_table`.
- But `executeBatch().length == lockDOs.size()` evaluates to **false**.
- `doAcquireLocks` returns `false`, the caller does `conn.rollback()` (the inserted lock rows are discarded), and the branch register is reported as a lock conflict.

Single-row branches take the `doAcquireLock` (non-batch) path, which does not have this check, so they succeed — which is why the failure only appears for multi-row branches.

> Per the JDBC spec, the return value of `Statement.executeBatch()` is an array of update counts where each element may be a count, `SUCCESS_NO_INFO (-2)`, or `EXECUTE_FAILED (-3)`. Assuming `length == number of statements` is not guaranteed across drivers; relying on it is the bug.

## Ⅲ. How to reproduce

1. Seata Server `store.mode=db`, store DB = Dameng (DM8).
2. Any AT-mode business transaction whose single branch updates/inserts **≥ 2 rows** (so the branch reports `lockKey` with 2+ pks).
3. Branch register fails with `Global lock batch acquire failed` against an empty `lock_table`.

## Ⅳ. Expected behavior

The batch lock acquire should succeed when the lock rows are inserted without conflict, regardless of the driver-specific length of the `executeBatch()` return array. Real conflicts (duplicate `row_key`) still surface as `SQLIntegrityConstraintViolationException` and are already handled.

## Ⅴ. Environment

- Seata version: **2.6.0** (the same code is present on the current `2.x` branch).
- DB store: Dameng (DM8), DM JDBC driver `DmJdbcDriver18`.
- The fix is proposed in the accompanying PR.

## Ⅵ. Suggested fix

Stop relying on `executeBatch().length`. Detect failure via `Statement.EXECUTE_FAILED` instead:

```java
int[] result = ps.executeBatch();
for (int updated : result) {
    if (updated == java.sql.Statement.EXECUTE_FAILED) {
        return false;
    }
}
return true;
```

This is correct for drivers that return one element per statement and for drivers (like DM) that aggregate the result array, while preserving the existing conflict handling (the `catch (SQLIntegrityConstraintViolationException ...)` branch).

lhozy:
# PR: bugfix: fix global lock batch acquire false-failure on Dameng(DM)

## Ⅰ. Describe what this PR does

Fix `LockStoreDataBaseDAO#doAcquireLocks` so that batch global-lock acquisition no longer produces a **false "lock acquire failed"** on databases (e.g. **Dameng / DM**) whose JDBC driver returns an `executeBatch()` array whose length differs from the number of batched statements.

Fixes #8144.

## Ⅱ. Why make this change

`doAcquireLocks` judged batch success with:

```java
return ps.executeBatch().length == lockDOs.size();
```

The JDBC spec does **not** guarantee `executeBatch()` returns exactly one element per batched statement (elements may be counts, `SUCCESS_NO_INFO`, or `EXECUTE_FAILED`, and some drivers aggregate). The DM driver aggregates the per-row results (`DmdbPreparedStatement.executeBatchByRow` → `ExecuteRetInfo.union`), so the returned array length ≠ statement count.

Consequence on DM: when a branch locks ≥ 2 rows, the lock rows are inserted successfully but the `length == size` check returns `false`, so the caller rolls back the inserted locks and the branch register is reported as a lock conflict. Single-row branches (non-batch `doAcquireLock`) are unaffected, so the bug only manifests for multi-row branches — making AT mode unusable on DM for typical multi-row business transactions.

## Ⅲ. The change

`server/src/main/java/org/apache/seata/server/storage/db/lock/LockStoreDataBaseDAO.java`

```diff
             for (LockDO lockDO : lockDOs) {
                 ...
                 ps.addBatch();
             }
-            return ps.executeBatch().length == lockDOs.size();
+            // Do not rely on executeBatch().length == size: per the JDBC spec the
+            // length is not guaranteed to equal the number of statements, and some
+            // drivers (e.g. Dameng/DM) aggregate the per-statement results, returning
+            // an array of a different length. Detect failure via EXECUTE_FAILED instead;
+            // real conflicts (duplicate row_key) still throw SQLIntegrityConstraintViolationException
+            // and are handled by the catch block below.
+            int[] result = ps.executeBatch();
+            for (int updated : result) {
+                if (updated == java.sql.Statement.EXECUTE_FAILED) {
+                    return false;
+                }
+            }
+            return true;
         } catch (SQLIntegrityConstraintViolationException e) {
             LOGGER.error("Global lock batch acquire error: {}", e.getMessage(), e);
             // return false,let the caller go to conn.rollback()
             return false;
```

## Ⅳ. How tested

- **Unit test (new): `LockStoreDataBaseDAOBatchAcquireTest`** — pure Mockito, no database, runs in CI by default. It stubs `PreparedStatement.executeBatch()` to return:
  - (a) an aggregated array **shorter** than the batched-statement count (the DM behavior, e.g. `{3}` for 3 statements) → asserts acquisition **succeeds**;
  - (b) an array containing `Statement.EXECUTE_FAILED` → asserts acquisition **fails**;
  - (c) one update count per statement (`{1,1,1}`) → asserts **success**.

  Reverting this PR's change makes (a) and (b) fail, confirming the test guards the regression. Run: `mvn -pl server test -Dtest=LockStoreDataBaseDAOBatchAcquireTest`.
- Existing behavior preserved on databases that return one element per statement (MySQL, Oracle, PostgreSQL…): no `EXECUTE_FAILED` → returns `true`; genuine duplicate-`row_key` conflicts still throw `SQLIntegrityConstraintViolationException` → handled → `false`.
- On DM8 (store.mode=db): a branch acquiring multiple row locks now succeeds; a cross-resource global transaction with multi-row branches both commits and rolls back correctly, where before every multi-row branch register failed with `Global lock batch acquire failed`.

## Ⅴ. Checklist

- [x] Unit test added — `LockStoreDataBaseDAOBatchAcquireTest` asserts `doAcquireLocks` returns `true` when `executeBatch()` returns an aggregated/short array, and `false` on `EXECUTE_FAILED`.
- [x] Changelog entry — registered in `changes/en-us/2.x.md` and `changes/zh-cn/2.x.md`.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -23,6 +23,8 @@ Add changes here for all PR submitted to the 2.x branch.
 
 ### bugfix:
 
+- [#8145](https://github.com/apache/incubator-seata/pull/8145) fix global lock batch acquire false-failure on Dameng(DM)
+
 
 ### optimize:
 
@@ -44,6 +46,7 @@ Thanks to these contributors for their code commits. Please report an unintended
 <!-- Please make sure your Github ID is in the list below -->
 
 - [slievrly](https://github.com/slievrly)
+- [lhozy](https://github.com/lhozy)
 
 
 Also, we receive many valuable issues, questions and advices from our community. Thanks for you all.
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -23,6 +23,8 @@
 
 ### bugfix:
 
+- [#8145](https://github.com/apache/incubator-seata/pull/8145) 修复达梦(DM)数据库下全局锁批量获取被误判为失败的问题
+
 
 ### optimize:
 
@@ -44,6 +46,7 @@
 <!-- 请确保您的 GitHub ID 在以下列表中 -->
 
 - [slievrly](https://github.com/slievrly)
+- [lhozy](https://github.com/lhozy)
 
 
 同时，我们收到了社区反馈的很多有价值的issue和建议，非常感谢大家。
--- a/server/src/main/java/org/apache/seata/server/storage/db/lock/LockStoreDataBaseDAO.java
+++ b/server/src/main/java/org/apache/seata/server/storage/db/lock/LockStoreDataBaseDAO.java
@@ -390,7 +390,18 @@ public class LockStoreDataBaseDAO implements LockStore {
                 ps.setInt(8, lockDO.getStatus());
                 ps.addBatch();
             }
-            return ps.executeBatch().length == lockDOs.size();
+            // Do not rely on executeBatch().length == size: per the JDBC spec the length is
+            // not guaranteed to equal the number of statements, and some drivers (e.g. Dameng/DM)
+            // aggregate the per-statement results, returning an array of a different length.
+            // Detect failure via EXECUTE_FAILED instead; real conflicts (duplicate row_key) still
+            // throw SQLIntegrityConstraintViolationException and are handled by the catch block below.
+            int[] result = ps.executeBatch();
+            for (int updated : result) {
+                if (updated == java.sql.Statement.EXECUTE_FAILED) {
+                    return false;
+                }
+            }
+            return true;
         } catch (SQLIntegrityConstraintViolationException e) {
             LOGGER.error("Global lock batch acquire error: {}", e.getMessage(), e);
             // return false,let the caller go to conn.rollback()
--- a/server/src/test/java/org/apache/seata/server/storage/db/lock/LockStoreDataBaseDAOBatchAcquireTest.java
+++ b/server/src/test/java/org/apache/seata/server/storage/db/lock/LockStoreDataBaseDAOBatchAcquireTest.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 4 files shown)


funky-eyes (approved):
LGTM

funky-eyes:
If you're using the DingTalk app, please send your DingTalk account to my email at [jianbin@apache.org](mailto:jianbin@apache.org), and I will invite you to join the Seata developer group

PR #8165, opened 2026-07-14
feature: add GraalVM native image build support for seata-namingserver

### LINKED ISSUE #8164: [Feature] Provide GraalVM Native Image packaging for Seata namingserver
### Check Ahead

### Why you need it?

Seata namingserver currently runs only on the JVM, requiring a JDK/JRE runtime environment. This introduces:

- **Larger footprint**: JVM-based deployment needs a full JDK/JRE, increasing image size and memory consumption.
- **Slower startup**: JVM class loading and warmup add seconds to startup time, which is problematic in containerized/cloud-native environments where fast scaling is essential.
- **Operational complexity**: Managing JDK versions, JVM tuning parameters, and GC behavior adds overhead.

Supporting GraalVM native image compilation would enable the namingserver to be compiled into a standalone native binary, delivering **sub-second startup**, **reduced memory footprint**, and **simplified deployment** — aligning with the cloud-native and Kubernetes-native deployment trends.

### Build Pipeline

1. **Install phase** (`install-namingserver-jar`):
    - Build and locally install the namingserver JAR (with `spotless-apply` as a prerequisite).

2. **Metadata generation phase** (`run-namingserver-native-jar`):
    - Run the namingserver JAR with `-agentlib:native-image-agent` to trace reflection, resources, serialization, JNI, and proxy usage.
    - Output reachability metadata to `target/native-image-config/`.

3. **Metadata merge phase** (`run-merge-native-namingserver`):
    - Run the Java-based merge via the `test-native-metadata-merge` module (`ExecuteMergeNativeImageMetadataTests#namingServer`) to merge the agent-generated `reachability-metadata.json` into the committed native-image metadata file. Uses key-based array merging for idempotent, deterministic results.

4. **Native compilation phase** (`package-namingserver-native`):
    - Run `spring-boot:process-aot` for Spring Boot AOT processing.
    - Run `native:compile` via GraalVM `native-maven-plugin` to produce the native binary with a platform-specific classifier (e.g., `seata-namingserver-2.8.0-SNAPSHOT-linux-amd64`).

5. **Verification phase**:
    - Start the native binary and verify it passes a health check on port 8081.
    - Start the Seata server in seata registry mode (connecting to the native namingserver).
    - Run the `test-native-namingserver` test suite (18 test classes) against the running native binary.

### Build Infrastructure (`pom.xml`, `build/pom.xml`, `namingserver/pom.xml`, `server/pom.xml`)

- Add **GraalVM `native-maven-plugin`** (`org.graalvm.buildtools`, v1.1.5) to parent POM (`build/pom.xml`) plugin management, with `native.platform` property for OS/arch-aware binary naming.
- Add OS/arch platform profiles in root `pom.xml` (`native-linux-amd64`, `native-linux-aarch64`, `native-darwin-aarch64`, `native-windows-amd64`) that auto-activate based on the build host and set `native.platform` for native binary naming. **Note**: macOS Intel (`native-darwin-x86_64`) is not included — GraalVM has discontinued updates for that platform.
- Add a **`native` profile** to `namingserver/pom.xml` with `spring-boot:process-aot` and `native:compile` goals.
- Add a **`release-seata-jar` profile** in `server/pom.xml` for producing the executable fat JAR via `spring-boot:repackage` (needed when the Seata server connects to the native namingserver in tests).
- Add Maven profiles `test-native-namingserver` and `test-native-metadata-merge` in root `pom.xml` to conditionally include the new test modules.

### Native Image Metadata

- Add a single **`reachability-metadata.json`** (GraalVM standard format, schema v1.2.0) consolidating reflection, resource bundles, serialization, JNI, and proxy metadata — auto-generated via `-agentlib:native-image-agent` and maintained via the merge module.

### Metadata Merge Tooling (`test-suite/test-native-metadata-merge/`)

- Java-based merge module replacing the earlier Python script approach. Uses key-based array merging (by `name`/`type`/`pattern` keys) for idempotent, deterministic metadata updates.
- Includes the GraalVM `reachability-metadata.json` JSON schema (v1.2.0) for validation.
- Unit tests (`MergeNativeImageMetadataTests`) verify merge logic; execution test (`ExecuteMergeNativeImageMetadataTests`) drives the actual merge from Makefile/CI.

### Integration Tests (`test-suite/test-native-namingserver/`)

- 18 test classes exercising the namingserver HTTP API against a running native binary: group management (`AddGroupTests`, `ChangeGroupTests`), instance registration/discovery (`RegisterInstanceTests`, `UnregisterInstanceTests`, `DiscoveryTests`), batch operations (`BatchRegisterTests`), cluster management (`ClusterDataTests`, `ClustersTests`), global sessions/locks (`GlobalSessionTests`, `GlobalLockTests`), namespaces (`NamespaceTests`, `NamespaceV1Tests`), health check (`HealthTests`), login (`LoginTests`), overview (`OverviewTests`), watch list (`WatchListTests`), version (`VersionTests`), and application-level smoke tests.
- These tests require a running native binary and Seata server to execute. In CI, these are started automatically; for local development, the Makefile provides step-by-step targets with workload prompts.

### Configuration

- **`logback-spring.xml`**: Simplified to remove `<if>` conditional logic and `LOG_BASH_DIR` property (not supported in GraalVM native images at build time). Now always uses classpath `resource` includes for `console-appender.xml` and `file-appender.xml`. This is the **default** configuration for both `java -jar` and native image use.
- **`logback-spring-jvm.xml`** (new): Retains the original `<if>` conditional logic with external `conf/` directory support. Used **only** in the distribution package — `distribution/release-seata.xml` copies it as `conf/logback-spring.xml` so the launch script picks up the JVM-only variant. This preserves backward compatibility for JVM distribution users.
- **`distribution/release-seata.xml`**: Updated to source `logback-spring-jvm.xml` (renamed to `logback-spring.xml` at packaging time).
- Root `pom.xml`: Configured `spotless-maven-plugin` to format JSON files under `META-INF/native-image/`.

### CI Workflow (`.github/workflows/native-namingserver.yml`)

- Multi-platform native image build matrix: `ubuntu-24.04` (x86_64), `ubuntu-24.04-arm` (ARM64), `macos-26` (Apple Silicon only), `windows-latest` (x86_64).
- Triggers: push to `test*`/version branches/`2.x`, pull request to `2.x`/`develop`/`master` (with path filtering for relevant sources), and `workflow_call` for reuse.
- Build steps: GraalVM JDK 25 setup, Maven cache (restore/save), dependency install, native image compilation, SHA256 checksum, binary analysis (`file` + `ldd` on Linux), git SHA artifact, artifact upload.
- Post-build verification: starts the native binary, health-checks it, starts the Seata server (JVM) in seata registry mode, waits for both, then runs the full `test-native-namingserver` test suite.
- Concurrency: group by workflow + PR number/ref, cancel in-progress on PR updates.

### Code Changes

- **`SingleResult.java`**: Added a no-arg constructor for deserialization support (required by `RestTemplate` in the `test-native-namingserver` module). This is the only production code change outside of build/config/test files.
- **`SingleResultTest.java`**: Added a unit test for the new no-arg constructor.

### Changelog

- Updated `changes/en-us/2.x.md` and `changes/zh-cn/2.x.md` with the GraalVM native image feature entry ([#8165](https://github.com/apache/incubator-seata/pull/8165)).

### Makefile

- **Server JAR targets**: `install-server-jar`, `run-server-jar`, `install-run-server-jar`, `install-run-server-jar-registry-seata`, `run-server-jar-registry-seata`.
- **Namingserver metadata targets**: `install-namingserver-jar`, `install-run-namingserver-native-jar`, `run-namingserver-native-jar` (with workload step prompts).
- **Native image targets**: `install-namingserver-native`, `package-namingserver-native`, `run-namingserver-native` (with workload step prompts).
- **Test & merge targets**: `test-native-namingserver`, `run-merge-native-namingserver`.
- **Code quality**: `spotless-check` and `spotless-apply` with native test modules included.
- General improvements: consolidated `.PHONY` declarations, `SERVER_VERSION` / `NATIVE_PLATFORM` / `NACOS_SERVER_ADDR` variables, expanded help text, sorted/improved target dependencies.

### Other related information

- **Target version**: Seata 2.x
- **GraalVM**: Requires GraalVM JDK 25+ with `native-image` installed.
- **CI-tested platforms**: Linux x86_64 (`ubuntu-24.04`), Linux ARM64 (`ubuntu-24.04-arm`), macOS Apple Silicon (`macos-26`), Windows x86_64 (`windows-latest`).
- **Local-dev tested platforms**: macOS (aarch64), Linux (amd64).
- The metadata merge is now handled by a Java-based module (`test-native-metadata-merge`) with key-based array merging — more maintainable and CI-friendly than the earlier Python script approach. Additional config types can be supported by extending the merge logic in `MergeNativeImageMetadata.java`.
- The `logback-spring-jvm.xml` file preserves backward compatibility for JVM distribution users who need external `conf/` directory support via `<if>` conditionals. Native image users get the simplified `logback-spring.xml` without conditionals.

xuxiaowei-com-cn:

This PR adds **GraalVM native image compilation support for the Seata namingserver**, enabling the namingserver to be compiled into a standalone native binary with sub-second startup and reduced memory footprint.

#### Key changes:

- **Build infrastructure** (`pom.xml`, `build/pom.xml`, `namingserver/pom.xml`, `server/pom.xml`):
    - Added `native-maven-plugin` (`org.graalvm.buildtools`, v1.1.5) to parent POM (`build/pom.xml`) plugin management, with `native.platform` property for OS/arch-aware binary naming.
    - Added OS/arch platform profiles in root `pom.xml` (`native-linux-amd64`, `native-linux-aarch64`, `native-darwin-aarch64`, `native-windows-amd64`) that auto-activate based on the build host and set `native.platform` for native binary naming.
    - Added `native` profile to `namingserver/pom.xml` with `spring-boot:process-aot` and `native:compile` goals.
    - Added `release-seata-jar` profile in `server/pom.xml` for producing the executable fat JAR via `spring-boot:repackage` (needed when the Seata server connects to the native namingserver in tests).
    - Added Maven profiles `test-native-namingserver` and `test-native-metadata-merge` in root `pom.xml` to conditionally include the new test modules.

- **Native image metadata** (`namingserver/src/main/resources/META-INF/native-image/`):
    - Added `reachability-metadata.json` — consolidated native-image metadata (auto-generated via `-agentlib:native-image-agent`), covering reflection, resource bundles, serialization, JNI, and proxy configuration for Seata classes, Spring Boot auto-configuration, Jackson serialization, and Actuator endpoints. The metadata is in GraalVM's standard `reachability-metadata.json` format (schema v1.2.0).

- **Metadata merge tooling** (`test-suite/test-native-metadata-merge/`):
    - Added a new Java-based test module (`test-native-metadata-merge`) for merging agent-generated `reachability-metadata.json` into the committed native-image metadata files. Uses key-based array merge for idempotent, deterministic results.
    - Includes unit tests (`MergeNativeImageMetadataTests`) and an execution test (`ExecuteMergeNativeImageMetadataTests`) that can be driven via the Makefile.
    - Includes the GraalVM `reachability-metadata.json` JSON schema (v1.2.0) for validation.
    - Replaces the earlier Python-based `merge_native_image_config.py` script with a Maven-integrated Java solution.

- **Integration tests** (`test-suite/test-native-namingserver/`):
    - Added a new test module with **18 test classes** that exercise the namingserver HTTP API against a running native binary: `AddGroupTests`, `BatchRegisterTests`, `ChangeGroupTests`, `ClusterDataTests`, `ClustersTests`, `DiscoveryTests`, `GlobalLockTests`, `GlobalSessionTests`, `HealthTests`, `LoginTests`, `NamespaceTests`, `NamespaceV1Tests`, `OverviewTests`, `RegisterInstanceTests`, `SeataTestNativeNamingserverApplicationTests`, `UnregisterInstanceTests`, `VersionTests`, `WatchListTests`.
    - These tests verify the native binary correctly handles all namingserver REST API endpoints.

- **Configuration**:
    - `logback-spring.xml`: Removed `<if>` conditional logic and `LOG_BASH_DIR` property (not supported in GraalVM native images at build time). Now always uses classpath `resource` includes for `console-appender.xml` and `file-appender.xml`. This is the **default** configuration for both `java -jar` and native image use.
    - `logback-spring-jvm.xml` (new): Retains the original `<if>` conditional logic with external `conf/` directory support. This file is used **only** in the distribution package — `distribution/release-seata.xml` copies it as `conf/logback-spring.xml` so the launch script picks up the JVM-only variant.
    - `distribution/release-seata.xml`: Updated to source `logback-spring-jvm.xml` (renamed to `logback-spring.xml` at packaging time) for the distribution package, preserving external configuration support for JVM users.
    - Root `pom.xml`: Added Spotless JSON formatting for `META-INF/native-image/*.json` files.

- **CI workflow** (`.github/workflows/native-namingserver.yml`):
    - Multi-platform native image build matrix: `ubuntu-24.04` (x86_64), `ubuntu-24.04-arm` (ARM64), `macos-26` (Apple Silicon only), `windows-latest` (x86_64).
    - Triggers: push to `test*`/version branches/`2.x`, pull request to `2.x`/`develop`/`master` (with path filtering for relevant sources), and `workflow_call` for reuse.
    - Build steps: GraalVM JDK 25 setup, Maven cache (restore/save), dependency install, native image compilation, **SHA256 checksum calculation**, **binary analysis** (`file` and `ldd` on Linux), **git SHA artifact** (`git-sha.txt`), artifact upload, native binary smoke test (health check on port 8081).
    - Post-build verification: Starts the Seata server (JVM) connected to the native namingserver, waits for both to be healthy, then runs the `test-native-namingserver` test suite against the running native binary.
    - Concurrency: group by workflow + PR number/ref, cancel in-progress on PR updates.

- **Code changes**:
    - `SingleResult.java`: Added a no-arg constructor for deserialization support (required by `RestTemplate` in the `test-native-namingserver` module).

- **Changelog**:
    - Updated `changes/en-us/2.x.md` and `changes/zh-cn/2.x.md` with the GraalVM native image feature entry ([#8165](https://github.com/apache/incubator-seata/pull/8165)).

- **Makefile**:
    - Added `install-server-jar` / `run-server-jar` / `install-run-server-jar` — build and run the Seata server JAR.
    - Added `install-namingserver-jar` / `run-namingserver-native-jar` — build and run the namingserver JAR with the native-image agent for metadata collection.
    - Added `install-run-server-jar-registry-seata` / `run-server-jar-registry-seata` — run the Seata server in seata registry mode (connecting to the namingserver).
    - Added `test-native-namingserver` — run the native namingserver integration test suite.
    - Added `run-merge-native-namingserver` — merge collected native-image metadata via the Java merge module.
    - Added `install-namingserver-native` / `package-namingserver-native` — compile the native binary.
    - Added `run-namingserver-native` — run the compiled native binary (with workload step prompts).
    - Added `spotless-check` / `spotless-apply` targets with the native test modules included.
    - General improvements: consolidated `.PHONY` declarations, added `SERVER_VERSION` and `NATIVE_PLATFORM` variables, expanded help text, sorted/improved target dependencies.

https://github.com/apache/incubator-seata/issues/8164

This PR adds build infrastructure, CI configuration, and test modules — not functional changes to the namingserver business logic. The native binary is verified through:

1. **CI workflow**: The native binary starts and responds to a health check on port 8081; the full `test-native-namingserver` test suite runs against it with a live Seata server backend.
2. **Integration test module** (`test-suite/test-native-namingserver`): 18 test classes cover the namingserver REST API (group management, instance registration, discovery, health, login, etc.) against a running native binary.
3. **Metadata merge module** (`test-suite/test-native-metadata-merge`): Unit tests verify the idempotent merge logic for `reachability-metadata.json`.

The changes to `logback-spring.xml`, `logback-spring-jvm.xml`, and native-image metadata files are configuration-level and do not alter business logic. The single code change (`SingleResult.java` no-arg constructor) is purely for deserialization support. Existing unit/integration tests continue to pass on the JVM path.

1. **Prerequisites**: Install GraalVM JDK 25+ with `native-image`.

2. **Generate metadata** (optional — baseline metadata is already committed):
   ```bash
   make install-run-namingserver-native-jar   # Build & run with native-image agent
   # Exercise the namingserver API, then stop (Ctrl+C) to flush metadata
   make run-merge-native-namingserver         # Merge collected metadata
   ```

3. **Build the native binary**:
   ```bash
   make install-namingserver-native           # Build the native image
   ```

4. **Run the native binary**:
   ```bash
   make run-namingserver-native
   ```
   Verify the namingserver starts successfully (expect sub-second startup) and responds to HTTP requests at `http://localhost:8081/naming/v1/health`.

5. **Run integration tests** (requires a running native namingserver + Seata server):
   ```bash
   # In separate terminals:
   make run-namingserver-native               # Start native namingserver
   make run-server-jar-registry-seata         # Start Seata server in seata registry mode
   make test-native-namingserver              # Run the test suite
   ```

- The `reachability-metadata.json` is auto-generated by the GraalVM native-image agent. It should be reviewed for completeness but expect future iterations as more runtime paths are exercised.
- The `logback-spring.xml` change simplifies configuration by always using classpath resources. The original `<if>`-based behavior with external `conf/` directory support is preserved in `logback-spring-jvm.xml`, which the distribution package uses. Reviewers should confirm this split is acceptable for both the JVM distribution path and the native image path.
- The metadata merge module (`test-native-metadata-merge`) uses key-based array merging in Java, ensuring idempotent and deterministic results. The merge is driven via `ExecuteMergeNativeImageMetadataTests` with an environment variable flag.
- The native namingserver test module (`test-native-namingserver`) requires a running native binary and Seata server to execute. In CI, these are started automatically; for local development, the Makefile provides step-by-step targets with workload prompts.
- `SingleResult.java`: A single no-arg constructor was added for `RestTemplate` deserialization — this is the only production code change outside of build/config/test files.


xuxiaowei-com-cn:
@funky-eyes 

I have added a new test module, `test-suite/test-native-namingserver`, dedicated to end‑to‑end (E2E) testing of the `namingserver` built as a GraalVM native image. I would appreciate your feedback and suggestions if you have a moment.

slievrly (commented):
Big and genuinely useful feature — GraalVM native support for the namingserver is a natural target given its startup/memory profile. The build infrastructure is thorough (multi-platform CI, Makefile targets, an integration test module), and committing baseline reflection/resource metadata is the right call for reproducibility.

That said, this PR bundles several changes that don't obviously belong to "native image support for namingserver". I'd like to raise a few concerns for discussion — none are hard blockers, but a couple deserve either a split or a clearer justification before merge.

### 1. Scope creep

Two changes look unrelated to the stated goal:

- **`server/pom.xml`** adds a new `release-seata-jar` profile that repackages the TC as an executable Spring Boot fat JAR. The CI workflow uses it to `spring-boot:run` the TC as a smoke dependency for the namingserver native tests, but the profile itself is a cross-cutting change to the TC packaging story and should stand on its own PR — otherwise anyone bisecting future TC packaging behavior will end up in a native-image PR.
- **`common/src/main/java/org/apache/seata/common/result/SingleResult.java`** adds a no-arg constructor. Presumably this is for Jackson deserialization inside the new integration tests (`restTemplate.getForObject(..., SingleResult.class)`). That's a fair reason, but it's a change to a shared `common` type driven purely by a test-side need — worth a short comment explaining why it exists so no one deletes it later as "unused". Even better, deserialize into a dedicated test-side DTO to avoid coupling `common` to the test suite.

### 2. `logback-spring.xml` regression

The removal of the `<if>` conditional loses the ability to override log config location via `spring.config.additional-location`. Today users can drop a `logback/console-appender.xml` next to `application.yml` and it will be picked up; after this change only the classpath resource is used.

The PR body says this is because `<if>` isn't supported in GraalVM native-image at build time — that's accurate, but the fix drops the JVM path along with the native one. Suggested alternatives:

- Split into `logback-spring.xml` (JVM, with `<if>`) and `logback-spring-native.xml` (native, without), and swap them based on a Spring profile activated by the AOT step.
- Or keep the classpath include as default but re-add the `<if>` branch guarded so it's evaluated only when Janino is on the classpath (which it won't be in the native image).

Losing customizable log paths is user-visible; at minimum it needs a note in the changelog.

### 3. Python as a build dependency

`script/native/merge_native_image_config.py` (351 lines) introduces Python 3 as a build tool for anyone regenerating metadata. The generated baseline is committed, so it's not required in CI, but it does introduce a maintenance surface (JSON schema drift, Python 3 version pinning, etc.). Two lower-friction options:

- A Maven plugin execution (e.g. `exec-maven-plugin` with `groovy-maven-plugin`, or a small Ant task) — stays in-tree.
- A small Java `main` class in a build-helper module — everyone already has javac.

Not blocking, but "Python required to regenerate native metadata" should at least be documented in the Makefile help or a README next to the script.

### 4. Baseline metadata — reviewability and stability

- `reflect-config.json` is 5781 lines; `resource-config.json` is 1391 lines. These can't be meaningfully code-reviewed. That's fine — as long as we're clear that the review is on the tooling around them, not the entries themselves.
- Auto-generated metadata drifts with JDK / GraalVM / Spring Boot version. Once this is committed, every dep bump risks a diff of thousands of lines. Consider a CI job that detects drift (agent-run vs committed diff) and either fails or opens a PR — otherwise the metadata will silently go stale.
- Any runtime path that isn't in the recorded set will fail at runtime in the native image (`ClassNotFoundException`, `NoSuchMethodException`). Since GraalVM's `add-reachability-metadata` goal is already configured, the vendor-provided metadata for Spring Boot / Spring Framework should cover most of the framework side — the committed `reflect-config.json` should ideally only contain seata-specific entries. Worth a pass to see whether entries like `ch.qos.logback.classic.*`, `apple.security.AppleProvider`, etc. are actually needed on top of the vendor-provided metadata.

### 5. Test coverage & CI shape

- The `test-native-namingserver` tests are integration-style — they assume a native binary is listening on `127.0.0.1:8081`. The CI workflow starts it, but running these locally requires the whole native build first. A comment at the top of each test class stating that requirement would save future contributors some head-scratching.
- The native CI matrix builds only `namingserver`. The `server` (TC) is only started via `spring-boot:run` in JVM mode inside the same workflow. If TC native support is on the roadmap, worth documenting that intent; if not, the `release-seata-jar` profile in `server/pom.xml` (see point 1) is a bit surprising.
- Windows native image builds are historically slow and fragile — worth watching this job's timing after a few CI runs and considering `continue-on-error: true` if it becomes a false-positive source.
- `macos-26` runners are quite new; if availability/pricing is a concern for the Apache org's action minutes, `macos-latest` or `macos-14` would be cheaper.

### 6. Minor

- `native-maven-plugin` version `1.1.3` — worth double-checking this against the upstream release stream (the GraalVM Native Build Tools 0.10.x line was the current stable when I last checked). If `1.x` is intentional, ignore.
- `<buildArgs></buildArgs>` is empty in `namingserver/pom.xml` — either drop it, or add typical hardening args like `-H:+ReportExceptionStackTraces --initialize-at-build-time=`.
- The Makefile `package-namingserver-native` target invokes `spotless-apply` before building. That's convenient but couples "make me a native binary" to formatting the whole tree — surprising for a user just trying to build. Might be worth pulling `spotless-apply` out as an explicit prerequisite documented in the help text.

### Summary

The direction is good and the effort is impressive. My main asks before merge:

1. Split out the `server/pom.xml` `release-seata-jar` addition to its own PR, or explain why it's a hard dependency of this one.
2. Justify (or move to a test-side DTO) the `SingleResult` no-arg constructor.
3. Preserve the `spring.config.additional-location`-driven logback override for the JVM path.

Everything else (Python vs Maven, metadata drift detection, matrix scope, comments in tests) can be follow-up. Nice work overall.

xuxiaowei-com-cn:
@slievrly 

1. Scope creep:
   1. `release-seata-jar` is used to package into a .jar file, then the following command is used to collect GraalVM native data. Reference: https://github.com/xuxiaowei-com-cn/incubator-seata/blob/xuxiaowei/schedule-native/README.md
       ```shell
       ${GRAALVM_HOME}/bin/java -agentlib:native-image-agent=config-output-dir=./target/native-image-config -jar ./namingserver/target/seata-namingserver.jar
       ```
   2. Comments have been added to the SingleResult.java no-arg constructor.

2. logback-spring.xml:
    1. When using logback-spring-native.xml, it is difficult to specify after the GraalVM native build is completed.
    2. A changelog entry for the logback-spring.xml changes has been added.

3. Python as a build dependency:
    1. Has been replaced with a pure Java implementation (no external dependencies).

4. Baseline metadata — reviewability and stability:
    1. https://github.com/xuxiaowei-com-cn/incubator-seata/blob/xuxiaowei/schedule-native/README.md describes the daily scheduled data collection approach.
    2. Some dependencies do not natively support GraalVM native by default, so they need to be added in this project. Entries like `ch.qos.logback.classic.*` and `apple.security.AppleProvider` are non-seata and are correct.

5. Test coverage & CI shape:
    1. test-native-namingserver is an e2e test, not activated by default. It is used only when CI or standalone testing of GraalVM native build artifacts is performed.
    2. Comments have been added to the test-native-namingserver test class.
    3. macos-26: Public project CI should be free of charge. macos-latest currently points to macos-26. This is a pinned version here; see: https://github.com/actions/runner-images
    4. In CI, running the server is only required for executing test-native-namingserver e2e tests against the GraalVM native namingserver API after the namingserver build is completed.

6. Minor:
    1. For released versions of native-maven-plugin, see: https://repo1.maven.org/maven2/org/graalvm/buildtools/native-maven-plugin/maven-metadata.xml
    2. native-maven-plugin 1.1.5 is currently the highest released version.
    3. The empty buildArgs block has been removed.

funky-eyes (commented):
1. [Blocking] Native CI is skipped for most changes that can break the native image

Suggested location: .github/workflows/native.yml, under pull_request.paths

The current pull_request.paths filter only includes the workflow file and the committed native-image metadata directory.

As a result, changes to any of the following can be merged without rebuilding or testing the native image:

namingserver/src/main/java/**
namingserver/pom.xml
namingserver/src/main/resources/**, including logback-spring.xml
the root/build POMs
common/** and console/**
test-suite/test-native-namingserver/**
dependency version changes

These are all direct inputs to the native image.

The push trigger also does not include 2.x, so the merged commit is not revalidated on the target branch.

Please either remove the path filter or include all native build/runtime/test inputs, and run the workflow on pushes to 2.x as well.

2. [Blocking] The custom metadata merger silently drops valid GraalVM metadata

Suggested location: script/native/MergeNativeImageConfig.java, around KEY_TO_FILE, mergeListFormat, and mergeIncludesFormat

The merger currently supports only the reflection and resources top-level sections, and unsupported sections are only logged and skipped.

There are several correctness problems here:

JNI, serialization, and predefined-class metadata can be silently discarded.
Reflection entries whose type is not a plain string, such as proxy metadata, are skipped.
For an existing reflection type, only methods and fields are merged. Flags such as allDeclaredConstructors, allDeclaredMethods, allPublicMethods, unsafeAllocated, and condition are not merged.
Resource bundle entries do not contain glob, so they are skipped.
A reachability-metadata glob is copied directly into the legacy resource-config.json pattern field, although glob and Java regular-expression syntax are not equivalent.

This can produce a successful build while silently omitting runtime metadata, which directly conflicts with the goal of avoiding runtime missing-class/resource failures.

Please prefer keeping a standard reachability-metadata.json directly, or use GraalVM's supported agent/plugin metadata merge facilities. If this custom merger is retained, it should fully preserve the GraalVM schema and fail on unsupported metadata rather than silently skipping it.

3. [Blocking] Unicode escapes are corrupted by the custom JSON parser

Suggested location: script/native/MergeNativeImageConfig.java, parseString()

The \uXXXX branch currently replaces every Unicode escape with ?:

case 'u':
    sb.append('?');
    index += 4;
    break;

This is lossy JSON parsing. It can corrupt resource names, regular expressions, class names, conditions, or other metadata values containing escaped Unicode. Surrogate pairs are not handled either.

Because this tool rewrites committed native-image metadata, a lossy parser is unsafe. Please use a standard JSON implementation already available in the project, or implement complete Unicode escape and surrogate-pair decoding.

4. [Blocking] The documented native build target is not self-contained on a clean checkout

Suggested location: Makefile, target package-namingserver-native

The target currently builds only:

-pl namingserver

without -am.

However, seata-namingserver depends on reactor artifacts such as seata-console and seata-common. On a clean checkout with an empty Maven local repository, this command depends on those project SNAPSHOT artifacts already being installed.

The CI succeeds because an earlier step explicitly runs an install with -am, but the documented Make target does not perform that bootstrap.

Please make the target reproducible from a clean repository, for example by adding -am or making it depend on an explicit reactor installation target. It would also be useful to validate it in CI with an empty Maven repository.

5. [Blocking] Metadata regeneration currently records startup only

Suggested location: Makefile, target run-namingserver-native-metadata

The target only launches the JAR with the Native Image tracing agent. It does not send any requests or execute a defined application workload.

The tracing agent records only the dynamic paths exercised during that run. Therefore, following the documented metadata-generation steps can produce startup-only metadata and still miss registration, discovery, authentication, long polling, serialization, MCP, metrics, and error-handling paths.

Please make metadata regeneration deterministic:

Start the JAR with the tracing agent.
Wait for readiness.
Run the complete integration/contract suite against it.
Shut it down cleanly so the agent flushes its output.
Merge or copy the output using GraalVM-supported tooling.

The workload used to generate committed metadata should also be documented and executable in CI.

6. [Blocking] The test suite does not yet prove java -jar and native-image behavioral parity

Suggested location: top-level review comment or test-suite/test-native-namingserver

The current tests prove startup and several console/management APIs, but they do not cover the complete naming-server contract.

Important endpoints include:

register and batch register
unregister
discovery
add/change group
long-polling watch and watch-list behavior
static Console resources
actuator/Prometheus resources
external configuration and custom logging configuration
negative/error response paths

A native binary may start successfully while a less frequently used endpoint later fails because of missing reflection, resources, serialization, or proxy metadata.

Please introduce a black-box contract suite and execute the same suite against both:

the executable java -jar artifact, and
the native executable.

At minimum, a scenario such as register -> cluster query -> discovery/watch -> unregister should be covered, and the JAR/native responses should be compared.

7. [Blocking] This changes existing JVM logging behavior instead of adding native-only support

Suggested location: namingserver/src/main/resources/logback-spring.xml

Previously, when spring.config.additional-location was configured, the naming server loaded:

${LOG_BASH_DIR}/logback/console-appender.xml
${LOG_BASH_DIR}/logback/file-appender.xml

The PR removes that branch and always loads the classpath resources.

The distribution launcher explicitly provides an external conf directory and an external logging.config. After this change, users can still replace logback-spring.xml, but modifications to the external conf/logback/console-appender.xml and conf/logback/file-appender.xml files are ignored.

This is a behavior regression for the regular JVM distribution and conflicts with the goal of making native mode behave like java -jar.

Please preserve the existing JVM logging behavior and introduce a native-specific logging configuration or native hint instead. A parity test using the real external distribution conf directory should be added.

8. [Major] Define and test the minimum supported Linux runtime

Suggested location: .github/workflows/native.yml or the PR documentation

I inspected the uploaded Linux x64 artifact. It is dynamically linked and requires:

GLIBC_2.34
libz.so.1
/lib64/ld-linux-x86-64.so.2

Therefore, this binary will not run on Linux distributions whose glibc is older than 2.34, even though the regular java -jar distribution may run there.

Please explicitly define the supported Linux baseline. If broader compatibility is required, build on the oldest supported distribution/container, or provide a musl/static variant. CI should run the produced binary in the chosen minimum-runtime container rather than only on the build host.

9. [Major] The macOS x86_64 profile is not actually validated

Suggested location: .github/workflows/native.yml

The workflow contains only one macos-26 job, and the actual GitHub-hosted runner used by the latest successful build is ARM64.

Therefore, the darwin-x86_64 Maven profile is currently untested.

Please either add an Intel macOS build/smoke-test job or explicitly scope macOS support to Apple Silicon for the first release.

10. [Suggestion] Reduce and make the generated metadata conditional

Suggested location: the committed native-image metadata files

The current build reports more than 11,000 types registered for reflection, and the reflection configuration contains environment-specific entries such as Apple security provider classes.

This looks like a broad, unfiltered agent snapshot. It increases image size, build time, review noise, and the chance that metadata becomes stale, while still not proving that all application paths were exercised.

Please consider:

agent access/caller filters
conditional metadata
native:list-libraries-missing-metadata
separating application-owned metadata from dependency metadata
documenting the exact workload and GraalVM version used to regenerate it

Generated metadata should ideally be reproducible and produce a minimal or explainable diff.

xuxiaowei-com-cn:
@funky-eyes 

1. Done
2. Done, using test-suite/test-native-metadata-merge/src/test/java/org/apache/seata/metadata/ExecuteMergeNativeImageMetadataTests.java
3. Done, same as above
4. Done, see: install-namingserver-native, package-namingserver-native
5. Done, see commit c165c028
   1. Tests have been added in test-suite/test-native-namingserver/src/test/java/org/apache/seata/namingserver
   2. The existing reachability-metadata.json data comes from https://github.com/xuxiaowei-com-cn/incubator-seata/actions/runs/30513683171
      namingserver-macos-26-997d0d9b5c6fc11a0b46ff3c4329222ce538d083-reachability-metadata (fully reproducible),
      and is already the most comprehensive (covering Linux ARM, Linux AMD, Windows).
   3. According to the subsequent Artifacts from https://github.com/xuxiaowei-com-cn/incubator-seata/actions/workflows/schedule-native-namingserver-metadata.yml,
      no diff has been generated so far, indicating that the existing reachability-metadata.json is already the most complete data
6. Done, same as above
7. Done
8. Although Linux uses dynamically linked libraries, it can run as long as ldd version is greater than 2.32. Common Docker image ldd versions can be found at: https://xuxiaowei.io/t/ldd/52
   Personally tested on a debian 12.0 (released June 2023) Docker image, and it runs normally
   
    image
    
    img

9. Comments have been added in .github/workflows/native-namingserver.yml
10. See 5

funky-eyes (approved):
LGTM

PR #8167, opened 2026-07-17
bugfix: fix xid carrying the host name when SEATA_IP is set to a host name

### LINKED ISSUE #8158: XIDLoadBalance fails to match when SEATA_IP is a domain name
### Check Ahead

### Ⅰ. Issue Description

SEATA_IP=seata.aaa.com 时，XID 生成的是域名前缀 seata.aaa.com:8092:12345。
XIDLoadBalance.select() 用 InetSocketAddress.equals() 比较 XID 的域名和通道池里解析后的 IP，
域名 vs IP 永远不匹配。

每次分支事务注册都走 fallback 随机负载均衡，高并发下有性能开销。

Root Cause：
  XID.java:55-62 直接拼接 SEATA_IP，不做解析
  InetSocketAddress 带域名创建时是 unresolved，equals() 不解析只比字符串

建议：
  在 XIDLoadBalance.select() 中先判断 isUnresolved()，如果 true 解析成 IP 再比
  或者文档明确要求 SEATA_IP 必须填 IP

### Ⅱ. Describe what happened

SEATA_IP=seata.aaa.com 时，XID 生成的是域名前缀 seata.aaa.com:8092:12345。
XIDLoadBalance.select() 用 InetSocketAddress.equals() 比较 XID 的域名和通道池里解析后的 IP，
域名 vs IP 永远不匹配。

每次分支事务注册都走 fallback 随机负载均衡，高并发下有性能开销。

Root Cause：
  XID.java:55-62 直接拼接 SEATA_IP，不做解析
  InetSocketAddress 带域名创建时是 unresolved，equals() 不解析只比字符串

建议：
  在 XIDLoadBalance.select() 中先判断 isUnresolved()，如果 true 解析成 IP 再比
  或者文档明确要求 SEATA_IP 必须填 IP

_No response_

_No response_

_No response_

### Ⅵ. Environment

_No response_

xucq07:

When `SEATA_IP` (or `--host`) is set to a host name, the xid carries the host name instead of an ip address, e.g. `seata.aaa.com:8091:123456`.

`Server#start` validates the host with `NetUtil.isValidIp`, which internally calls `convertIpIfNecessary` to resolve a host name to its ip address before validating it. But the validation result is thrown away and the **original** string is stored:

```java
if (NetUtil.isValidIp(parameterParser.getHost(), false)) {
    XID.setIpAddress(parameterParser.getHost());   // the raw host name
}
```

The host name then propagates into every xid, and `XIDLoadBalance#select` has to build an `InetSocketAddress` from it on every branch register to compare against the resolved addresses in the channel list. That has three consequences:

1. A DNS lookup on the branch register hot path. The JVM positive cache defaults to 30s, so this is a recurring cost rather than a one-off.
2. If the lookup fails at select time, the `InetSocketAddress` is unresolved. `InetSocketAddress#equals` then compares host name strings against resolved invokers, never matches, and every branch register silently degrades to the random fallback — sticky routing to the tc owning the global transaction is lost.
3. If the name has multiple A records, `getByName` returns the first one, which may point at a different tc than the one that created the xid.

This PR resolves the host name once, at startup, so the xid always carries an ip literal:

- Store `NetUtil.convertIpIfNecessary(host)` instead of the raw host. `isValidIp` on the preceding line already performed the same resolution, so this hits the JVM cache and adds no startup cost.
- Log the `host name -> ip` mapping at INFO, which makes case 3 above diagnosable.
- Make `NetUtil#convertIpIfNecessary` public (it was private and only used by `isValidIp`); behaviour unchanged.

`XIDLoadBalance` is deliberately left alone. Resolving there would repeat work already done by the `InetSocketAddress` constructor and would keep the DNS lookup on the hot path; fixing the value at the source removes the problem for all three cases at once.

The other three `XID.setIpAddress` call sites already pass ip literals (`NetUtil.getIgnoredInterfacesLocalIp` / `NetUtil.getLocalIp`), so `Server.java` is the only place where a host name can reach an xid.

fixes #8158

Note that the root cause differs from the analysis in the issue: `new InetSocketAddress(host, port)` does resolve the name, and `InetSocketAddress#equals` compares the resolved `InetAddress` while ignoring the host name, so a host name and an ip do match in the common case. The failure only appears when resolution fails at select time or when the name resolves to a different address, as described above.

Added `NetUtilTest#testConvertIpIfNecessary`, covering the ipv4/ipv6 literal pass-through, host name resolution (using `localhost`, so it does not depend on external DNS) and the unresolvable host case.

`mvn -pl common -am test -Dtest=NetUtilTest`

End to end: start a server with `SEATA_IP` set to a resolvable host name and check that the generated xid carries the ip address rather than the name, and that the startup log records the resolution.

`convertIpIfNecessary` throws an unchecked `RuntimeException` wrapping `UnknownHostException` when the host name cannot be resolved, so an unresolvable `SEATA_IP` fails server startup instead of falling back to the local ip branch. That behaviour is pre-existing — `isValidIp` on the preceding line already throws for the same input — and this PR does not change it. It seems worth addressing separately.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -27,6 +27,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] fix TCC fence cleanup deleting in-progress/unexpired sibling branch records
 - [[#8145](https://github.com/apache/incubator-seata/pull/8145)] fix global lock batch acquire false-failure on Dameng(DM)
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] fix Saga auto-configuration being skipped on Spring Boot 4.x because of a hard `DataSourceAutoConfiguration` class reference
+- [[#8159](https://github.com/apache/incubator-seata/pull/8159)] fix xid carrying the host name instead of the ip address when `SEATA_IP` is set to a host name
 
 
 
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -27,7 +27,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] fix TCC fence cleanup deleting in-progress/unexpired sibling branch records
 - [[#8145](https://github.com/apache/incubator-seata/pull/8145)] fix global lock batch acquire false-failure on Dameng(DM)
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] fix Saga auto-configuration being skipped on Spring Boot 4.x because of a hard `DataSourceAutoConfiguration` class reference
-- [[#8159](https://github.com/apache/incubator-seata/pull/8159)] fix xid carrying the host name instead of the ip address when `SEATA_IP` is set to a host name
+- [[#8167](https://github.com/apache/incubator-seata/pull/8167)] fix xid carrying the host name instead of the ip address when `SEATA_IP` is set to a host name
 
 
 
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -27,7 +27,7 @@
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] 修复 TCC fence 清理时误删同一全局事务中仍在进行(TRIED)或未过期的分支记录的问题
 - [#8145](https://github.com/apache/incubator-seata/pull/8145) 修复达梦(DM)数据库下全局锁批量获取被误判为失败的问题
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] 修复 Spring Boot 4.x 下由于硬编码 `DataSourceAutoConfiguration` 类引用导致 Saga 自动配置被跳过的问题
-- [[#8159](https://github.com/apache/incubator-seata/pull/8159)] 修复 `SEATA_IP` 配置为域名时 xid 中携带域名而非 IP 的问题
+- [[#8167](https://github.com/apache/incubator-seata/pull/8167)] 修复 `SEATA_IP` 配置为域名时 xid 中携带域名而非 IP 的问题
 
 
 ### optimize:
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -27,6 +27,7 @@
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] 修复 TCC fence 清理时误删同一全局事务中仍在进行(TRIED)或未过期的分支记录的问题
 - [#8145](https://github.com/apache/incubator-seata/pull/8145) 修复达梦(DM)数据库下全局锁批量获取被误判为失败的问题
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] 修复 Spring Boot 4.x 下由于硬编码 `DataSourceAutoConfiguration` 类引用导致 Saga 自动配置被跳过的问题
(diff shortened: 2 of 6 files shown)


pyshiweijia:
I checked the current head (`a329bcaf`) locally with:

`mvn -pl common -am test -Dtest=NetUtilTest -Dsurefire.failIfNoSpecifiedTests=false -Dlicense.skip=true -B`

All 31 `NetUtilTest` cases pass. I found two follow-ups before merge:

1. Both `changes/en-us/2.x.md` and `changes/zh-cn/2.x.md` link to `#8159`, but this PR is `#8167`.
2. The new test covers `convertIpIfNecessary`, whose resolution behavior already existed; it would still pass if `Server` accidentally kept passing `rawHost` to `XID.setIpAddress`. Could you add focused coverage that configures a hostname and asserts that the value assigned to `XID` is the resolved IP literal? Extracting the host-selection/assignment logic into a package-private method may make that path easy to test.

That second test would directly guard the regression reported in #8158.

slievrly (commented):
Thanks for the thorough root-cause write-up — the three failure modes in `XIDLoadBalance#select` and why the current `isValidIp` + raw-host store leaves an unresolved `InetSocketAddress` are laid out clearly.

That said, I'd like to push back on the direction rather than the code. Instead of silently accepting a host name and translating it to an IP at startup, I think `SEATA_IP` (and `--host`) should reject any non-IP input outright. Reasoning:

### 1. Troubleshooting becomes indirect

With host-name → IP resolution folded into startup:

- The operator sees `SEATA_IP=seata.aaa.com` in config.
- The XID carries `1.2.3.4:8091:...`.
- The client-side load balancer matches on `1.2.3.4`.
- Correlating "which TC owns this xid" now requires reading the startup log to recover the host-name → IP mapping at the exact moment the process started.

If the DNS record changes after startup (an operator moves the TC, or the record TTL expires and a rotating answer resolves differently), the XID still carries the frozen-at-startup IP — but the operator, looking at the config, sees a host name and may reasonably assume the current DNS answer is what's in use. This is exactly the kind of hidden indirection that turns a five-minute incident into a two-hour one.

### 2. It affects client-side load balancing

`XIDLoadBalance` picks a TC by matching the XID address against resolved invoker addresses. That only works if the address in the XID is stable and unambiguous. Host-name input leaves two failure modes even after this fix:

- **Split resolution.** If the same host name resolves to different IPs on the TC and on the client (split-horizon DNS, DNS-cache skew, multiple A records returned in different orders), sticky routing silently degrades to the random fallback. The client and the TC agree the address is "correct" but disagree on which machine that is.
- **Frozen mapping surprises operators.** If someone later assumes "I can repoint the TC by updating DNS" — a reasonable inference from "I configured a host name" — they'll find that XIDs already in flight are pinned to the old IP, and new ones use whatever the TC resolved *at its own startup*, not what the DNS says today.

Fail-fast on non-IP input makes "IP is the contract" explicit at deploy time, rather than something operators discover during an incident.

### 3. Suggested direction

- In `Server#start` (or a dedicated validator), reject `parameterParser.getHost()` if it isn't an IPv4/IPv6 literal — including host names that happen to resolve — with a clear message such as `SEATA_IP must be an IP address literal; got "seata.aaa.com". Resolve the host name at deploy time and pass the IP.`
- Keep the `NetUtil.convertIpIfNecessary` visibility change and the `NetUtilTest` addition — they're useful on their own.
- The `XIDLoadBalance` hot-path DNS lookup and the three failure modes in the description become structurally impossible rather than smoothed over.

If backward compatibility is a hard constraint (existing operators may already pass host names), a staged path is:

1. Log at **WARN**, not INFO, when a host name is used, and call out the frozen-at-startup behaviour explicitly.
2. Deprecate host-name input, document the recommended pattern (`convertIpIfNecessary` at deploy time / entrypoint script), and remove support in a later release.

The code change itself is clean and the failure analysis is accurate — this is purely about which contract we want `SEATA_IP` to enforce. Happy to discuss.

PR #8170, opened 2026-07-17
bugfix: use UTF-8 for ConfigTools encryption

### LINKED ISSUE #8168: ConfigTools encryption corrupts non-ASCII content on non-UTF-8 platforms
### Check Ahead

### Ⅰ. Issue Description

`ConfigTools.publicEncrypt` and `ConfigTools.privateEncrypt` convert plaintext with `String.getBytes()`, which uses the platform-default charset. Both decrypt methods decode the recovered bytes as UTF-8. The mismatch corrupts non-ASCII content when the default charset is not UTF-8.

### Ⅱ. Describe what happened

On Windows with a GBK default charset, an RSA round trip changes the plaintext `中文` to `????`.

```java
KeyPair keyPair = ConfigTools.getKeyPair();
String publicKey = ConfigTools.getPublicKey(keyPair);
String privateKey = ConfigTools.getPrivateKey(keyPair);
String content = "\u4e2d\u6587";

String encrypted = ConfigTools.publicEncrypt(content, publicKey);
assertThat(ConfigTools.privateDecrypt(encrypted, privateKey)).isEqualTo(content);
```

```text
expected: "中文"
 but was: "????"
```

The private-encrypt/public-decrypt path has the same encoding mismatch.

Both encryption paths should encode plaintext as UTF-8, matching the existing UTF-8 decoding in the decryption paths. RSA round trips should preserve Unicode content on every supported platform.

1. Run on a platform whose default charset is not UTF-8, such as Windows with GBK.
2. Add the test above to `ConfigToolsTest`.
3. Run `mvn -pl common -am test -Dtest=ConfigToolsTest -Dlicense.skip=true -B`.

A minimal fix is to use `content.getBytes(StandardCharsets.UTF_8)` in both encryption methods. ASCII behavior remains unchanged. A regression test and implementation are ready on branch `pyshiweijia:bugfix/config-tools-utf8`.

### Ⅵ. Environment

- Branch: `2.x`
- Revision: `e6d0860a4345b10cb59c65c78215ec51d67f59d1`
- JDK: Oracle JDK 17.0.14; targeted regression also passes after the fix on JDK 8u202
- Maven: 3.8.4
- OS: Windows 11 amd64
- Default charset: GBK

pyshiweijia:

### I. Describe what this PR did

Use UTF-8 explicitly when `ConfigTools` converts plaintext to bytes for public-key and private-key encryption. Extend the round-trip test with Unicode content.

### II. Does this pull request fix one issue?

fixes #8168

### III. Why don't you add test cases (unit test/integration test)?

A regression test is included. Before the fix, it fails on a Windows/GBK environment because the Unicode text `中文` is decrypted as `????`.

### IV. Describe how to verify it

`mvn -pl common -am test -Dlicense.skip=true -B`

Result: 689 tests run, 0 failures, 0 errors, 4 skipped.

Compatibility checks:

- JDK 17: complete `common` suite passed.
- JDK 8: the Unicode regression test passed.
- JDK 8 complete-suite baseline currently has two unrelated failures in `HttpClientUtilTest`: it expects `ConnectException`, while this Windows environment returns `SocketTimeoutException` for `localhost:9999`.

### V. Special notes for reviews

Decryption already decodes plaintext bytes with UTF-8. This change makes both encryption paths use the same deterministic charset and removes platform-dependent behavior. ASCII input is unchanged.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -27,6 +27,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] fix TCC fence cleanup deleting in-progress/unexpired sibling branch records
 - [[#8145](https://github.com/apache/incubator-seata/pull/8145)] fix global lock batch acquire false-failure on Dameng(DM)
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] fix Saga auto-configuration being skipped on Spring Boot 4.x because of a hard `DataSourceAutoConfiguration` class reference
+- [[#8170](https://github.com/apache/incubator-seata/pull/8170)] fix ConfigTools encryption corrupting non-ASCII content on non-UTF-8 platforms
 
 
 
@@ -58,6 +59,7 @@ Thanks to these contributors for their code commits. Please report an unintended
 - [lhozy](https://github.com/lhozy)
 - [Zhengcy05](https://github.com/Zhengcy05)
 - [neu-hsc](https://github.com/neu-hsc)
+- [pyshiweijia](https://github.com/pyshiweijia)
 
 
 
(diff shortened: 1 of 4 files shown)


funky-eyes (approved):
LGTM

PR #8175, opened 2026-07-22
test: fix laky testCachedConfigurationChangeListener due to async race condition

### LINKED ISSUE #8174: Bug Report — testCachedConfigurationChangeListener Flaky Test
### Check Ahead

### Ⅰ. Issue Description

`ConfigurationChangeListenerTest.testCachedConfigurationChangeListener()` is a **flaky test** that fails intermittently in CI and local environments:

```
[ERROR] Failures: 
[ERROR]   ConfigurationChangeListenerTest.testCachedConfigurationChangeListener:234 expected: <true> but was: <false>
```

https://github.com/apache/incubator-seata/actions/runs/29933243660

Image

### Ⅱ. Describe what happened

The test asserts `changeEventCalled.get()` returns `true`, but it returns `false` intermittently.

**Root cause:** The default implementation of `ConfigurationChangeListener.onProcessEvent()` submits the `onChangeEvent()` callback to a thread pool **asynchronously** via `getExecutorService().submit(...)`, and the method returns immediately:

```java
// config/seata-config-core/src/main/java/org/apache/seata/config/ConfigurationChangeListener.java:63-69
default void onProcessEvent(ConfigurationChangeEvent event) {
    getExecutorService().submit(() -> {   // ⚠️ async submission
        beforeEvent(event);
        onChangeEvent(event);
        afterEvent(event);
    });
}
```

The test code before the fix was synchronous:

```java
listener.onProcessEvent(event);                    // async submission, returns immediately
Assertions.assertTrue(changeEventCalled.get());    // ❌ race: callback may not have executed yet
```

There is a **race condition** between the test thread and the thread-pool worker. Most of the time the pool executes the callback quickly enough for the test to pass, but under CI load or scheduling delays, the assertion runs before the callback, causing the failure.

The test should pass deterministically, unaffected by thread scheduling timing.

1. Run the test on a CI environment or a machine under load:

    ```bash
    mvn test -pl config/seata-config-core -Dtest=ConfigurationChangeListenerTest#testCachedConfigurationChangeListener
    ```

2. Run it repeatedly (e.g., in a loop with `-DfailIfNoTests=false`) to surface the intermittent failure.
3. Error message: `expected: <true> but was: <false>`

**Fix approach:** Use `CountDownLatch` to synchronously wait for the async callback (commit f1cf46c3):

1. Call `latch.countDown()` inside the `onChangeEvent()` callback to signal the test thread.
2. Call `latch.await(5, TimeUnit.SECONDS)` to wait up to 5 seconds.
3. The timeout acts as a safety net to prevent the test from hanging indefinitely in edge cases.

### Ⅵ. Environment

- JDK version: any
- Seata client/server version: 2.x
- OS: any (higher reproduction rate on CI / constrained environments)

xuxiaowei-com-cn:
onProcessEvent() submits onChangeEvent async to a thread pool by default. The test asserted changeEventCalled immediately, racing the executor thread. Use CountDownLatch with 5s timeout to wait for the async callback before asserting.

Fix flaky `ConfigurationChangeListenerTest.testCachedConfigurationChangeListener()`.

**Problem:** The default implementation of `ConfigurationChangeListener.onProcessEvent()` submits the `onChangeEvent()` callback to a thread pool via `ExecutorService.submit()`, returning immediately. The original test asserted `changeEventCalled == true` **right after** calling `onProcessEvent()`, creating a race condition with the pool worker thread. Under CI load this fails intermittently:

```
[ERROR]   ConfigurationChangeListenerTest.testCachedConfigurationChangeListener:234 expected: <true> but was: <false>
```

**Fix:** Introduce a `CountDownLatch` synchronization mechanism:

- Call `latch.countDown()` inside the anonymous `CachedConfigurationChangeListener.onChangeEvent()` callback.
- Call `latch.await(5, TimeUnit.SECONDS)` to wait for async callback completion with a 5-second timeout.
- Add `throws InterruptedException` to the test method signature.

fixes https://github.com/apache/incubator-seata/issues/8174

This change is the test fix itself — no additional test cases are needed. The updated test uses `CountDownLatch` to ensure the async callback completes before asserting, eliminating the race condition.

```bash
# Single run
mvn test -pl config/seata-config-core -Dtest=ConfigurationChangeListenerTest#testCachedConfigurationChangeListener

# Stress test (loop to ensure stability)
for i in $(seq 1 50); do
  mvn test -pl config/seata-config-core -Dtest=ConfigurationChangeListenerTest#testCachedConfigurationChangeListener -q
  echo "Run $i: $?"
done
```

Before the fix, failures occur intermittently. After the fix, the test should pass 100% of the time.

- Test-only change — no production code is touched.
- `CountDownLatch` is the lightest-weight "wait for async completion" mechanism in the Java concurrency standard library with zero external dependencies.
- The 5-second timeout is sufficient for normal execution while preventing the test from hanging indefinitely in extreme scenarios.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -41,6 +41,7 @@ Add changes here for all PR submitted to the 2.x branch.
 ### test:
 
 - [[#8146](https://github.com/apache/incubator-seata/pull/8146)] test: cover rm-datasource resource id initializers
+- [[#8175](https://github.com/apache/incubator-seata/pull/8175)] fix flaky testCachedConfigurationChangeListener due to async race condition
 
 ### refactor:
 
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -40,6 +40,7 @@
 ### test:
 
 - [[#8146](https://github.com/apache/incubator-seata/pull/8146)] test: 覆盖 rm-datasource resource id initializers
+- [[#8175](https://github.com/apache/incubator-seata/pull/8175)] 修复 testCachedConfigurationChangeListener 因异步竞态条件导致的偶发失败
 
 ### refactor:
 
--- a/config/seata-config-core/src/test/java/org/apache/seata/config/ConfigurationChangeListenerTest.java
+++ b/config/seata-config-core/src/test/java/org/apache/seata/config/ConfigurationChangeListenerTest.java
@@ -19,8 +19,10 @@ package org.apache.seata.config;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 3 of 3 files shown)


funky-eyes (approved):
LGTM

PR #8179, opened 2026-07-24
bugfix: set java.version to 17 for console module to fix compilation error

### LINKED ISSUE #8178: console module uses java.net.http.HttpClient (requires Java 11+) but inherits Java 1.8 compilation settings from parent POM, causing compilation errors
### Check Ahead

### Ⅰ. Issue Description

The `console` module uses `java.net.http.HttpClient` (introduced in Java 11), but the module is compiled with Java 8 compatibility settings inherited from the parent POM. This causes compilation errors when running in IDEA.

### Ⅱ. Describe what happened

When running the `console` module in IDEA, a compilation error related to `java.net.http.HttpClient` occurs:

Image

**video**

https://github.com/user-attachments/assets/27275343-a77d-4a66-9d33-f59e6b8d5173

**Root Cause Analysis:**

1. `java.net.http.HttpClient` was added in Java 11.
   
2. The `console` module does not explicitly configure `java.version`, `maven.compiler.source`, or `maven.compiler.target`.
3. The `console` module's parent is `org.apache.seata:seata-parent`, which also does not define these properties.
4. `org.apache.seata:seata-parent`'s parent is `org.apache.seata:seata-build`, which configures:
   ```xml
   <java.version>1.8</java.version>
   <maven.compiler.source>${java.version}</maven.compiler.source>
   <maven.compiler.target>${java.version}</maven.compiler.target>
   ```
5. As a result, the `console` module inherits Java 1.8 compile target, but the source code uses `java.net.http.HttpClient` which requires Java 11+.

The `console` module should compile successfully with a Java version that supports `java.net.http.HttpClient` (Java 11+), without compilation errors in IDEA.

1. Open the Seata project in IntelliJ IDEA.
2. Navigate to the `console` module.
3. Observe the compilation error on `java.net.http.HttpClient` usage — it will be flagged as unavailable because the module is configured with Java 1.8 source/target level.
4. The error occurs because `java.net.http.HttpClient` is available starting from Java 11, but the inherited Maven compiler settings target Java 1.8.

The fix is to add the following properties to `console/pom.xml`:

```xml
<java.version>17</java.version>
<maven.compiler.source>${java.version}</maven.compiler.source>
<maven.compiler.target>${java.version}</maven.compiler.target>
```

This overrides the inherited Java 1.8 setting from `seata-build` and aligns with the Spring Boot 4.0.6 / Spring Framework 7.0.7 baseline already used in the console module.

### Ⅵ. Environment

IntelliJ IDEA 2026.2.0.1

xuxiaowei-com-cn:

Fix the `console` module compilation error caused by `java.net.http.HttpClient` requiring Java 11+ while the module inherited Java 1.8 compiler settings from the parent POM.

**Changes:**

In `console/pom.xml`, add Java version properties under `<properties>`:

```xml
<java.version>17</java.version>
<maven.compiler.source>${java.version}</maven.compiler.source>
<maven.compiler.target>${java.version}</maven.compiler.target>
```

This overrides the inherited `<java.version>1.8</java.version>` from `org.apache.seata:seata-build`, ensuring the `console` module compiles with Java 17, which is consistent with the Spring Boot 4.0.6 and Spring Framework 7.0.7 baseline already used by the console module.

fixes https://github.com/apache/incubator-seata/issues/8178

This is a build configuration change only (Maven compiler properties). The existing test suite covers the runtime behavior. No new test cases are needed for compile-level configuration adjustments.

1. Import the project in IntelliJ IDEA.
2. Open the `console` module — confirm that `java.net.http.HttpClient` usages no longer show compilation errors.
3. Run `mvn compile -pl console` — the build should succeed without errors.

- `java.net.http.HttpClient` was introduced in Java 11, so the `<java.version>` must be ≥ 11.
- Java 17 is chosen to align with Spring Boot 4.0.6 / Spring Framework 7.0.7 (which require Java 17+).
- The `console` module already specifies `spring-boot-for-server.version=4.0.6` and `spring-framework-for-server.version=7.0.7`, confirming Java 17 is the correct baseline.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -27,6 +27,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] fix TCC fence cleanup deleting in-progress/unexpired sibling branch records
 - [[#8145](https://github.com/apache/incubator-seata/pull/8145)] fix global lock batch acquire false-failure on Dameng(DM)
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] fix Saga auto-configuration being skipped on Spring Boot 4.x because of a hard `DataSourceAutoConfiguration` class reference
+- [[#8179](https://github.com/apache/incubator-seata/pull/8179)] set java.version to 17 for console module to fix compilation error
 
 
 
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -27,6 +27,7 @@
 - [[#8138](https://github.com/apache/incubator-seata/pull/8138)] 修复 TCC fence 清理时误删同一全局事务中仍在进行(TRIED)或未过期的分支记录的问题
 - [#8145](https://github.com/apache/incubator-seata/pull/8145) 修复达梦(DM)数据库下全局锁批量获取被误判为失败的问题
 - [[#8157](https://github.com/apache/incubator-seata/pull/8157)] 修复 Spring Boot 4.x 下由于硬编码 `DataSourceAutoConfiguration` 类引用导致 Saga 自动配置被跳过的问题
+- [[#8179](https://github.com/apache/incubator-seata/pull/8179)] 将 console 模块的 java.version 设置为 17 以修复编译错误
 
 
 ### optimize:
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -32,6 +32,9 @@
     <description>console for Seata built with Maven</description>
 
     <properties>
+        <java.version>17</java.version>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
+        <maven.compiler.target>${java.version}</maven.compiler.target>
         <spring-boot-for-server.version>4.0.6</spring-boot-for-server.version>
         <spring-framework-for-server.version>7.0.7</spring-framework-for-server.version>
         <snakeyaml-for-server.version>2.0</snakeyaml-for-server.version>
--- a/console/pom.xml
+++ b/console/pom.xml
@@ -32,6 +32,7 @@
     <description>console for Seata built with Maven</description>
 
     <properties>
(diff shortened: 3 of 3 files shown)


xuxiaowei-com-cn (commented):
LGTM

slievrly (approved):
Small, targeted fix — the console module genuinely uses `java.net.http.HttpClient` (Java 11+) in `RestClientConfig`, and it inherits `<java.version>1.8</java.version>` from `build/pom.xml`, so the compile error is real. Java 17 aligns with the Spring Boot 4.0.6 / Spring Framework 7.0.7 baseline the module already pulls in. LGTM.

A couple of non-blocking observations for later:

### 1. Consider `<maven.compiler.release>` instead of `source`/`target`
`--release N` binds against the correct JDK bootclasspath and prevents accidental use of newer APIs (e.g. a Java 21 API slipping through on a build machine running JDK 21). The project uses `<source>/<target>` in the parent, so this PR staying consistent with that pattern is fine, but it'd be worth switching project-wide at some point.

### 2. `java.version` is now inconsistent across modules
After this merges:

| module | `java.version` |
| --- | --- |
| `build` (parent) | 1.8 |
| `server` | (inherits 1.8) |
| `console` | 17 |
| `namingserver` | 25 |

`server` and `console` are both on Spring Boot 4.x, which requires Java 17+ at runtime — so `server` inheriting 1.8 works today only because it doesn't happen to use any Java 11+ language features or Java 11+ APIs, but it's fragile. Same class of surprise as #8178. Might be worth a follow-up issue to align the Spring Boot 4.x modules on a single baseline (17), and if `namingserver` really needs 25, document why.

### 3. Nit: a `<!-- required for java.net.http.HttpClient in RestClientConfig -->` comment next to the property would keep the next reader from wondering why console specifically needs 17.

Approving as-is — the fix does exactly what the linked issue calls for.

xuxiaowei-com-cn:
@slievrly 

- I don't understand why the namingserver module needs to use JDK 25.
- The `maven.compiler.release` property is already being used in the console module.

PR #8181, opened 2026-07-24
test: fix flaky AuthControllerWithRandomPasswordTest by isolating context with TestPropertySource

### LINKED ISSUE #8180: AuthControllerWithRandomPasswordTest occasionally fails when running together with other namingserver tests
### Check Ahead

### Ⅰ. Issue Description

`AuthControllerWithRandomPasswordTest` occasionally fails when running alongside all other naming server test classes, but passes reliably when executed individually.

https://github.com/apache/incubator-seata/actions/runs/30069011778/job/89410147885?pr=8167

Image

### Ⅱ. Describe what happened

When all tests under the `namingserver` module are executed together (e.g., via `mvn test` or running the entire test suite in the IDE), `AuthControllerWithRandomPasswordTest` occasionally fails. The test passes reliably when executed in isolation.

The test class is annotated with `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)` and uses `@AutoConfigureMockMvc` to test the authentication endpoints. The test relies on parsing the auto-generated random password from application startup log output via `CapturedOutput`.

When the test runs in isolation, it gets its own fresh Spring Boot application context — the server starts, a random password is generated and logged, and the test successfully extracts it from the captured output to authenticate.

However, when all naming server tests run together, Spring Boot's test context caching mechanism may cause context sharing or resource conflicts between tests using different `webEnvironment` modes:

| Test Class | webEnvironment | Properties / Notes |
|---|---|---|
| `AuthControllerWithRandomPasswordTest` | `RANDOM_PORT` | No explicit properties, `@AutoConfigureMockMvc` |
| `NamingControllerLoggerPrintSmokeTest` | `RANDOM_PORT` | `properties = "console.user.password="`, no MockMvc |
| `NamingControllerPropertiesSmokeTest` | `RANDOM_PORT` | `properties = {"console.user.username=seata", "console.user.password=foo"}`, no MockMvc |
| `AuthControllerWithCustomPropertiesTest` | `MOCK` (default) | `@TestPropertySource(properties = {"console.user.username=seata", "console.user.password=foo"})` |
| `WebSecurityConfigTest` | `MOCK` (default) | No extra properties |
| `NamingControllerTest` | `MOCK` (default) | No extra properties |
| Other tests | Various | — |

Because the test requires the random password generated at application startup (captured via `CapturedOutput`), if the Spring context is reused or the startup log output is not fresh, `CapturedOutput.getOut()` will not contain the expected `"Use the auto-generated password: [...]"` pattern, causing the assertion `assertTrue(matcher.find(), "captured password not found in logs")` to fail.

**In IntelliJ IDEA, when you click on a test class, in simple terms: if a banner appears, it means the test runs in an isolated context; if no banner appears, it means it shares the context with other test classes.**

Image

Image

Image

Image

`AuthControllerWithRandomPasswordTest` should pass both when run individually and when run together with all other naming server tests, without being affected by context caching or execution order.

1. Navigate to the `namingserver` module
2. Run the full test suite:
   ```
   mvn test -pl namingserver
   ```
3. Observe that `AuthControllerWithRandomPasswordTest` fails
4. Run the test in isolation:
   ```
   mvn test -pl namingserver -Dtest=AuthControllerWithRandomPasswordTest
   ```
5. Observe that `AuthControllerWithRandomPasswordTest` passes

**Root Cause:** Spring Boot's test context caching mechanism uses the test class's configuration (annotations, properties, web environment) as the cache key. When multiple `@SpringBootTest(webEnvironment = RANDOM_PORT)` tests run in the same suite without distinguishing properties, the framework may reuse a cached context. If `AuthControllerWithRandomPasswordTest` reuses a context started by another test (e.g., `NamingControllerPropertiesSmokeTest`), the `CapturedOutput` won't contain the `"Use the auto-generated password: [...]"` startup log line — because that context was for a different test run. The `CapturedOutput` extension only captures the output produced during the first startup of that cached context.

**Fix:** Adding `@TestPropertySource(properties = {"uuid=<unique>"})` gives this test class a property that differs from all other `RANDOM_PORT` tests. Spring Boot includes `@TestPropertySource` properties in the context cache key, so this ensures `AuthControllerWithRandomPasswordTest` always gets its own fresh application context — with its own random password generation and its own captured startup log output.

### Ⅵ. Environment

_No response_

xuxiaowei-com-cn:
Fix `AuthControllerWithRandomPasswordTest` by adding a unique `@TestPropertySource` property to ensure Spring Boot creates a dedicated application context, preventing occasional test failures when running alongside other naming server tests.

Fix `AuthControllerWithRandomPasswordTest` to prevent occasional test failures when running alongside other naming server tests.

The test uses `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)` with `@AutoConfigureMockMvc` and relies on `CapturedOutput` to extract an auto-generated random password from application startup logs. When all naming server tests run together, Spring Boot's test context caching mechanism may reuse a context from another `RANDOM_PORT` test (e.g., `NamingControllerPropertiesSmokeTest`). Since `CapturedOutput` only captures the startup log from the first time that cached context was created, the password log line `"Use the auto-generated password: [...]"` will not appear in the output, causing the test to fail.

The fix adds `@TestPropertySource(properties = {"uuid=26d766d7-8aee-4940-bc50-d02ffa6e8a3a"})` — a unique UUID property that differentiates this test's configuration from all other `RANDOM_PORT` tests. Spring Boot includes `@TestPropertySource` properties in the context cache key, so this ensures `AuthControllerWithRandomPasswordTest` always gets its own fresh application context with its own random password and clean startup log output.

fixes https://github.com/apache/incubator-seata/issues/8180

This PR itself fixes an existing test. No new test cases are needed.

1. Run the individual test to confirm it still passes:
   ```
   mvn test -pl namingserver -Dtest=AuthControllerWithRandomPasswordTest
   ```

2. Run the full naming server test suite to confirm the test now passes in the suite context:
   ```
   mvn test -pl namingserver
   ```

3. Verify all other naming server tests continue to pass.

**Root Cause Analysis:**

The `AuthControllerWithRandomPasswordTest` has the following Spring Boot test configuration:
- `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)`
- `@AutoConfigureMockMvc`
- `@ExtendWith(OutputCaptureExtension.class)`
- No explicit `properties` or `classes`

Other naming server tests have different context configurations:

| Test Class | webEnvironment | Properties |
|---|---|---|
| `NamingControllerLoggerPrintSmokeTest` | `RANDOM_PORT` | `console.user.password=` (empty) |
| `NamingControllerPropertiesSmokeTest` | `RANDOM_PORT` | `console.user.username=seata, console.user.password=foo` |
| `AuthControllerWithCustomPropertiesTest` | `MOCK` | `console.user.username=seata, console.user.password=foo` |
| `WebSecurityConfigTest` | `MOCK` | None |
| `NamingControllerTest` | `MOCK` | None |

Since `AuthControllerWithRandomPasswordTest` relies on `CapturedOutput.getOut()` containing the auto-generated password log line (`"Use the auto-generated password: [...]"`) from application startup, it requires a fresh context. Without distinguishing properties, Spring Boot's test context caching may reuse a context from another `RANDOM_PORT` test (e.g., `NamingControllerPropertiesSmokeTest`), whose `CapturedOutput` does not contain the password log line.

The fix adds `@TestPropertySource(properties = {"uuid=26d766d7-8aee-4940-bc50-d02ffa6e8a3a"})` — a unique UUID property. Spring Boot includes `@TestPropertySource` properties in the context cache key, so this differentiates the test's configuration from all other tests, ensuring it always gets its own fresh application context.

**Alternative approaches considered:**
1. **Add a unique `@TestPropertySource` property** (chosen) — Ensures context isolation with minimal change; the UUID property serves purely as a cache-key differentiator without affecting any application logic
2. **Add `@DirtiesContext`** — Would work, but dirties all contexts rather than providing targeted isolation; heavier side effect on overall test suite performance
3. **Specify explicit password properties** — Would make the test configuration identical to a smoke test, but defeats the purpose of testing random password generation
4. **Restructure test execution** (e.g., `@Suite` classes) — More invasive, changes test organization

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -42,6 +42,7 @@ Add changes here for all PR submitted to the 2.x branch.
 
 - [[#8146](https://github.com/apache/incubator-seata/pull/8146)] test: cover rm-datasource resource id initializers
 - [[#8175](https://github.com/apache/incubator-seata/pull/8175)] fix flaky testCachedConfigurationChangeListener due to async race condition
+- [[#8181](https://github.com/apache/incubator-seata/pull/8181)] fix AuthControllerWithRandomPasswordTest flaky failure caused by Spring Boot test context caching
 
 ### refactor:
 
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -41,6 +41,7 @@
 
 - [[#8146](https://github.com/apache/incubator-seata/pull/8146)] test: 覆盖 rm-datasource resource id initializers
 - [[#8175](https://github.com/apache/incubator-seata/pull/8175)] 修复 testCachedConfigurationChangeListener 因异步竞态条件导致的偶发失败
+- [[#8181](https://github.com/apache/incubator-seata/pull/8181)] 修复 AuthControllerWithRandomPasswordTest 因 Spring Boot 测试上下文缓存导致的偶发失败
 
 ### refactor:
 
--- a/namingserver/src/test/java/org/apache/seata/namingserver/AuthControllerWithRandomPasswordTest.java
+++ b/namingserver/src/test/java/org/apache/seata/namingserver/AuthControllerWithRandomPasswordTest.java
@@ -27,6 +27,7 @@ import org.springframework.boot.test.system.CapturedOutput;
 import org.springframework.boot.test.system.OutputCaptureExtension;
 import org.springframework.boot.webmvc.test.autoconfigure.AutoConfigureMockMvc;
 import org.springframework.http.MediaType;
+import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 import tools.jackson.databind.ObjectMapper;
@@ -41,6 +42,15 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
+/**
+ * Use a unique UUID property to ensure this test class gets its own isolated
+ * Spring Boot application context rather than reusing a cached context from
+ * other RANDOM_PORT tests (e.g. NamingControllerPropertiesSmokeTest).
+ * Without this, when all namingserver tests run together, the test framework
+ * may reuse a cached context whose startup logs don't contain the
+ * "Use the auto-generated password: [...]" line, causing the test to fail.
+ */
+@TestPropertySource(properties = {"uuid=26d766d7-8aee-4940-bc50-d02ffa6e8a3a"})
 @SpringBootTest
 @ExtendWith(OutputCaptureExtension.class)
 @AutoConfigureMockMvc
--- a/namingserver/src/test/java/org/apache/seata/namingserver/AuthControllerWithRandomPasswordTest.java
+++ b/namingserver/src/test/java/org/apache/seata/namingserver/AuthControllerWithRandomPasswordTest.java
@@ -43,14 +43,11 @@ import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 /**
- * Use a unique UUID property to ensure this test class gets its own isolated
- * Spring Boot application context rather than reusing a cached context from
- * other RANDOM_PORT tests (e.g. NamingControllerPropertiesSmokeTest).
- * Without this, when all namingserver tests run together, the test framework
- * may reuse a cached context whose startup logs don't contain the
- * "Use the auto-generated password: [...]" line, causing the test to fail.
+ * Add a unique, namespaced test property so Spring's test context cache treats this class as a distinct configuration.
+ * This prevents reuse of an already-started context (e.g. other @SpringBootTest + @AutoConfigureMockMvc tests),
+ * which would omit the startup log line "Use the auto-generated password: [...]" from {@link CapturedOutput}.
  */
-@TestPropertySource(properties = {"uuid=26d766d7-8aee-4940-bc50-d02ffa6e8a3a"})
+@TestPropertySource(properties = "seata.test.context.cache.key=AuthControllerWithRandomPasswordTest")
 @SpringBootTest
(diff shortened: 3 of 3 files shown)


xuxiaowei-com-cn (commented):
LGTM

xuxiaowei-com-cn (commented):
LGTM

slievrly (approved):
Root-cause analysis is solid and the fix is minimally invasive. The `@TestPropertySource` value forces a distinct Spring test-context cache key, which is exactly what's needed to guarantee a fresh context so `CapturedOutput` sees the auto-generated password on startup. Approving.

A few non-blocking observations for follow-up or the PR description:

### 1. Consider `@DirtiesContext(classMode = BEFORE_CLASS)` as an alternative
Semantically it says what we actually mean here — "I need a fresh context because I assert on startup logs" — instead of relying on the more indirect cache-key differentiation. Both are correct; the current choice is fine, just worth mentioning in the code comment so the next maintainer understands *why* this bespoke property exists.

### 2. PR description drift
- The description says the added property is `uuid=26d766d7-8aee-4940-bc50-d02ffa6e8a3a`, but the code actually uses `seata.test.context.cache.key=AuthControllerWithRandomPasswordTest`. The code choice is better (self-documenting), but the description should be updated to match.
- The description says this class uses `@SpringBootTest(webEnvironment = RANDOM_PORT)`, but the actual annotation is a bare `@SpringBootTest` (defaults to `MOCK`) together with `@AutoConfigureMockMvc`. The context-caching problem is still real — the collision is with other `MOCK` + `@AutoConfigureMockMvc` tests (`NamingControllerTest`, `NamingControllerV2Test`, `WebSecurityConfigTest`) rather than with the `RANDOM_PORT` smoke tests. The fix still works either way, but the root-cause table in the PR body is misleading.

### 3. Long-term fragility (out of scope for this PR)
Grepping the startup log for `Use the auto-generated password: [...]` couples the test to a log message format that any refactor could break silently. If the auto-generated password is exposed as a bean or via a testable seam, the test could read it directly and stop depending on `CapturedOutput` altogether — that would kill this class of flakiness at the root. Worth a follow-up issue.

### 4. Namespacing nit
`seata.test.context.cache.key` is a good namespace choice — clearly a test-only key, unlikely to collide with anything in the app. 👍

PR #8020, opened 2026-03-18
feature: add UnregisterRM protocol to notify server on client destroy

WangzJi:
- [x] I have read the [CONTRIBUTING.md](https://github.com/apache/incubator-seata/blob/2.x/CONTRIBUTING.md) guidelines.
- [x] I have registered the PR [changes](https://github.com/apache/incubator-seata/tree/2.x/changes).

### Ⅰ. Describe what this PR did
```mermaid
sequenceDiagram
    participant App as Application
    participant RM as RmNettyRemotingClient
    participant CM as NettyClientChannelManager
    participant TC as TC Server
    participant Proc as UnregRmProcessor
    participant ChMgr as ChannelManager

    App->>RM: destroy()
    RM->>CM: getServerVersion(serverAddress)
    CM-->>RM: serverVersion

    alt serverVersion >= 2.6.0
        RM->>TC: UnregisterRMRequest(appId, group, resourceIds)
        TC->>Proc: process(ctx, rpcMessage)
        Proc->>ChMgr: unregisterRMChannel(channel, resourceIds)
        ChMgr->>ChMgr: remove from RM_CHANNELS and IDENTIFIED_CHANNELS
        ChMgr-->>Proc: done
        Proc->>TC: UnregisterRMResponse(identified=true)
        TC-->>RM: UnregisterRMResponse
    else serverVersion < 2.6.0 or null
        Note right of RM: Skip server not supported
    end

    RM->>CM: clearServerVersions()
    RM->>App: shutdown complete
```

When RmNettyRemotingClient.destroy() is called (e.g., by ShardingSphere closing Seata Client), the server side is not notified that the client is going offline, causing background threads like TableMetaRefreshHolder to continue running and
  eventually throw NullPointerException.

  This PR implements two features to fix this:

  1. Cache server version from RegisterRMResponse (#7593): After successful RM registration, the server version is cached in `NettyClientChannelManager.serverVersionMap` so the client knows what protocol features the server supports.
  2. Add UnregisterRMRequest/Response protocol (#7594): A new message type that allows the client to notify the server to unregister its resources before destroying. This includes:
    - New protocol messages: UnregisterRMRequest (type=105) / UnregisterRMResponse (type=106)
    - Server-side: UnregRmProcessor to handle unregister requests, ChannelManager.unregisterRMChannel() to clean up RM channel entries by resource ID
[...]

PR #8183, opened 2026-07-25
feature: expose TC version to client side via ServerVersionHolder

WangzJi:

When a RM/TM registers to a TC, the TC always echoes its own version back in the register response (`AbstractIdentifyResponse#version`, transferred by both the v0/v1 and the v2 codec as well as by the protobuf convertor). The client, however, only wrote that version into a log line, so upper layers had no way to know which features the target server supports. This is the client side counterpart of `RpcContext#getVersion()`, which the server has been using for compatibility decisions such as `ServerOnRequestProcessor` and `AbstractNettyRemotingServer`.

This PR adds that missing context:

- **Add `ServerVersionHolder`** (`org.apache.seata.core.protocol`), keyed by the server address used to connect, so that any module can ask whether a given TC is above or equal to a version. An unknown version is deliberately treated as "not satisfied", so a feature relying on the check stays disabled until the version is really known.
- **Record the server version on both RM and TM registration.** `RmNettyRemotingClient` fed a private map introduced by #8020, while `TmNettyRemotingClient` did not record the version at all. Both now write to the holder.
- **Remove the package private `serverVersionMap` in `NettyClientChannelManager`.** It was only reachable from within `org.apache.seata.core.rpc.netty` and only fed by the RM path, so it could not serve feature detection outside the rpc layer. Its only consumer, the `UnregisterRMRequest` version check, now uses the holder.
- **Pass the peer version to `registerChannel`.** The client used to store *its own* version in `Version#VERSION_MAP` keyed by the *server* address, while the server stores the *peer* version keyed by the client address. `Version#getChannelVersion` now means "the version of the peer" on both sides.

Entries in the holder are written on every successful registration and are never removed on disconnection. This is intentional: TM and RM reach the same servers through two independent channel managers, so removing an entry when one of them disconnects would break feature detection for the other one. A version is an intrinsic attribute of the server at a given address, and a stale entry can never be read for a live request, because a request always goes through a channel whose registration has just refreshed the entry.

No.

Test cases are added:

- `ServerVersionHolderTest`: put/get, overwrite on re-registration, blank input ignored, unknown server treated as not satisfied, `clear`.
- `TmNettyClientTest#onRegisterMsgSuccessRecordsServerVersionTest`: new, covers the previously missing TM path.
- `RmNettyClientTest`: asserts that the peer version is the one handed to `registerChannel` and that it lands in the holder; the `UnregisterRMRequest` version cases now drive the holder instead of stubbing the removed manager method.
- `NettyClientChannelManagerTest`: the pure version cases moved to `ServerVersionHolderTest`, the destroy/cleanup cases keep asserting the channel, pool key and lock cleanup.

```
./mvnw -pl core -am test -Dtest='ServerVersionHolderTest,VersionTest,NettyClientChannelManagerTest,RmNettyClientTest,TmNettyClientTest'
```

For a behavioural check, start a TC and a client, and observe that after registration `ServerVersionHolder.getServerVersion(serverAddress)` returns the TC version for both TM and RM connections, and that the `UnregisterRMRequest` on client destroy is still sent only to servers of 2.6.0 and above.

The change to `registerChannel` is the only behavioural change outside the new holder, and it is safe: the sole client side reader of `Version#VERSION_MAP` is `MsgVersionHelper#versionNotSupport`, whose `SKIP_MSG_CODE_V0` contains only `TYPE_RM_DELETE_UNDOLOG`, a message sent by the TC to the RM and never in the opposite direction. The check is therefore inert on the client side today, and correcting the stored version aligns the two sides instead of changing runtime behaviour.

`ServerVersionHolder` is kept separate from `Version` on purpose. `Version#VERSION_MAP` is keyed by `channel.remoteAddress()`, whereas the holder is keyed by the server address obtained from the registry; the two are not always equal (NAT, proxy, container port mapping), so merging them into one class would invite mixing the keys up.

`Version#VERSION_2_6_0` is promoted from private to public so that the comparison target can be expressed as a constant at the call site.


slievrly (commented):
Nice consolidation — the client-side `RpcContext#getVersion()` counterpart is genuinely missing today, and fixing the `registerChannel(..., version)` argument to be the peer's version (not the client's own) closes a real inconsistency with the server side. The `ServerVersionHolder` API is clean and the test coverage is solid.

One thing I'd like to push back on though: **why remove the cleanup mechanism entirely?**

The old per-manager `serverVersionMap` was removed on `destroyChannel` / `cleanupDisconnectedChannelMetadata` / `destroy()`. The new `ServerVersionHolder` is a JVM-wide static that is only ever written to — nothing in production code calls `clear()`, and the disconnect / destroy paths no longer touch it. The PR body explains the rationale (TM and RM are two independent channel managers, so removing on one disconnect would break feature detection for the other), and that reasoning is internally consistent for a shared holder. But it trades one problem for a different one:

### 1. Unbounded growth over the JVM lifetime

Every server address the client has ever registered to stays in the map forever. In a stable cluster this is a few entries. In deployments where TC addresses rotate — blue/green rollouts, autoscaling, Kubernetes pod IPs, registry churn — every historical address accumulates. Each entry is small, but there is no upper bound and no eviction. That's a slow leak the old design didn't have.

### 2. Static global state with no destroy hook

The old `serverVersionMap` was owned by the channel manager and cleaned up on `destroy()`. `ServerVersionHolder`'s state now outlives every client instance. In embedded scenarios where the Seata client is torn down and re-initialized in the same JVM (uncommon but real — hot reload, test harness, multi-tenant containers), the second instance inherits the first instance's map. If the TC at a given address was replaced between the two instantiations, the stale version wins until the next successful registration overwrites it. `ServerVersionHolder.clear()` exists but is only called from `@BeforeEach` in the test — nothing calls it in production.

### 3. Suggested alternatives

Any of these keeps the "don't break the other manager on single-manager disconnect" invariant while restoring a cleanup path:

- **Reference-counting on the shared holder.** Track which managers have registered each address (e.g. a `Set<String> managerIds` per address); remove the version entry only when the count drops to zero. Cleanup then happens automatically when the last manager for that address disconnects.
- **Cleanup on client-level destroy, not per-channel disconnect.** Keep per-channel disconnects as no-ops (as this PR does), but have the RM and TM clients coordinate a "when both are destroyed, clear the holder" hook. Simpler than ref-counting, still bounded.
- **Bounded cache instead of `ConcurrentHashMap`.** A small size cap (e.g. Caffeine with maximumSize=1024) makes the memory footprint statically bounded without requiring lifecycle coordination.

If the deliberate choice is "we accept the leak because in practice the address set is small", it'd be worth calling that out explicitly in the Javadoc so a future reader doesn't try to add cleanup and re-introduce the TM/RM interference bug.

Not blocking — just want to make sure the removal of the existing cleanup is a considered decision rather than a side-effect of moving to a shared holder.

WangzJi:
@slievrly Thanks for the thoughtful feedback. I agree that the shared holder needs a production cleanup path.

I've added client-lifecycle coordination: RM and TM attach on initialization and detach on destroy; versions are cleared only after the last active client is destroyed. This preserves the shared state across a single manager/channel disconnect while preventing stale entries from surviving a full client teardown. Tests cover both cases.

PR #8187, opened 2026-07-28
bugfix: fix read registry network config from local configuration

Zhengcy05:
- [x] I have read the [CONTRIBUTING.md](https://github.com/apache/incubator-seata/blob/2.x/CONTRIBUTING.md) guidelines.
- [x] I have registered the PR [changes](https://github.com/apache/incubator-seata/tree/2.x/changes).

### Ⅰ. Describe what this PR did

This PR fixes the server-side registry network configuration lookup for `registry.preferredNetworks` and `registry.ignoredInterfaces`.

Previously, `Server` read these options through `ConfigurationFactory.getInstance()`. When `seata.config.type=nacos` is configured, that instance points to the config center, so local Spring Boot configuration such as `seata.registry.preferred-networks` and `seata.registry.ignored-interfaces` in `application.yml` may be ignored.

This PR changes the lookup to:
- read from `ConfigurationFactory.CURRENT_FILE_INSTANCE` first, which is consistent with other registry-local configuration lookup paths
- fallback to `ConfigurationFactory.getInstance()` when the local registry configuration does not provide the value
- add `ignoredInterfaces` to `RegistryProperties`
- add regression coverage for the local-first lookup behavior

### Ⅱ. Does this pull request fix one issue?

fixes #8182

### Ⅲ. Why don't you add test cases (unit test/integration test)?

Added test cases.

### Ⅳ. Describe how to verify it

- `mvn -pl server -am -Dtest=org.apache.seata.server.ServerConfigTest -DfailIfNoTests=false test`
- `mvn -pl spring/seata-spring-autoconfigure/seata-spring-autoconfigure-core -am -Dtest=org.apache.seata.spring.boot.autoconfigure.properties.registry.RegistryPropertiesTest -DfailIfNoTests=false test`

### Ⅴ. Special notes for reviews

Issue #8169, opened 2026-07-17
NetUtil cached local address bypasses ignoredInterfaces and preferredNetworks


pyshiweijia:
### Check Ahead

- [x] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

- [x] I am willing to try to fix this bug myself.

### Ⅰ. Issue Description

`NetUtil.getIgnoredInterfacesLocalAddress(String[], String...)` returns the global `LOCAL_ADDRESS` immediately when it has already been initialized. As a result, `ignoredInterfaces` and `preferredNetworks` are not evaluated when an earlier call has populated the default local-address cache. The global cache predates the ignored-interface support added in #8090.

### Ⅱ. Describe what happened

After a default lookup caches a local address, a lookup that ignores every interface still returns the cached address.

```java
@Test
public void testIgnoredInterfacesAreAppliedAfterDefaultAddressIsCached() {
    assertThat(NetUtil.getLocalAddress()).isNotNull();
    assertThat(NetUtil.getIgnoredInterfacesLocalAddress(new String[] {".*"})).isNull();
}
```

```text
expected: null
 but was: /192.168.124.25
```

The same early return also bypasses `preferredNetworks`.

### Ⅲ. Describe what you expected to happen

Parameterized lookups should evaluate the supplied interface filters and preferred-network rules regardless of whether an unfiltered address was previously cached. The unfiltered no-argument lookup may continue to use the global cache.

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. Checkout the `2.x` branch.
2. Add the test above to `NetUtilTest`.
3. Run `mvn -pl common -am test "-Dtest=NetUtilTest#testIgnoredInterfacesAreAppliedAfterDefaultAddressIsCached" "-Dsurefire.failIfNoSpecifiedTests=false" "-Dlicense.skip=true" -B`.

The test fails because `getIgnoredInterfacesLocalAddress` returns before calling `getLocalAddress0`.

### Ⅴ. Anything else we need to know?

Suggested direction: keep caching only for semantically unfiltered lookups. Calls with non-empty `ignoredInterfaces` or `preferredNetworks` should resolve using their parameters without reading or overwriting the default cache. A cache keyed by normalized parameters is another option.

Impact is conditional: server startup is unaffected when the configured lookup is the first local-address query. Existing tests only assert non-null results and do not verify that the supplied rules affect selection.

### Ⅵ. Environment

[...]

PR #8090, opened 2026-05-06
optimize: support ignored network interfaces configuration

xuxiaowei-com-cn:
Add the environment variable `REGISTRY_IGNOREDINTERFACES=VMware.*,VirtualBox.*` configuration to filter out unnecessary network interfaces (e.g., VMware, VirtualBox, Docker).

This solves IP address selection issues in environments with multiple network interfaces, especially when VM software is installed.

Changes:
- Add getIgnoredInterfacesLocalIp/Address methods in NetUtil
- Add ignoreInterface method with regex pattern matching support
- Add REGISTRY_IGNORED_INTERFACES constant in StarterConstants
- Integrate ignoredInterfaces config in Server startup

Inspired by Spring Cloud's `spring.cloud.inetutils.ignored-interfaces`

- [x] I have read the [CONTRIBUTING.md](https://github.com/apache/incubator-seata/blob/2.x/CONTRIBUTING.md) guidelines.
- [x] I have registered the PR [changes](https://github.com/apache/incubator-seata/tree/2.x/changes).

### Ⅱ. Does this pull request fix one issue?

https://github.com/apache/incubator-seata/issues/8089

PR #8202, opened 2026-08-13
bugfix: apply ignoredInterfaces and preferredNetworks when the default address is cached

angus-guo:
## Issue

`getIgnoredInterfacesLocalAddress(String[], String...)` returns the global `LOCAL_ADDRESS` immediately when it has already been initialized, regardless of the supplied `ignoredInterfaces` or `preferredNetworks` patterns. Once an unfiltered lookup (through `getLocalAddress()` or `getLocalIp()`) has populated the cache, a subsequent filtered lookup is answered from that cache without evaluating its patterns.

**Symptom:** After `Server.start()` made registry settings reachable from `application.yml` in #8187, the filters configured in YAML are read successfully but then silently ignored here.

This was discovered in #8169:

```java
@Test
public void testIgnoredInterfacesAreAppliedAfterDefaultAddressIsCached() {
    assertThat(NetUtil.getLocalAddress()).isNotNull();
    assertThat(NetUtil.getIgnoredInterfacesLocalAddress(new String[] {".*"})).isNull();
}
```

```text
expected: null
 but was: /192.168.124.25
```

## Change

Keep caching only for semantically unfiltered lookups. A call with non-empty `ignoredInterfaces` or `preferredNetworks` resolves through `getLocalAddress0(...)` without reading or overwriting the default cache:

```java
if (CollectionUtils.isEmpty(ignoredInterfaces) && CollectionUtils.isEmpty(preferredNetworks)) {
    if (LOCAL_ADDRESS != null) {
        return LOCAL_ADDRESS;
    }
    InetAddress localAddress = getLocalAddress0(ignoredInterfaces, preferredNetworks);
    LOCAL_ADDRESS = localAddress;
    return localAddress;
}
return getLocalAddress0(ignoredInterfaces, preferredNetworks);
```

`getLocalAddress()` and `getLocalIp()` (no arguments → both `null`) remain cached. `Server.start()` performs its filtered lookup once at startup, so no hot path loses caching.

The global cache predates the ignored-interface support added in #8090, and the early return was never conditioned on whether the supplied parameters matched the cached call's parameters.

## Tests

Three new cases in `NetUtilTest`:

- **`testIgnoredInterfacesAreAppliedAfterDefaultAddressIsCached`** — the pattern `.*` matches every interface, so the result is `null` rather than the cached address. **Fails before this change.**
- **`testPreferredNetworksAreAppliedAfterDefaultAddressIsCached`** — the same early return bypassed `preferredNetworks`. Ignoring every interface through the preferred-network path also yields no candidate, so the result is `null`. **Fails before this change.**
- **`testUnfilteredLookupRemainsCached`** — the unfiltered lookup keeps returning the same cached instance (no regression).

Both failing tests demonstrate the bug from #8169 and confirm the fix. Existing tests in the same file only assert non-null results and do not verify that the supplied rules affect selection, which is why the bug was not caught earlier.

Verified:
- All 33 tests in `NetUtilTest` pass.
- All 693 tests in the `common` module pass.
- `spotless:check` and `checkstyle:check` pass on the `common` module.

## Checklist

- [x] **Check Ahead**
  - [x] I have searched the [issues](https://github.com/apache/incubator-seata/issues) of this repository and believe that this is not a duplicate.
- [ ] I have added the change log of the current PR to [changes/en-us/2.x.md](https://github.com/apache/incubator-seata/blob/2.x/changes/en-us/2.x.md) and [changes/zh-cn/2.x.md](https://github.com/apache/incubator-seata/blob/2.x/changes/zh-cn/2.x.md). *(Will add once the PR number is assigned.)*
- [x] I have passed all the tests locally.
- [x] I have added the unit test for my change.

### DIFF
--- a/changes/en-us/2.x.md
+++ b/changes/en-us/2.x.md
@@ -33,6 +33,7 @@ Add changes here for all PR submitted to the 2.x branch.
 - [[#8179](https://github.com/apache/incubator-seata/pull/8179)] set java.version to 17 for console module to fix compilation error
 - [[#8170](https://github.com/apache/incubator-seata/pull/8170)] fix ConfigTools encryption corrupting non-ASCII content on non-UTF-8 platforms
 - [[#8182](https://github.com/apache/incubator-seata/issues/8182)] fix registry preferred networks and ignored interfaces from Spring Boot application.yml
+- [[#8202](https://github.com/apache/incubator-seata/pull/8202)] fix `ignoredInterfaces` and `preferredNetworks` being silently ignored when the unfiltered local address is already cached
 
 
 
--- a/changes/zh-cn/2.x.md
+++ b/changes/zh-cn/2.x.md
@@ -34,6 +34,7 @@
 - [[#8179](https://github.com/apache/incubator-seata/pull/8179)] 将 console 模块的 java.version 设置为 17 以修复编译错误
 - [[#8170](https://github.com/apache/incubator-seata/pull/8170)] 修复 ConfigTools 加密在非 UTF-8 默认字符集平台损坏非 ASCII 内容的问题
 - [[#8182](https://github.com/apache/incubator-seata/issues/8182)] 修复 Spring Boot application.yml 中 registry preferred networks 和 ignored interfaces 配置不生效的问题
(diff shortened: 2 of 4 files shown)


angus-guo:
Closing this as it duplicates #8191, which was merged on Aug 15. Both PRs fixed the same bug (filtered lookups bypassing the cached default address) with the same implementation and similar test coverage.

Thanks for catching this independently — good to see the same issue being spotted from different angles.

PR #8203, opened 2026-08-15
optimize: merge channel availability log output and adjust log level

### LINKED ISSUE #5474: optimize channel availability thread log output

### Ⅰ. Issue Description

image
建议改为debug下输出堆栈，info下不输出堆栈，并且将所有失败的地址合并输出，而不是10个里面有5个地址不可用输出五次日志得形式

### Ⅱ. Describe what happened

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- Seata version:  
- OS :
- Others:

shenshichao163-oss:
## What this PR does

Optimizes the channel availability thread log output in `NettyClientChannelManager`.

### Changes

1. **info level: no stack traces** — Changed `LOGGER.error()` to `LOGGER.info()` for the summary log, which outputs all failed addresses in a single entry without stack traces.
2. **debug level: detailed stack traces** — Changed `LOGGER.error()` to `LOGGER.debug()` for per-address detailed logs with full stack traces.
3. **Merged output** — All failed addresses are output in one log entry at info level (using `failedMap.keySet()`), instead of one log per failed address.
4. **Independent log levels** — Changed `else if` to two independent `if` blocks, so info and debug levels work independently rather than being mutually exclusive.

### Issue Link

Fixes #5474

### Before

- `LOGGER.isInfoEnabled()` → `LOGGER.error()` (wrong level, outputs error-level logs when info is enabled)
- `else if LOGGER.isDebugEnabled()` → `LOGGER.error()` (only reached when info is disabled)
- Log level mismatch: info branch uses error, debug branch uses error

### After

- `LOGGER.isInfoEnabled()` → `LOGGER.info()` (summary, no stack traces)
- `LOGGER.isDebugEnabled()` → `LOGGER.debug()` (per-address detailed stack traces)
- Both levels work independently

### Checklist

- [x] I have searched the [issues](https://github.com/apache/incubator-seata/issues) of this repository and believe that this is not a duplicate.
- [x] I have checked that this modification does not cause CI to fail.
- [x] I will sign the Apache CLA if required.

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientChannelManager.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/NettyClientChannelManager.java
@@ -251,17 +251,21 @@ class NettyClientChannelManager {
                 }
             }
             if (failedMap.size() > 0) {
+                // Merge all failed addresses into a single log entry to avoid flooding logs.
+                // info level: output summary without stack traces
                 if (LOGGER.isInfoEnabled()) {
-                    LOGGER.error(
+                    LOGGER.info(
                             "{} can not connect to {} cause:{}",
                             FrameworkErrorCode.NetConnect.getErrCode(),
                             failedMap.keySet(),
                             failedMap.values().stream()
                                     .map(Throwable::getMessage)
                                     .collect(Collectors.toSet()));
-                } else if (LOGGER.isDebugEnabled()) {
+                }
+                // debug level: output detailed stack traces for each failed address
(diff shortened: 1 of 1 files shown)


WangzJi (changes requested):
Thanks for the fix. The log level changes make sense, but #5474 is not fully addressed yet. `doConnect()` still logs the exception at ERROR before `doReconnect()` aggregates it, so an INFO-level client still prints one stack trace per unavailable address. Could you handle this path as well and keep stack traces only at DEBUG?

PR #8204, opened 2026-08-15
docs: add configuration architecture section to config-center README

### LINKED ISSUE #5608: confused about the documentation [important]

shenshichao163-oss:
## What this PR does

Adds a **Configuration Architecture** section to `script/config-center/README.md` to clarify the relationship between the two configuration files (`application.yml` and `config.txt`) and how they relate to each other.

### Changes

Added a new section before "Script Introduction" that includes:

1. **Explanation of the two configuration files**:
   - `application.yml` — Server-side Spring Boot configuration (port, logging, config/registry/store type)
   - `config.txt` — Client-side transaction configuration (transport, vgroup mapping, store settings)

2. **Architecture diagram** showing the config flow:
   - Server reads `application.yml` → connects to Config Center
   - Config Center stores `config.txt` content (pushed by scripts)
   - Client reads from Config Center

3. **FAQ section** answering:
   - Where should `seata.config.nacos.data-id` be placed?
   - What does `config.txt` do?
   - Should I use `config.txt` or `application.yml`?

### Issue Link

Fixes #5608

### Motivation

Issue #5608 reports confusion about Seata configuration: which file goes where, what each config item means, and how `application.yml` and `config.txt` relate to each other. This PR adds clear documentation directly in the config-center README to address these questions.

### Checklist

- [x] I have searched the issues of this repository and believe this is not a duplicate.
- [x] I have checked that this modification does not cause CI to fail.
- [x] I will sign the Apache CLA if required.

### DIFF
--- a/script/config-center/README.md
+++ b/script/config-center/README.md
@@ -30,6 +30,77 @@ you only need to follow the instructions below and keep the corresponding config
 | store.db.user            | service.vgroupMapping.default_tx_group                       |
 | store.db.password        | service.disableGlobalTransaction                             |
 
+## Configuration Architecture
+
+Before using these scripts, it is important to understand the Seata configuration architecture:
+
+### Two Configuration Files
+
+Seata uses two distinct configuration files:
+
+1. **`server/src/main/resources/application.yml`** — The Seata **server** configuration file (Spring Boot style). It contains:
+   - Server port, logging, and Spring application settings
+   - `seata.config.type` — Where the server reads its configuration from (`file`, `nacos`, `consul`, `apollo`, `zk`, `etcd3`)
+   - `seata.registry.type` — Where the server registers itself for service discovery (`file`, `nacos`, `eureka`, `redis`, `zk`, `consul`, `etcd3`, `sofa`, `seata`)
+   - `seata.store.mode` — Transaction log storage mode (`file`, `db`, `redis`, `raft`)
+
+2. **`script/config-center/config.txt`** — The **client-side** configuration file (key-value properties format). It contains:
+   - Transport settings (protocol, heartbeat, thread pool)
+   - Transaction group mappings (`service.vgroupMapping.xxx`)
+   - Client-side settings (lock, undo, retry policies)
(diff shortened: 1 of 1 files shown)


WangzJi (commented):
Thanks for working on this. The new section is helpful, but a couple of configuration roles need to be corrected:

1. `seata.config.nacos.data-id` is not client-only. The server also needs it when using an all-in-one Nacos configuration such as `seataServer.properties`.
2. `config.txt` is not a client-side configuration file. It is the source used by the publishing scripts and contains shared, client-only, and server-only settings.
3. The server does not read a local `config.txt` at runtime, and using both files is not always required.

Could you update the description and diagram to reflect these distinctions?

PR #1966, opened 2019-11-30
feature: add single send request for client.

objcoding:
### Ⅰ. Describe what this PR did
add single send request for client.
add 'transport.enable-client-batch-send-request' config param.

PR #1105, opened 2019-05-24
refactor TmRpcClient & RmClient for common use.

cherrylzhao:
### Ⅰ. Describe what this PR did
  1. simplify TmRpcClient & RmRpcClient
  2. extract  `NettyClientChannelManager` for common use.

### Ⅲ. Why don't you add test cases (unit test/integration test)? 
have added unit test

### Ⅳ. Describe how to verify it
NettyClientChannelManagerTest

### Ⅴ. Special notes for reviews
I have extracted a `NettyClientChannelManager` and `RpcClientBootstrap`, this can simplify TmRpcClient and RmRpcClient. there exist a lot of redundancy codes before.

PR #8205, opened 2026-08-20
optimize: park merge send thread when idle to avoid 1ms polling CPU spin

### LINKED ISSUE #6041: rpcMergeMessageSend_TMROLE_1、rpcMergeMessageSend_RMROLE_1的cpu使用率过高

### Ⅰ. Issue Description
seata版本：1.4.2
服务配置：-Xms512m -Xmx1024m -XX:MaxNewSize=128m
因一直收到线上某服务cpu使用率过高的告警，就先nacos进行下线，准备重启大法，发现还是一直告警。然后arthas排查，最后发现无流量情况下，
rpcMergeMessageSend_TMROLE_1、rpcMergeMessageSend_RMROLE_1这2线程的cpu使用率过高
### Ⅱ. Describe what happened
"rpcMergeMessageSend_TMROLE_1" Id=31 cpuUsage=31% TIMED_WAITING on java.lang.Object@1c0c220d
    at java.lang.Object.wait(Native Method)
    -  waiting on java.lang.Object@1c0c220d
    at io.seata.core.rpc.netty.AbstractRpcRemotingClient$MergedSendRunnable.run(AbstractRpcRemotingClient.java:301)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
    at java.lang.Thread.run(Thread.java:748)

    Number of locked synchronizers = 1
    - java.util.concurrent.ThreadPoolExecutor$Worker@4998e74b

"rpcMergeMessageSend_RMROLE_1" Id=34 cpuUsage=30% TIMED_WAITING on java.lang.Object@25b70c1e
    at java.lang.Object.wait(Native Method)
    -  waiting on java.lang.Object@25b70c1e
    at io.seata.core.rpc.netty.AbstractRpcRemotingClient$MergedSendRunnable.run(AbstractRpcRemotingClient.java:301)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
    at java.lang.Thread.run(Thread.java:748)

    Number of locked synchronizers = 1
    - java.util.concurrent.ThreadPoolExecutor$Worker@2b08772d
    - 

```
```

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version(e.g. `java -version`):1.8
- Seata client/server version:  1.0.0/1.4.2
- Database version:
- OS(e.g. `uname -a`):
- Others:

CuriousLinYu:
Fixes #6041

## Problem

`MergedSendRunnable` polls every 1ms (`MAX_MERGE_SEND_MILLS = 1`) even when all baskets are empty. With no traffic, the thread wakes up 1000 times per second, each cycle acquiring `mergeLock`, iterating `basketMap` and going back to sleep. Issue #6041 reported this at ~30% CPU per thread (thread dump shows the `rpcMergeMessageSend_*_1` threads in `TIMED_WAITING`), and it still reproduces on 2.x (RM client enables batch send by default: `DEFAULT_ENABLE_RM_CLIENT_BATCH_SEND_REQUEST = true`).

## Solution

When `basketMap` is empty, the merge thread parks on `mergeCondition.await()` (no timeout) instead of a 1ms timed wait. Producers already offer to the basket and then signal (`sendSyncRequest`), so the parked thread is woken up as soon as a message arrives and drains the basket immediately.

About the removed 1ms timed wait:
- The offer-then-signal path has existed since the early implementation (the merge thread predates `AbstractRpcRemotingClient`; see the history of `AbstractNettyRemotingClient.java`, e.g. PR #1105 and PR #1966) — `sendSyncRequest` always signals when the merge thread is not sending.
- funky-eyes pointed out the same idea in issue #6041: the signal from `sendSyncRequest` is enough to wake the thread, so a timed wait is not required for correctness.
- The 1ms wait therefore acted as a fallback poll on top of the signal path; keeping it after the park adds a >=1ms floor to the first message after an idle period without buying meaningful batching (see measurements below).

### Race safety

- The check-and-wait is atomic: `isBasketEmpty()` is evaluated and `await()` is entered while holding `mergeLock`.
- Producers offer to the basket *before* acquiring `mergeLock` and signalling (unchanged code), so a message can never sit in the basket with the merge thread parked indefinitely: either the producer's signal wakes the parked thread, or the producer sees `isSending == true` and the thread is already in the send loop which drains the basket on the next iteration.
- No lost wake-up: a producer can only signal after the thread has parked (`Condition.await` releases the lock atomically), and the empty-check is re-evaluated inside the `while` loop.

## Measurements

Measured locally with an instrumented test client (basket-level offer/poll instrumentation, no network needed) across three variants. Thread CPU time via `ThreadMXBean.getThreadCpuTime` over a 10s idle window; batch sizes over 10 rounds of a 4x25 concurrent burst.

| variant | idle CPU (10s window) | first-message drain latency | batch size (avg / max) |
|---|---|---|---|
| original (1ms poll) | 31ms (0.312%), TIMED_WAITING | 0.17 - 6.2ms | 21.3 / 95 |
| park + 1ms await | 0ms (0.000%), WAITING | 1.4 - 15.3ms | 83.3 / 100 |
| park + immediate drain (this PR) | 0ms (0.000%), WAITING | 0.09 - 0.21ms | 50.0 / 100 |

The absolute idle-CPU ratio depends on the machine and CPU quota (issue #6041 measured ~30% in a constrained container). The relevant comparison is original > 0 vs this PR = 0. The "park + 1ms await" variant shows the >=1ms latency floor that this change removes, while batching does not regress vs the original.

## Tests

- Added `testMergedSendRunnableIdleWaitState`: asserts the `rpcMergeMessageSend` thread is NOT in `TIMED_WAITING` after 300ms idle, then submits a request and verifies the basket is drained.
- Full `NettyRemotingClientBehaviorTest`: 69 tests, 0 failures.

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
@@ -587,6 +587,20 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
             while (true) {
                 mergeLock.lock();
                 try {
+                    // Park until there are pending messages, so the merge thread no longer
+                    // burns CPU with a 1ms polling cycle when idle. The check-and-wait is
+                    // atomic under mergeLock and producers offer to the basket before
+                    // signalling (see sendSyncRequest), so no wake-up can be lost.
+                    while (isBasketEmpty()) {
+                        isSending = false;
+                        mergeCondition.await();
+                    }
+                    isSending = true;
+                    if (LOGGER.isDebugEnabled()) {
+                        LOGGER.debug("merge send thread woken up with pending messages");
+                    }
+                    // Keep the original 1ms merge window, so messages arriving within this
+                    // window are still batched into a single request as before.
                     mergeCondition.await(MAX_MERGE_SEND_MILLS, TimeUnit.MILLISECONDS);
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
@@ -637,6 +651,21 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
                 });
                 isSending = false;
             }
+        }
+
+        /**
+         * Checks whether all baskets are empty. The merge thread parks itself
+         * when this returns true, avoiding the idle 1ms polling busy loop.
+         *
+         * @return true if every basket in basketMap is empty
+         */
+        private boolean isBasketEmpty() {
+            for (BlockingQueue<RpcMessage> basket : basketMap.values()) {
+                if (!basket.isEmpty()) {
+                    return false;
+                }
+            }
+            return true;
         }
 
         private void printMergeMessageLog(MergedWarpMessage mergeMessage) {
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
@@ -82,7 +82,6 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
     private static final String MSG_ID_PREFIX = "msgId:";
     private static final String FUTURES_PREFIX = "futures:";
     private static final String SINGLE_LOG_POSTFIX = ";";
-    private static final int MAX_MERGE_SEND_MILLS = 1;
     private static final String THREAD_PREFIX_SPLIT_CHAR = "_";
     private static final int MAX_MERGE_SEND_THREAD = 1;
     private static final long KEEP_ALIVE_TIME = Integer.MAX_VALUE;
@@ -596,12 +595,6 @@ public abstract class AbstractNettyRemotingClient extends AbstractNettyRemoting 
                         mergeCondition.await();
                     }
                     isSending = true;
-                    if (LOGGER.isDebugEnabled()) {
-                        LOGGER.debug("merge send thread woken up with pending messages");
-                    }
-                    // Keep the original 1ms merge window, so messages arriving within this
-                    // window are still batched into a single request as before.
-                    mergeCondition.await(MAX_MERGE_SEND_MILLS, TimeUnit.MILLISECONDS);
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
                     LOGGER.warn("MergedSendRunnable wait interrupted", e);
--- a/core/src/test/java/org/apache/seata/core/rpc/netty/NettyRemotingClientBehaviorTest.java
+++ b/core/src/test/java/org/apache/seata/core/rpc/netty/NettyRemotingClientBehaviorTest.java
@@ -1493,12 +1493,13 @@ public class NettyRemotingClientBehaviorTest {
 
             // The merge thread must be parked on Condition.await (WAITING) instead of
             // spinning on a 1ms timed wait (TIMED_WAITING) when idle
-            for (Map.Entry<Thread, StackTraceElement[]> entry : Thread.getAllStackTraces().entrySet()) {
+            for (Map.Entry<Thread, StackTraceElement[]> entry :
+                    Thread.getAllStackTraces().entrySet()) {
                 Thread thread = entry.getKey();
                 if (thread.getName().startsWith("rpcMergeMessageSend")) {
-                    assertFalse(Thread.State.TIMED_WAITING == thread.getState(),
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingClient.java
```diff
@@ -587,6 +587,20 @@ public void run() {
             while (true) {
                 mergeLock.lock();
                 try {
+                    // Park until there are pending messages, so the merge thread no longer
+                    // burns CPU with a 1ms polling cycle when idle. The check-and-wait is
+                    // atomic under mergeLock and producers offer to the basket before
+                    // signalling (see sendSyncRequest), so no wake-up can be lost.
+                    while (isBasketEmpty()) {
+                        isSending = false;
+                        mergeCondition.await();
```
funky-eyes, line 595: With this approach, whenever a message arrives, the processing thread waits for up to another 1 ms. Why is this additional delay necessary?

With the implementation you’re currently using, consider the case where only a single message arrives: it wakes up the thread, but the thread then waits for another 1 ms before sending the message. This effectively adds about 1 ms to the request latency.

I don’t think an `await` of 1 ms alone should cause such high CPU usage. Could the high CPU utilization be related to Arthas being enabled? Given the performance of modern CPUs, 1,000 wake-ups per second that do essentially no work should not normally be enough to consume around 30% CPU.

CuriousLinYu: Thanks for the review — both points were spot on. I've removed the post-wakeup `await(1ms)` entirely: the thread now drains immediately after being signalled. Measured locally: first-message drain latency drops from ~1.4-15.3ms (with the 1ms await) to ~0.09-0.21ms, and batching does not regress — avg batch size is 50 vs 21.3 on the original under a 4x25 concurrent burst. The full measurement table is in the PR description.

On the 30% CPU number — fair point that Arthas sampling and container CPU quota matter. What the issue confirms is the thread dump: the `rpcMergeMessageSend_*_1` threads sit in `TIMED_WAITING` while burning CPU. Locally (ThreadMXBean, 10s window) the original implementation consumes a steady 31ms of thread CPU time per 10s at idle; this PR takes it to 0. The absolute ratio is environment-dependent, but the idle wake-up cost is real and now gone.

CuriousLinYu:
Hi @funky-eyes, thanks for the review — both points were addressed in `2e39148`: the post-wakeup `await(1ms)` is gone, so the merge thread now drains immediately after being signalled (first-message drain latency drops from ~1.4-15.3ms to ~0.09-0.21ms, and batching does not regress — avg batch size 50 vs 21.3 on the original under a 4x25 burst). Details are in the reply above and in the PR description.

Two process-side items are still blocking the merge:

1. **CI has not run on this PR at all** — there are no check runs on the head commit and the combined status is `pending`. For a first-time contributor the workflow usually needs a maintainer to approve the run. Could you approve it when you get a chance?
2. The review thread above is still open (the code it refers to has since changed). If `2e39148` looks right to you, resolving it would clear the `blocked` merge state.

Happy to make any further adjustments.

CuriousLinYu:
Pushed a merge with the latest `2.x` (`41694e140`) to re-trigger CI — the fork's workflow runs need approval again.

On the previous spotless failure: I could not reproduce it locally. `./mvnw spotless:check` passes on the full reactor (103 modules) under JDK 17, and `core` is clean (366 files) after clearing the spotless index. The failing run finished in 22s, while a neighbouring PR's spotless job took 37s and passed, so I suspect that run failed while resolving dependencies rather than on an actual violation.

Could someone approve the run so we can see a real result? If it comes back with a specific file, I'll fix it immediately.

PR #8209, opened 2026-08-22
feature: upgrade Netty to 4.2.15.Final

xjlgod:

- Upgrade Netty from `4.1.137.Final` to `4.2.15.Final` through the Netty BOM.
- Migrate Seata transport event loops from deprecated Netty 4.1 APIs to the Netty 4.2 `IoHandler` / `MultiThreadIoEventLoopGroup` APIs.
- Keep the allocator explicitly configured as `pooled`, and remove use of Netty internal APIs.
- Ensure Seata's Netty BOM takes precedence over Spring Boot dependency management in server-related modules.
- Align related gRPC/Protobuf dependency management and add required direct Netty dependencies for the SkyWalking plugin.
- Update the gRPC interceptor test to use an actual `grpc-netty` TCP server and client.
- Add Chinese and English change-log entries under `changes/`.

No.

N/A. This PR updates existing tests and executes the relevant test suites.

In particular, `GrpcTest` was migrated from the in-process transport to a real `grpc-netty` TCP transport, so the test now verifies the Netty-based gRPC request path. Existing core remoting, HTTP/2, Etcd registry, and version-compatibility tests were also executed.

1. Run the relevant unit tests on JDK 8:
   - Core Netty remoting, HTTP/2 handler, channel event, and Netty client bootstrap tests.
   - Etcd registry tests.
   - `extensions/rpc/seata-grpc` interceptor tests.

2. Run the cross-version compatibility tests on JDK 21:
   - Verify client/server combinations across the old and new versions.

3. Build the release distribution with JDK 25 and the `release-seata` profile.

4. Inspect the generated server and naming-server distributions:
   - All bundled Netty artifacts must resolve to `4.2.15.Final`.
   - No `netty-4.1.x` artifacts should remain.
   - Startup scripts must retain `-Dio.netty.allocator.type=pooled`.

- Netty 4.2 is not intended to coexist with Netty 4.1 in the same runtime classpath. This PR centralizes version resolution through the Netty BOM and overrides Spring Boot-managed Netty versions where necessary.
- The migration is intentionally scoped to deprecated event-loop APIs and direct internal-API usage; business protocol behavior is unchanged.
- The SkyWalking plugin now declares direct `provided` Netty dependencies because it directly uses Netty `Channel` and `AttributeMap` types.


WangzJi (commented):
Could you confirm whether seata-console also needs to be aligned with Netty 4.2.15? Under the source-release profile, its effective dependency tree still resolves io.netty:* to 4.2.12.Final via Spring Boot 4.0.6. If only the final server/namingserver distributions need alignment and they already resolve to 4.2.15, no change is needed; otherwise, console should import the Netty BOM before Spring Boot as well.

Issue #2531, opened 2020-04-09
IDENTIFIED_CHANNELS  缓存堆积导致 SeataServer FULLGC 挂掉


objcoding:
- [ ] I have searched the [issues](https://github.com/seata/seata/issues) of this repository and believe that this is not a duplicate.

### Ⅰ. Issue Description
在K8S环境下部署的Nacos+SeataServer，在跑一段时间后SeataServer里面io.seata.core.rpc.ChannelManager的IDENTIFIED_CHANNELS map会被Put大量对象，造成SeataServer不断的FULLGC最后挂掉。

### Ⅱ. Describe what happened
![image](https://user-images.githubusercontent.com/24354760/78905862-2f532680-7ab1-11ea-9c01-8a814042f0d5.png)

  If there is an exception, please attach the exception trace:

```
Just paste your stack trace here!
```

### Ⅳ. How to reproduce it (as minimally and precisely as possible)

1. xxx
2. xxx
3. xxx

### Ⅵ. Environment:

- JDK version :
- OS :
- Others:

PR #8223, opened 2026-09-11
optimize: fix IDENTIFIED_CHANNELS SeataServer FULLGC OOM #2531

saul-c:
optimize: fix IDENTIFIED_CHANNELS SeataServer FULLGC OOM #2531


saul-c:
close

PR #8224, opened 2026-09-12
bugfix: fix IDENTIFIED_CHANNELS leak causing TC FullGC and OOM

saul-c:

When TM/RM clients disconnect (e.g. K8s pod rolling update, HPA scaling), RpcContext.release() only nulled the clientIDHolderMap field instead of removing the dead channel from ChannelManager.IDENTIFIED_CHANNELS. Dead Netty channels (with RpcContext/pipeline/ByteBuf) accumulated forever until the TC ran into continuous FullGC and OOM.

- RpcContext.release(): remove the channel from IDENTIFIED_CHANNELS
- ChannelManager.releaseRpcContext(): remove the channel defensively
- AbstractNettyRemotingServer.handleDisconnect: route through the unified releaseRpcContext entry (used by both READER_IDLE and channelInactive)
- NettyClientChannelManager.releaseChannel(): release rpc context too
- add ChannelManagerOOMTest reproducing repeated reconnect/disconnect

Related: [https://github.com/apache/incubator-seata/issues/2531](https://github.com/apache/incubator-seata/issues/2531)

### DIFF
--- a/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingServer.java
+++ b/core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingServer.java
@@ -215,6 +215,7 @@ public abstract class AbstractNettyRemotingServer extends AbstractNettyRemoting 
             }
             if (rpcContext != null && rpcContext.getClientRole() != null) {
                 rpcContext.release();
+                ChannelManager.releaseRpcContext(ctx.channel());
                 if (LOGGER.isInfoEnabled()) {
                     LOGGER.info("remove channel:" + ctx.channel() + "context:" + rpcContext);
                 }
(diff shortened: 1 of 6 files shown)


WangzJi (changes requested):
@saul-c Thanks for fixing this leak! 

Removing the context breaks TM response fallback in `getSameClientChannel()` and can cause an NPE if disconnect cleanup runs between `isRegistered()` and the context lookup. Could you preserve the client identity needed to route pending responses, retrieve the request context once with an explicit null check, and add regression tests for both cases?

Also, the license check is failing for `ChannelManagerOOMTest.java`; please use the standard ASF license header.

core/src/main/java/org/apache/seata/core/rpc/netty/AbstractNettyRemotingServer.java
```diff
@@ -215,6 +215,7 @@ private void handleDisconnect(ChannelHandlerContext ctx) {
             }
             if (rpcContext != null && rpcContext.getClientRole() != null) {
                 rpcContext.release();
+                ChannelManager.releaseRpcContext(ctx.channel());
```
funky-eyes, line 218: If the server happens to need this channel to send a response at that moment, releasing it prematurely could result in an NPE. Therefore, I believe there should be a delayed cleanup mechanism to ensure that the channel is only released once the server can no longer reference or use it.

PR #8229, opened 2026-09-21
optimize: fix LatencyUtils LICENSE pointer and console npm ci

Bughue:
## Summary
- Point `org.latencyutils:LatencyUtils` at `licenses/LatencyUtils-BSD-2-Clause` instead of HdrHistogram's file (different copyright holders). Stop excluding that file from the binary assembly. Same class of error as babel-plugin-macros.
- Fix nested JS `see:` pointers: `string_decoder` and `postcss-value-parser`.
- Update the three confirmed console **runtime** versions that drifted from `package-lock.json`: `@alicloud/console-components-console-menu` 1.3.0, `axios` 1.16.1, `lodash` 4.18.1.
- Use `npm ci` (and designer `npm ci`) so RC frontend builds follow the lockfile.
- Keep Snap.svg in binary NOTICE: `copyDesigner.js` bundles the saga designer (including PathMap.js) into `seata-console`, which both server and namingserver ship.

This does **not** fully regenerate the frontend LICENSE from the webpack bundle. Nested copies of the same npm name still have multiple versions; blindly rewriting every line from the lock top-level version would be wrong. A bundle-based inventory remains follow-up.

Out of scope here: Docker Hub tag cleanup, replacing `v2.7.0` with a new RC tag.

## Test plan
- [ ] `distribution/LICENSE` LatencyUtils line uses `LatencyUtils-BSD-2-Clause`
- [ ] `release-seata.xml` no longer excludes that file
- [ ] Unpacked binary `licenses/` contains `LatencyUtils-BSD-2-Clause` and LICENSE `see:` targets exist
- [ ] console build uses `npm ci`
- [ ] Do not start the next IPMC vote on the old `v2.7.0` tag (`b08ffbd1c`)

Made with [Cursor](https://cursor.com)


funky-eyes (approved):
LGTM

